PR #5, opened 2019-04-01
Creates an endpoint for bulk loading csv files.

mildewey:
The csv files must provide appropriate _key fields for nodes and _to and _from fields for edges.


mildewey:
I verified that we are good on both points. Thanks!

PR #14, opened 2019-04-25
Add necessary watchers for paging functions

waxlamp:
@mildewey


waxlamp:
@mildewey: are any more needed? I will squash these commits into a single one when we've found all instances.

PR #18, opened 2019-04-26
Hooking up the ability to add tables. 

mildewey:
This is not fully fleshed out as it ignores fields and primaryKey.

### DIFF
--- a/multinet/multinet/resolvers.py
+++ b/multinet/multinet/resolvers.py
@@ -108,7 +108,7 @@ def create_graph(root, info, workspace, name, nodeTypes, edgeTypes):
     db.create_graph(graph, nodeTypes, edgeTypes)
     return graph
 
-def create_table(root, info, workspace, name, primaryKey='_id', fields=[]):
+def create_table(root, info, workspace, name, edges=False, primaryKey='_id', fields=[]):
     table = Table(workspace, name)
-    db.create_table(table)
+    db.create_table(table, edges)
     return table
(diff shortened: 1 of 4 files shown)


multinet/multinet/resolvers.py
```diff
@@ -108,7 +108,7 @@ def create_graph(root, info, workspace, name, nodeTypes, edgeTypes):
     db.create_graph(graph, nodeTypes, edgeTypes)
     return graph
 
-def create_table(root, info, workspace, name, primaryKey='_id', fields=[]):
+def create_table(root, info, workspace, name, edges=False, primaryKey='_id', fields=[]):
```
waxlamp: This `edges` parameter seems like it could present a pitfall. Does it make more sense to make explicit `create_node_table` and `create_edge_table` actions, or is that overdesigning? The tradeoff here is "forgetting" to set `edges` when it's an edge table, vs being overdesigned. On the other hand, this might be moot if we will eventually create, say, an endpoint to create the different kinds of table, and then this flag just becomes hidden complexity.

@mildewey what do you think?

mildewey: Yeah, it's a headache. I really didn't want to introduce a differentiation between edge tables and regular tables, but we're stuck with it.

Another alternative to what you suggested is inferring edges or not from the fields. If the fields include _to or _from we could use it.

A point in favor of your plan, though, is that we could have edge tables specify not just their primary key, but their 'source' and 'target' keys. WDYT?

waxlamp: I think for now we should leave `_from` and `_to` as the "source" and "target" keys. So let's just go with your solution, and if we need to generalize, or if this `edges` flag causes trouble, we can revisit.

PR #19, opened 2019-04-29
Implement file upload and create button component

jenrogers1207:


client/src/views/test.vue
```diff
@@ -0,0 +1,21 @@
+<template>
```
mildewey: Please remove this file from the commit.

jenrogers1207: Will do! If I push my changes to this branch, it will be reflected in this pull request?

mildewey: Yes

PR #21, opened 2019-04-29
Break resolvers apart into multiple files

mildewey:
The resolvers are roughly organized around already-defined graphql types or logical groups of those types.


mildewey:
@waxlamp we had considered the idea of defining the types locally in these files. I tried to do that and found that I was stuck constantly referencing other files where those types were defined. In the end I felt like that was more confusing and smelly than just leaving the types in a single file.

multinet/multinet/attribute.py
```diff
@@ -0,0 +1,6 @@
+
```
waxlamp: Is this line blank for a specific reason?

multinet/multinet/entity.py
```diff
@@ -0,0 +1,43 @@
+from . import db
+from .types import *
```
waxlamp: I'm not a fan of `import *` because it makes it harder to cross-reference names in the file later. I'd prefer if this were an explicit import of the names used, or just `from . import types`.

multinet/multinet/graph.py
```diff
@@ -0,0 +1,31 @@
...
+# that represent the possible properties of that edge
+def edge_types(graph, info):
+    return db.graph_edge_types(graph)
+
+# each node type is a set of keys that can be pulled from various tables
+# that represent the possible properties of that node
+def node_types(graph, info):
+    return db.graph_node_types(graph)
+
+def graph_name(graph, info):
```
waxlamp: I think this should just be `name` since the method name can be fully qualified at the call site as `graph.name()`. What do you think?

multinet/multinet/table.py
```diff
@@ -0,0 +1,21 @@
+from . import db
+from .types import *
+
+# A table is a raw block of data that is not directly associated with a graph.
+# Each table has a logical set of fields and rows. 
+
+def table_rows(table, info):
```
waxlamp: s/table_//

multinet/multinet/table.py
```diff
@@ -0,0 +1,21 @@
...
+    return RowQuery(table.workspace, table.table, None, None)
+
+def table_name(table, info):
```
waxlamp: s/table_//

multinet/multinet/table.py
```diff
@@ -0,0 +1,21 @@
...
+    return table.table
+
+def table_fields(table, info):
```
waxlamp: s/table_//

multinet/multinet/workspace.py
```diff
@@ -0,0 +1,20 @@
+from . import db
+from .types import *
+
+# a workspace is a representation of tables and graphs that are inter-related and
+# potentially co-accessible.
+
+def workspace_name(workspace, info):
```
waxlamp: s/workspace_//

multinet/multinet/workspace.py
```diff
@@ -0,0 +1,20 @@
...
+    return workspace
+
+def workspace_tables(workspace, info):
```
waxlamp: s/workspace_//

multinet/multinet/workspace.py
```diff
@@ -0,0 +1,20 @@
...
+    return [Table(workspace, table) for table in db.workspace_tables(workspace)]
+
+def workspace_graphs(workspace, info):
```
waxlamp: s/workspace_//

multinet/multinet/attribute.py
mildewey: Nope, removed.

multinet/multinet/entity.py
mildewey: Now that we have smaller scopes, this cleaned up nicely.

multinet/multinet/graph.py
mildewey: I had done similar things in other places and have now done it here too. I also simplified query and mutation.

multinet/multinet/table.py
mildewey: done

multinet/multinet/table.py
mildewey: done

multinet/multinet/table.py
mildewey: done

multinet/multinet/workspace.py
mildewey: done

multinet/multinet/workspace.py
mildewey: done

multinet/multinet/workspace.py
mildewey: done

waxlamp:
Thanks @mildewey--the changes look great!

PR #23, opened 2019-05-02
Make miscellaneous small-scale improvements (GraphQL refactor; upload endpoint namespacing; bugfixes)

waxlamp:
~This pull request attempts to create a MultiNet endpoint that can be used as a drop-in replacement for the Phovea endpoints currently used by Juniper. Currently, the "get single node" action is implemented, but it works only for Author nodes.~

The PR ~also~ includes some bug fixes:
- a typo in the newick loader
- missing type imports in the GraphQL infrastructure

It also introduces application namespaces within the API. The Newick tree loader is under a `newick` namespace, while the ~Juniper actions will go under a `juniper` namespace~ bulk loader has been renamed to the CSV loader (in the `csv` namespace).

### DIFF
--- a/multinet/multinet/__init__.py
+++ b/multinet/multinet/__init__.py
@@ -39,9 +39,6 @@ class MultiNet(Resource):
         # Newick tree operations.
         self.route('POST', ('newick', 'tree', ':workspace', ':table'), self.tree)
 
-        # Operations for Juniper application.
-        self.route('POST', ('juniper', 'node'), self.juniper_get_node)
-
     @access.public
     @autoDescribeRoute(
         Description('Receive GraphQL queries')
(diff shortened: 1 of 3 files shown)


multinet/multinet/__init__.py
```diff
@@ -15,14 +15,32 @@
 from .schema import schema
 from . import db
 
+
+def graphql_query(query):
```
mildewey: I like moving this into its own function.

mildewey: It would probably be to our benefit to use a mutation query through this endpoint to create/update tables/graphs based on the bulk loading endpoints. That way we can avoid having db-specific code in the handlers for those endpoints. If we do that, we should probably also consider graphql endpoints for adding nodes and lists of nodes... That is all probably outside the scope of this task, though.

In short, keep this even if we only use it in one place for now.

multinet/multinet/__init__.py
```diff
@@ -15,14 +15,32 @@
...
     def __init__(self, port):
         super(MultiNet, self).__init__()
         self.resourceName = 'multinet'
         self.arango_port = port
         self.route('POST', ('graphql',), self.graphql)
         self.route('POST', ('bulk', ':workspace', ':table'), self.bulk)
-        self.route('POST', ('tree', ':workspace', ':table'), self.tree)
+
+        # Newick tree operations.
+        self.route('POST', ('newick', 'tree', ':workspace', ':table'), self.tree)
```
mildewey: A newick is a tree, so we don't need to re-specify that. I had used 'tree' when I thought we had a reader that handled multiple file formats that all essentially boiled down to being trees. I would advocate for keeping a form of this, and altering the bulk load endpoint to be a csv endpoint. We could also hypothetically use non-standard mime types, but that is discouraged.

So:
self.route('POST', ('newick', ':workspace', ':table'), self.tree)
self.route('POST', ('csv', ':workspace', ':table'), self.tree)

multinet/multinet/__init__.py
```diff
@@ -15,14 +15,32 @@
...
+
+        # Operations for Juniper application.
+        self.route('POST', ('juniper', 'node'), self.juniper_get_node)
```
mildewey: As discussed in chat, we would like to drop this endpoint in favor of moving this logic to the client and possibly expanding into new graphql query types.

waxlamp: Reverted the commit.

multinet/multinet/__init__.py
waxlamp: Done in 1a8dae4.

multinet/multinet/__init__.py
waxlamp: See issue #25

PR #24, opened 2019-05-03
fix bug in file type selection used in add table query

jenrogers1207:
There is a very initial table that loads the rows for a selected table as well, but the main reason for this is the bug fix.

### DIFF
--- a/client/src/components/FileInput.vue
+++ b/client/src/components/FileInput.vue
@@ -38,7 +38,7 @@ export default {
       let fileName = file.name.split('.')
       let extension = fileName[fileName.length - 1]
       for(let type in this.types){
-        if(this.types[type].includes(extension)){
+        if(this.types[type].extension.includes(extension)){
           return type
         }
       }
--- a/client/src/components/FileInput.vue
+++ b/client/src/components/FileInput.vue
@@ -49,12 +49,11 @@ export default {
   watch: {
     selectedType(){
        this.$emit("handle-file-input", [this.$refs.file.files, this.selectedType])
-       //this.$emit("handleFileInput", this.$refs.file.files)
     }
   }
 }
 </script>
 
 <style scoped>
-h1{ color:brown }
+
 </style>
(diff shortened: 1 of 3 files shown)


client/src/components/FileInput.vue
```diff
@@ -38,7 +38,7 @@ export default {
       let fileName = file.name.split('.')
       let extension = fileName[fileName.length - 1]
       for(let type in this.types){
-        if(this.types[type].includes(extension)){
+        if(this.types[type].extension.includes(extension)){
           return type
```
waxlamp: I think this whole section could possibly be simplified using JavaScript's `some()` and `endswidth()` methods. Something like:

```javascript
if (this.types[type].extension.some(ext => extension.endswith(ext))) {
  return type;
}
```

but it's not urgent to make this change.

client/src/views/TableDetail.vue
```diff
@@ -1,9 +1,17 @@
...
   <div>
     <h1>Table: {{`${this.workspace}/${this.table}`}}</h1>
+
+<!-- 
     <ul>
-      <li v-for="field in fields" :key="field">{{field}}</li>
+     <li v-for="field in fields" :key="field">{{field}}</li> 
+       <li v-for="row in rowKeys" :key="row[0].value">{{row[0].value}}</li>
     </ul>
+-->
```
waxlamp: If we're really dumping this code that's been commented out, I'd prefer if we actually cut it out altogether.

client/src/views/TableDetail.vue
```diff
@@ -15,18 +23,28 @@ export default {
   props: ['workspace', 'table'],
   data () {
     return {
-      fields: []
+      //fields: []
```
waxlamp: Same thing -- let's cut the code out if it shouldn't be here.

waxlamp:
If the table work isn't complete yet, it's totally fine to do a PR just to fix that bug. Let me know if you'd like to split the bugfix out from the rest, or if you'd still like to merge this as-is.

jenrogers1207:
Removed commented code. I'm thinking we can merge as-is and then I'll create a separate branch for the table improvement. I'm still currently styling this!

PR #26, opened 2019-05-06
Initial Table Detail Implemented

jenrogers1207:


client/src/views/TableDetail.vue
```diff
@@ -4,17 +4,19 @@
...
     </th>
     </tr>
   </thead>
-  <tbody v-for="row in rowKeys" :key="row[0].value" class="row-wrap">
-    <tr class="row">
+  <tbody v-for="(row, index) in rowKeys" :key="row[0].value" class="row-wrap">
+
+    <tr :class="classNameByIndex(index)">
       <td v-for="col in row" :key="col.key" class="col">
       {{col.value}}
```
waxlamp: I think this line and the next one need to be indented?

client/src/views/TableDetail.vue
```diff
@@ -29,10 +31,17 @@ export default {
   props: ['workspace', 'table'],
   data () {
     return {
-      rowKeys:[]
+      rowKeys:[],
+      headers:[]
     }
+  },
+  computed: {
```
waxlamp: If there are no computed properties, can we get rid of these lines?

client/src/views/TableDetail.vue
```diff
@@ -29,10 +31,17 @@ export default {
...
-      rowKeys:[]
+      rowKeys:[],
+      headers:[]
     }
+  },
+  computed: {    
+    
   },
   methods: {
+    classNameByIndex(index) {
```
waxlamp: I would probably call this something more directly descriptive, like `rowClassName()` maybe.

client/src/views/TableDetail.vue
```diff
@@ -52,7 +61,7 @@ export default {
   
       let table = response.data.data.tables[0];
       this.rowKeys = table.rows.rows.map(r=> r.columns.filter(c=> c.key != "_rev"))
-     
+      this.headers = this.rowKeys[0].map(k=> k.key.startsWith("_") ? k.key.split("_")[1] : k.key);
```
waxlamp: That `k.key.split("_")[1]` looks dangerous to me--it won't work properly if a key name has an underscore somewhere in the middle.

If the intent is just to strip off a leading underscore, then I think `k.key.slice(1)` will work better.

client/src/views/TableDetail.vue
```diff
@@ -75,9 +84,17 @@ table{
...
+ padding: 15px 25px;
+ letter-spacing:1.5px;
 }
-tr.row {
+tr.even-row {
   background-color: #F3F6F6;
+  padding: 10px 10px;
+}
+tr.odd-row {
+
```
waxlamp: Can we get rid of this blank line?

client/src/views/TableDetail.vue
```diff
@@ -2,27 +2,23 @@
...
+      <tr >
+      <th v-for="head in this.headers" :key="head" class="head">
+        {{head}}
+      </th>
+      </tr>
+    </thead>
+    <tbody v-for="(row, index) in rowKeys" :key="row[0].value" class="row-wrap">
+      <tr :class="rowClassName(index)">
+        <td v-for="col in row" :key="col.key" class="col">
+        {{col.value}}
```
waxlamp: Can you indent lines 15 and 16 with two more spaces?

client/src/views/TableDetail.vue
```diff
@@ -36,7 +32,7 @@ export default {
     }
   },
   methods: {
-    classNameByIndex(index) {
+    rowClassName(index) {
```
waxlamp: Does this method get called anywhere? I expected that if the name changed, some invocation of it would change as well.

jenrogers1207: It gets called on line 13

client/src/views/TableDetail.vue
jenrogers1207: fixed those!

client/src/views/TableDetail.vue
waxlamp: Ah, thanks, completely missed it!

PR #28, opened 2019-05-13
Remove 0-byte "Icon" files

waxlamp:


jenrogers1207:
I have no idea what these are for and I'm not using them - OK to delete!

PR #29, opened 2019-05-14
update table list in workspace detail

jenrogers1207:
add styling and a display name to table links

### DIFF
--- a/client/src/views/WorkspaceDetail.vue
+++ b/client/src/views/WorkspaceDetail.vue
@@ -14,11 +14,11 @@
                 <v-button :onClick="loadFile">create table</v-button>
             </div>
           </div>
-          <ul>
-          <li v-for="table in tables" :key="table">
-            <router-link :to="`/${workspace}/table/${table}`">{{table}}</router-link>
-          </li>
-        </ul>
+          <div class = "table-link-wrap">
+          <div v-for="table in tables" :key="table.name" class="table-link">
+            <router-link :to="`/${workspace}/table/${table.name}`">{{table.display}}</router-link>
+          </div>
+        </div>
       </div>
       <div>
         <label>Graphs</label>
@@ -69,7 +69,11 @@ export default {
           graphs { name }
         }
       }`});
-      this.tables = response.data.data.workspaces[0].tables.map(table => table.name);
+      this.tables = response.data.data.workspaces[0].tables.map(table => {
+        let name = table.name;
+        let displayName = name.includes("_") ? name.replace(/_/g, " ") : name;
+        return {name: table.name, display:displayName }
+      });
       this.graphs = response.data.data.workspaces[0].graphs.map(graph => graph.name);
     },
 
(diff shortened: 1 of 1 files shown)


client/src/views/WorkspaceDetail.vue
```diff
@@ -69,7 +69,11 @@ export default {
           graphs { name }
         }
       }`});
-      this.tables = response.data.data.workspaces[0].tables.map(table => table.name);
+      this.tables = response.data.data.workspaces[0].tables.map(table => {
+        let name = table.name;
+        let displayName = name.includes("_") ? name.replace(/_/g, " ") : name;
+        return {name: table.name, display:displayName }
```
waxlamp: Would be better to format this more clearly like

```javascript
return {
  name: table.name,
  display: displayName
};
```

client/src/views/WorkspaceDetail.vue
```diff
@@ -69,7 +69,11 @@ export default {
           graphs { name }
         }
       }`});
-      this.tables = response.data.data.workspaces[0].tables.map(table => table.name);
+      this.tables = response.data.data.workspaces[0].tables.map(table => {
+        let name = table.name;
+        let displayName = name.includes("_") ? name.replace(/_/g, " ") : name;
```
waxlamp: Two things:
1. Why are we changing underscores to spaces in table names?
2. I don't think you need to check for underscores before performing the substitution; you can just say

```javascript
const displayName = name.replace(/_/g, " ");
```

client/src/views/WorkspaceDetail.vue
jenrogers1207: My plan was to leave the graph side alone until we had the functionality to upload/view the graphs but I can definitely add the styling changes to the available graphs consistent with the tables

client/src/views/WorkspaceDetail.vue
jenrogers1207: re 1) - I made a display name variable that had a space instead of an underscore only for readability.

waxlamp: My intuition is that this will eventually lead to trouble. I'd say keep the underscores in the name, even if it looks better with spaces.

client/src/views/WorkspaceDetail.vue
jenrogers1207: added styling to reflect table list

client/src/views/WorkspaceDetail.vue
jenrogers1207: That totally makes sense  - I reverted to the original name containing the underscore

client/src/views/WorkspaceDetail.vue
jenrogers1207: This reverted to the original mapping so this return statement doesn't exist.

jenrogers1207:
This return statement no longer exists.

client/src/views/WorkspaceDetail.vue
waxlamp: Ah gotcha.

PR #31, opened 2019-05-15
Add mechanism for linking out to vis apps

waxlamp:


client/src/views/GraphDetail.vue
```diff
@@ -14,7 +14,15 @@
         <label>Edge Types</label>
         <ul>
           <li v-for="table in edgeTypes" :key="table">
-            <router-link :to="`/${workspace}/table/${table}`">{{table}}</router-link>
+            <router-link :to="`/workspaces/${workspace}/table/${table}`">{{table}}</router-link>
+          </li>
+        </ul>
+      </div>
+      <div style="border-style: solid; width: 100%">
```
jenrogers1207: At a later time - it might be worth replacing this inline styling with a class and adding a style tag?

waxlamp: Yes, definitely!

Actually going through the codebase and removing the inline styles is something I was thinking of doing pretty soon.

PR #37, opened 2019-05-22
change ROOT to APP in password variable name from step five of set up

jenrogers1207:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
waxlamp: Can you remove this file?

Issue #48, opened 2019-06-04
GraphQL Query Error - Table Data Returns Null.


jenrogers1207:
The table detail no longer shows data due to an error in the query. Old query structure no longer works. 

Current query used in application:
```
const response = await api().post('multinet/graphql', {query: `query {
        tables (workspace: "${this.workspace}", name: "${this.table}") {
          name,
          rows{
            total,
            rows(offset: 0, limit: 30){
              key,
              columns{
                key,
                value
              }
            }
          },
        }
      }`});
```
 Error that is returned from query  `Cannot query field "rows" on type "RowList" ` in returned response. Data returns null. 

Screenshot of error message:
<img width="919" alt="Screen Shot 2019-06-04 at 4 00 48 PM" src="https://user-images.githubusercontent.com/22663748/58919207-12f09b80-86e2-11e9-9eaf-c907a3577cac.png">

PR #49, opened 2019-06-04
fix table detail query

jenrogers1207:
Fixed query for table detail specified in issue #48

### DIFF
--- a/client/src/views/WorkspaceDetail.vue
+++ b/client/src/views/WorkspaceDetail.vue
@@ -75,6 +75,8 @@ export default {
 
     async loadFile(){
       let queryType = this.fileTypes[this.selectedType].queryCall;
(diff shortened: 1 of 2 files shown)


client/src/views/WorkspaceDetail.vue
```diff
@@ -75,6 +75,8 @@ export default {
 
     async loadFile(){
       let queryType = this.fileTypes[this.selectedType].queryCall;
+      console.log("queryType", queryType)
```
waxlamp: Let's get rid of this `console.log()` statement, and the one in the next line.

waxlamp: (This should also get rid of the failing style tests on CircleCI.)

PR #54, opened 2019-06-07
Switch project over to Vuetify

waxlamp:
This is to simplify our investment in dealing with component style, and in preparation for having Jared do a design once-over on the project.


waxlamp:
Ah yes I will take care of it!

PR #56, opened 2019-06-07
Add UI for graph creation

waxlamp:


waxlamp:
The general way is to merge master to the topic branch, fix conflicts, push, and re-review (which is exactly what I'll do here).

PR #76, opened 2019-07-11
Sidebar component

jtomeck:
Replace workspaces page with a persistent app-wide sidebar to both create and navigate workspaces.

**To do:**
~1. Integrate login/logout GWC~
~2. Add ability to delete/rename workspaces~

NOTE(waxlamp): I created new issues for both of these TODOs.


client/src/views/Index.vue
```diff
@@ -0,0 +1,57 @@
+<template>
```
waxlamp: Could we rename this to something like `FrontPage`?

client/src/components/WorkspaceDialog.vue
```diff
@@ -0,0 +1,79 @@
...
+  methods: {
+    async create () {
+      if (this.newWorkspace) {
+        const response = await api().post('multinet/graphql', {query: `mutation {
+          workspace (name: "${this.newWorkspace}" )
+        }`});
+
+        if (response.data.data) {
+          this.$router.push(`/workspaces/${this.newWorkspace}`);
+          this.dialog = false;
```
waxlamp: There's an issue in the creation logic where the Sidebar component doesn't always get refreshed with the new workspace. I believe it *does* get updated when the view actually changes. That is, in the following situations, the sidebar will correctly update:
1. If the view is the Index/front page, and you create a workspace, the view switches to WorkspaceDetail for the new workspace, and the sidebar correctly refreshes.
2. If the view is WorkspaceDetail or GraphDetail or TableDetail, and you create a new workspace, *then switch to a different `*Detail` view*, then the sidebar will refresh.

I think the fix is to have the creation logic inform its parent that a new workspace has been created, but I am not totally sure of the best way to express this in Vue. I'd like to discuss it with @subdavis before we put in a fix.

waxlamp: I spoke with @subdavis about this: he recommended one of three paths:
1. Use VueX and let the workspace list be updated from changes in application state.
2. Emit a "created" event from the new workspace dialog informing the sidebar of the newly created workspace name, and let it update its own internal state.
3. Emit a "please create" event from the new workspace dialog, and let the sidebar perform the creation and data update.

I think (1) is the most future-proof solution, since it can handle other kinds of application state as well, but since it doesn't seem to be needed right now I'm going with (2).

The problem with (3) is the question of what to do if the workspace creation fails for any reason; how and where do we then display the error condition? I would prefer that such errors simply appear in the modal dialog, so that if there's a fix, it can be offered there, or if not, the user can decide whether to try again or to abandon the action.

jtomeck:
> 2\. Make the "multinet" link at the top left go back to the "front page" view (unless @jtomeck thinks that's a bad UI/UX idea).

I don't think it's a bad idea, but I'm not sure it's really needed. The "front page" is just instructions on how to use the application and there's really no reason for them to go back there. Do you agree?

waxlamp:
> @waxlamp
> 
> > 1. Figure out the right fix for updating the sidebar on workspace creation.
> 
> I'm open to anything. Did you want to sit down and talk about it first thing on Monday?
> 

I think commit 7cb77fc is the right fix, at least for now.

> > 2. Make the "multinet" link at the top left go back to the "front page" view (unless @jtomeck thinks that's a bad UI/UX idea).
> 
> I don't think it's a bad idea, but I'm not sure it's really needed. The "front page" is just instructions on how to use the application and there's really no reason for them to go back there. Do you agree?

Yeah, I agree.

waxlamp:
@jtomeck thanks for all the work on this.

If it looks like it's in good shape, could you remove the [WIP] tag and merge?

PR #89, opened 2019-07-18
Page level styling

jtomeck:
Begin building a better structure to the different views (ie. workspace/table/node/graph detail pages). Currently this only includes the workspace detail page.

To do:
- [x] Refactor various blocks into separate components
- [ ] ~Build solution to forms~
- [ ] ~Get girder web component for multiple file upload installed~
- [ ] ~Build the rest of the views (which should be easy once one view is complete)~

UPDATE(waxlamp): I created #91 and #92 for TODO items 2 and 3 above, and TODO 4 can be taken care of view-by-view in separate PRs.

### DIFF
--- a/client/src/components/ItemPanel.vue
+++ b/client/src/components/ItemPanel.vue
@@ -1,0 +1,116 @@
+<template>
+  <v-list
+    dark
+    subheader
+  >
+    <v-subheader class="pr-2">
+      {{title}}
+
+      <v-spacer />
+
+        <v-tooltip top>
+          <template v-slot:activator="{ on }">
+            <v-scroll-x-transition>
+              <v-btn
+                flat
+                icon
+                v-if="anySelected"
+                v-on="on"
+              >
+                <v-icon color="red accent-2">delete_sweep</v-icon>
+              </v-btn>
+            </v-scroll-x-transition>
+          </template>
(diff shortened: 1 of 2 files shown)


waxlamp:
I can work on the refactoring.

What do you mean by "build solution to forms"?

As for GWC for multifile upload, and building the rest of the views, I think we should just handle those things one-by-one in follow-on PRs. What do you think?

jtomeck:
> What do you mean by "build solution to forms"?

Just that I haven't made any decisions on what I want to do. ie. bury them behind a button that toggles a modal, or have them served to you up front, etc.

> As for GWC for multifile upload, and building the rest of the views, I think we should just handle those things one-by-one in follow-on PRs. What do you think?

I think I agree, but the multi-file upload component would effect how I build/style the form in general so it might be good to have at least this one moving forward? What are your thoughts?

waxlamp:
@jtomeck take a look at my refactoring and let me know what you think.

jtomeck:
> one-by-one in follow-on PRs

@waxlamp I know you were only talking about one specific thing, but I wanted to ask if you want that for the other views as well. My plan was to take care of most of the styling of the rest of the views in this PR, but it might make sense to split them out into multiple.

PR #93, opened 2019-07-19
Factor the Sidebar component out of the views that reference it

### LINKED ISSUE #79: Use router-view to avoid repeating Sidebar component in every view

### LINKED ISSUE #84: Maintain sidebar scroll when navigating
I noticed in the demo @waxlamp gave today that the sidebar scrolls to the top on navigation. Perhaps it is re-rendering when it shouldn't?

waxlamp:
@jtomeck and @subdavis, please take a look to see if this looks right.

Closes #79.
Fixes #84 (attn @jeffbaumes).


jeffbaumes:
Not having tested it, the logic looks right. Put the sidebar outside the router view so it persists.

subdavis:
I don't know if applies in this situation, so this is more of a PSA than a recommendation:

https://vuejs.org/v2/guide/components-dynamic-async.html#keep-alive-with-Dynamic-Components

If any of your pages take significant cycles to render, this is one option to consider.

waxlamp:
subdavis wrote:
> I don't know if applies in this situation, so this is more of a PSA than a recommendation:
> 
> https://vuejs.org/v2/guide/components-dynamic-async.html#keep-alive-with-Dynamic-Components
> 
> If any of your pages take significant cycles to render, this is one option to consider. Looks good!

Very informative, thanks!

I thought the router specifically *doesn't* tear down and recreate the components used in its routes though? That tabbing example seems to say that the tabbed-between components are recreated.

subdavis:
> I thought the router specifically doesn't tear down and recreate the components

Without `keep-alive`, vue router will destroy components that aren't active on the view.  It's not a `v-show` type situation.   their destroyed and unmounted hooks will call.  

With `keep-alive` there are new lifecycle hooks `Activated` and `Deactivated` that inform the component of its state in the view.

PR #94, opened 2019-07-22
[WIP] Add CSV validation at rest endpoint, and catch response on front-end

### LINKED ISSUE #88: Add CSV Node Table Validation
Validate that there are no duplicate keys in a node table, when uploading via CSV.

AlmightyYakob:
Fixes #88
Clean up imports in schema.py
Add invalid test csv file
Add check for empty table when getting table fields

The front-end view of this error messages looks like this:
Screenshot from 2019-07-22 15-08-47

### DIFF
--- a/multinet/multinet/__init__.py
+++ b/multinet/multinet/__init__.py
@@ -4,6 +4,7 @@ from girder import plugin, logprint
 from girder.api import access
 from girder.api.rest import Resource, getBodyJson
 from girder.api.describe import Description, autoDescribeRoute
+from girder.exceptions import RestException
 
 import csv
 from io import StringIO
@@ -80,6 +81,7 @@ class MultiNet(Resource):
         .param('workspace', 'Target workspace', required=True)
         .param('table', 'Target table', required=True)
         .param('data', 'CSV data', paramType='body', required=True)
+        .errorResponse()
     )
     def bulk(self, params, workspace=None, table=None):
         """
@@ -92,6 +94,13 @@ class MultiNet(Resource):
         logprint('Bulk Loading', level=logging.INFO)
         rows = csv.DictReader(StringIO(cherrypy.request.body.read().decode('utf8')))
         workspace = db.db(workspace)
+
+        # Check for key uniqueness
+        if ('_key' in rows.fieldnames):
+            keys = [row['_key'] for row in rows]
+            uniqueKeys = set(keys)
+            if len(keys) != len(uniqueKeys):
+                raise RestException('CSV Validation Failed')
 
         # Set the collection, paying attention to whether the data contains
         # _from/_to fields.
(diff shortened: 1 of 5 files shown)


multinet/multinet/db.py
```diff
@@ -103,7 +103,7 @@ def workspace_graph(workspace, name, arango=None):
 def table_fields(query, arango=None):
     """Return a list of column names for table `query.table` in workspace `query.workspace`."""
     workspace = db(query.workspace, arango=arango)
-    if workspace.has_collection(query.table):
+    if workspace.has_collection(query.table) and workspace.collection(query.table).count() > 0:
```
AlmightyYakob: I added this check here, because I was hitting cases where the `sample` variable had a value of `None`, and as such `sample.keys()` was throwing an error. I'm not sure if this introduces bugs or undesired behavior, but I haven't seen any adverse effects yet. @waxlamp PTAL.

multinet/multinet/__init__.py
```diff
@@ -93,6 +95,13 @@ def bulk(self, params, workspace=None, table=None):
         rows = csv.DictReader(StringIO(cherrypy.request.body.read().decode('utf8')))
         workspace = db.db(workspace)
 
+        # Check for key uniqueness
+        if ('_key' in rows.fieldnames):
+            keys = [row['_key'] for row in rows]
+            uniqueKeys = set(keys)
+            if len(keys) != len(uniqueKeys):
+                raise RestException('CSV Validation Failed')
```
waxlamp: Can you augment this message with something like "non-unique key detected"?

multinet/multinet/db.py
waxlamp: I was just bitten by this bug earlier today; thanks for finding and fixing it.

multinet/multinet/__init__.py
AlmightyYakob: Sure thing. I could also specify which key was duplicated (or in the case of multiple, the first duplicate found). However, this would add a little more time complexity. I'll leave it to your judgement whether that's a real concern and/or if the functionality is worth it.

waxlamp: That's a good idea.

I don't think the time complexity should increase though. You could just insert each key one by one into the set, and for each key that is already found there, record it as a duplicate key, then report them all at the end.

Insertion and lookup into a set should both be O(log n), and you're doing this with n values, for a total complexity of O(n log n). Currently, you're inserting all the keys into a set, which seems like an O(n log n) operation as well.

If we do go this route, we can just merge the current PR as-is and open a new issue describing this change.

Let me know what you think.

AlmightyYakob: Since Python implements sets as hash tables, I would think insertion and deletion would be O(1) (amortized). So creating a set of n elements should be O(n). Though I agree the way you outlined would have the same time complexity. I'm good with merging this and opening a new PR for that change.

waxlamp: You're right, Python uses hash tables to implement sets. Learned something new today :)

Ok, since the situation is even better than O(n log n), let's just augment the message the way I requested, and then get this merged. I'll open an issue for the enhancement. Thanks @AlmightyYakob!

AlmightyYakob: Sounds good!

PR #96, opened 2019-07-23
Specify which keys are duplicate and fix front-end bug

### LINKED ISSUE #95: List all duplicated keys during failed CSV validation
This will follow on from the feature introduced in #94.

AlmightyYakob:
Fixes #95

### DIFF
--- a/client/src/views/WorkspaceDetail.vue
+++ b/client/src/views/WorkspaceDetail.vue
@@ -94,7 +94,7 @@
             <v-card-text>
               <v-layout row wrap>
                 <v-flex>
-                  <v-text-field v-model="newTable" placeholder="name your table" solo :error-messages="tableCreationErrors"/>
+                  <v-text-field v-model="newTable" placeholder="name your table" solo :error-messages="tableCreationError"/>
                 </v-flex>
               </v-layout>
               <v-layout row wrap>
@@ -205,7 +205,7 @@ export default {
       selectedType: null,
       graphNodeTables: [],
       graphEdgeTable: null,
-      tableCreationErrors: [],
+      tableCreationError: [],
     }
   },
   computed: {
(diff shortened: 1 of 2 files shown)


multinet/multinet/__init__.py
```diff
@@ -31,6 +31,23 @@ def graphql_query(query, variables=None):
     return dict(data=result.data, errors=errors, query=query)
 
 
+def csv_validation(rows):
```
waxlamp: Let's rename this to `validate_csv` so that the function name is a verb.

client/src/views/WorkspaceDetail.vue
```diff
@@ -268,10 +268,10 @@ export default {
           },
         }
         );
-        this.tableCreationErrors = [];
+        this.tableCreationErrors = null;
         this.update()
       } catch(err) {
-        this.tableCreationErrors.push(err.response.data.message);
+        this.tableCreationErrors = err.response.data.message;
```
waxlamp: Is this the front-end bug? What was the buggy behavior?

multinet/multinet/__init__.py
AlmightyYakob: Sure thing

client/src/views/WorkspaceDetail.vue
AlmightyYakob: Yes. The bug here is I was adding to the array of errors that the v-text-field would display, assuming it would display them all. It for some reason was only displaying once. Also, I didn't really think it was necessary to add that complexity when in reality it would probably only display one error at a time, so I just made that a single message. I should probably rename that variable `tableCreationError` (singular), to clarify that.

waxlamp: Gotcha, thanks for the explanation. I agree with you about changing it to the singular.

waxlamp:
@AlmightyYakob can we delete this branch?

AlmightyYakob:
> @AlmightyYakob can we delete this branch?

Sure thing

PR #98, opened 2019-07-25
Add Graph Validation on Creation

### LINKED ISSUE #71: Add graph validation
This should be a serverside function that checks to see that all edges in a graph reference existing nodes. It should also check to see that the keys referenced per table also exist.

~It is not necessary to run such validation on graph creation (unless requested) because it is conceivable that node data could be uploaded *after* graph creation, or that in the future the system may be able to add missing nodes to a node table, etc.~ The validation need not run on edge table upload for the reasons stated in the struck-out text; however, it *should* be performed on *graph creation*.

AlmightyYakob:
Fixes #71


waxlamp:
@AlmightyYakob can we delete this branch?

AlmightyYakob:
> @AlmightyYakob can we delete this branch?

@waxlamp I was going to continue to use this branch to add the additional features of verifying that for each valid table, all referenced keys also exist within that table.

PR #100, opened 2019-07-29
Add Syntax Validation on CSV upload

AlmightyYakob:


multinet/multinet/__init__.py
```diff
@@ -47,7 +51,24 @@ def validate_csv(rows):
...
+        validCell = re.compile('[^/]+/[^/]+')
+
+        for i, row in enumerate(rows):
+            fields = []
+            if not validCell.match(row['_from']):
+                fields.append('_from')
+            if not validCell.match(row['_to']):
+                fields.append('_to')
+
+            if fields:
```
waxlamp: Is the intent here to bail with an error as soon as a single bad row is detected? It seems like we could instead collect errors for multiple lines in one go, and report them all at the end. That would be more useful for fixing multiple errors as it wouldn't require multiple upload attempts to identify them all.

multinet/multinet/__init__.py
```diff
@@ -47,7 +51,24 @@ def validate_csv(rows):
                 uniqueKeys.add(key)
 
         if (len(duplicates) > 0):
-            raise RestException(f'CSV Validation Failed: Duplicate Keys {", ".join(duplicates)}.')
+            raise RestException(f'{BASE_ERROR_MSG} Duplicate Keys {", ".join(duplicates)}.')
+    elif '_from' in rows.fieldnames and '_to' in rows.fieldnames:
+        # Edge Table, check that each cell has the correct format
+        validCell = re.compile('[^/]+/[^/]+')
```
waxlamp: Could you change `validCell` to `valid_cell`?

test/data/membership_invalid.csv
```diff
@@ -5,316 +5,3 @@ neitherhere/8,northere/2
 members/8,clubs/3
```
waxlamp: Can we change this filename to something like `membership_invalid_missing_header.csv` or something like that?

multinet/multinet/__init__.py
AlmightyYakob: Currently I'm bailing on the first found, because I'm specifying which fields are malformed. Because including both this and your suggestion would produce error messages which are too complex, I think a choice has to be made between this method of listing the fields, or listing all rows which have malformed fields (without specifying the field). I'm indifferent, so whichever you think is more useful is the way we'll go.

multinet/multinet/__init__.py
AlmightyYakob: Sure thing

test/data/membership_invalid.csv
AlmightyYakob: This diff is a bit misleading, as it doesn't show the top of the file. The top of the file contains undefined node tables, so a good name for this might be something like `membership_invalid_referenced_tables.csv`.

multinet/multinet/__init__.py
waxlamp: (we discussed this a bit IRL)

Let's go ahead and merge this as-is for now; eventually I'd like to update the API to return a list of errors while signalling an HTTP 400 code to show that something went wrong with the input.

test/data/membership_invalid.csv
waxlamp: Oops!

Let's go with `membership_invalid_nonexistent_tables.csv`

multinet/multinet/__init__.py
waxlamp: See #101.

PR #102, opened 2019-07-31
Cut over from Girder to Flask

waxlamp:
This is a big PR that we need to handle carefully. It swaps out Girder for Flask. The following changes still need to be made:
- [x] merge in all changes from master and ensure they work in the flask app
- [x] fix all versions in Pipfile
- [x] remove Girder and move the flask files into the `multinet` directory
- [x] remove Girder dependencies from Pipfile
- [x] remove Mongo from the docker-compose setup
- [x] rename the repository to "multinet"
- [x] update documentation

~Currently, if you check this branch out, re-run `pipenv install`, then run `flask run` in the top level, the flask app will run on port 5000, and the client app should run properly without any changes.~ The usual sequence of commands will now start the server and client as usual (a few steps that had to do with building the Girder client, and setting up a user, are no longer necessary).

@AlmightyYakob, I will probably want to do some pair programming to get all issues squared away before final merge.


multinet/multinet/multinet.py
```diff
@@ -0,0 +1,286 @@
...
+    """Perform a GraphQL query using optional variable definitions."""
+    result = graphql(schema, query, variables=variables or {})
+    if result:
+        errors = [error.message for error in result.errors] if result.errors else []
+        app.logger.warn('Errors in request: %s' % len(errors))
+        for error in errors[:10]:
+            app.logger.warn(error)
+    else:
+        errors = []
+    return dict(data=result.data, errors=errors, query=query)
```
AlmightyYakob: It seems here that if the else cause was reached here, then result.data would not be defined. I would propose before checking result, doing something like 
```
data = {}
errors = []
```

And then conditionally overwriting them by checking each, that way there's a default value defined. Then instead of `data=result.data`, it would just be `data=data`. Please lmk what you think

waxlamp: Thanks, good catch. Check out 9785cf8.

multinet/multinet/multinet.py
```diff
@@ -0,0 +1,290 @@
...
+        duplicates = set()
+        for key in keys:
+            if key in uniqueKeys:
+                duplicates.add(key)
+            else:
+                uniqueKeys.add(key)
+
+        if (len(duplicates) > 0):
+            return {'error': 'duplicate',
+                    'detail': list(duplicates)}
```
AlmightyYakob: I personally don't like how we're returning the error here and handling in the `bulk` method, as it's a bit arbitrary. I think the `bulk` method should be agnostic to the errors thrown here, and whatever is returned here just be minimally wrapped as a tuple and returned from the `bulk` method. However, for viewing errors in detail on the client, we'll need to have a consistent was of returning possibly multiple errors. I think we need to have an offline discussion around this.

multinet/multinet/__init__.py
```diff
@@ -1,329 +1,26 @@
...
-    else:
-        errors = []
-    return dict(data=result.data, errors=errors, query=query)
+    # Register blueprints.
+    app.register_blueprint(multinet.bp)
 
+    @app.route('/hello')
+    def hello():
+        app.logger.debug('heyo')
+        return 'hello, world'
```
AlmightyYakob: No longer needed

client/src/api.js
```diff
@@ -1,10 +1,11 @@
-import { getCurrentToken } from '@girder/core/auth';
-import { getApiRoot } from '@girder/core/rest';
 import axios from 'axios'
 
+function getApiRoot() {
+  return `${window.location.origin}/api/v1`;
+}
```
AlmightyYakob: Switch to /api/

client/vue.config.js
```diff
@@ -2,8 +2,11 @@ module.exports = {
   devServer: {
     proxy: {
       '/api/v1': {
-        target: 'http://127.0.0.1:9090',
-        changeOrigin: true
+        target: 'http://127.0.0.1:5000',
```
AlmightyYakob: Replace with env var

multinet/multinet/__init__.py
```diff
@@ -1,329 +1,26 @@
...
-from graphql import graphql
-import cherrypy
-import newick
-import uuid
 
-from .schema import schema
-from . import db
+def create_app(config=None):
+    """Create a Multinet app instance."""
+    print(__name__)
```
AlmightyYakob: Unnecessary

multinet/multinet/__init__.py
```diff
@@ -1,329 +1,26 @@
...
-    if result:
-        errors = [error.message for error in result.errors] if result.errors else []
-        logprint('Errors in request: %s' % len(errors), level=logging.WARNING)
-        for error in errors[:10]:
-            logprint(error, level=logging.WARNING)
-    else:
-        errors = []
-    return dict(data=result.data, errors=errors, query=query)
+    # Register blueprints.
+    app.register_blueprint(multinet.bp)
```
AlmightyYakob: Add mount path

multinet/multinet/multinet.py
waxlamp: (Had an offline discussion, summarizing here)

- decided to minimally wrap what is returned from this function
- decided to look closer at returning detailed error objects (whether actual class instances, or formatted dicts/json objects) to describe each error
  - the client can receive the full list and decide what to do with it (just say how many there are; display an error message for each error; etc.)

I will change the code to stop rewrapping the error objects for now, and will open an issue to look more closely at this in the future.

multinet/multinet/__init__.py
waxlamp: See 4d8f9d1

client/src/api.js
waxlamp: See f7f1524

multinet/multinet/__init__.py
waxlamp: See 5414565

client/vue.config.js
waxlamp: See 2c78087

multinet/multinet/__init__.py
waxlamp: See 1abc81b

multinet/multinet/multinet.py
waxlamp: See 6694ab5 and #105

multinet/multinet/__init__.py
```diff
@@ -1,329 +1,29 @@
...
+    app.register_blueprint(multinet.bp, url_prefix='/multinet')
 
+    @app.route('/about')
+    def about():
+        return (
+            '<h1>Multinet API</h1>'
+            '<div>'
+            '    See <a href="https://multinet.app">Multinet website</a> for details.'
+            '</div>'
+        )
```
AlmightyYakob: I'm not able to reproduce the same error when removing this. That is, after removing this method, my flask application still runs fine. Is there a specific test case for this, or did the whole application fail to run?

client/vue.config.js
```diff
@@ -1,6 +1,9 @@
 // Read in .env file.
 const process = require('process');
-require('dotenv').config();
```
waxlamp: I thought about doing this but I didn't like it for some reason. Maybe that it makes the NPM package refer outside itself to pick up configs, but I don't think that's really a problem since you generally have to build the client in context of this whole repo.

Thanks for this change.

waxlamp:
I reverted your quotation mark commit and instead kept the multiline string, but used double quotes to delimit it.

flake8 was complaining about the use of triple single-quotes, rather than the multiline string itself. I prefer the multiline string because it's far more readable than the concatenated single-line strings.

multinet/multinet/__init__.py
waxlamp: I got a "missing Flake application" sort of error upon hitting the API after startup. Startup itself showed no errors.

Let's leave this as is and get rid of it if we can figure out what, if anything, is wrong.

PR #118, opened 2019-08-14
Add http error 500 on database connection problem

### LINKED ISSUE #103: Raise an HTTP error if Arango server not running

JackWilb:
Is this all you had in mind for this fix, @waxlamp ? Or should I implement something in the frontend that would alert the user? (closes #103)


multinet/multinet/multinet.py
```diff
@@ -52,7 +55,13 @@ def _graphql():
     app.logger.debug("request: %s" % query)
     app.logger.debug("variables: %s" % variables)
 
-    result = graphql_query(query, variables)
+    try:
+        result = graphql_query(query, variables)
+    except ValueError as e:
+        for arg in e.args:
+            app.logger.error(arg)
+        abort(500)
```
waxlamp: I don't think this is quite right: if the query succeeds in executing, but encounters problems, it's likely to be an issue with the input, and this should probably be a 400 rather than a 500.

multinet/multinet/multinet.py
```diff
@@ -30,6 +30,9 @@ def graphql_query(query, variables=None):
             if excess > 0:
                 app.logger.error(f'{excess} more error{"s" if excess > 1 else ""}')
 
+            raise ValueError("Error when connecting to the DB.")
+            return
```
waxlamp: This `return` seems to not be necessary, owing the the raised exception in the line above.

multinet/multinet/multinet.py
```diff
@@ -30,6 +30,9 @@ def graphql_query(query, variables=None):
             if excess > 0:
                 app.logger.error(f'{excess} more error{"s" if excess > 1 else ""}')
 
+            raise ValueError("Error when connecting to the DB.")
```
waxlamp: I don't think `ValueError` is the right choice for this. `ValueError`s are meant for cases when the wrong range of values is presented to a function (like a square root function receiving a negative argument).

I'm also not sure it's necessary to raise an exception here in the first place: the `errors` value in the result will already flow back to the caller, which can then signal the appropriate status code and send the errors back to the client.

multinet/multinet/multinet.py
JackWilb: Removed the return.

multinet/multinet/multinet.py
JackWilb: Agreed. Remove the ValueError in defcb68

JackWilb:
@waxlamp Can you please review this new direction and tell me if you think this is a better way to handle the database being down? If it is, there will be some more edits to expand this methodolgy through the db.py and query.py files

multinet/multinet/db.py
```diff
@@ -1,11 +1,30 @@
 """Low-level database operations to fulfill GraphQL queries."""
 import os
+import sys
```
waxlamp: Remove this.

multinet/multinet/db.py
```diff
@@ -1,11 +1,30 @@
...
 from arango import ArangoClient
+from requests.exceptions import ConnectionError
+
+from flask import current_app as app
 
 from multinet.types import Row, Entity, EntityType, Cursor
 
 
+def test_db(arango):
+    app.logger.info("in test_db")
```
waxlamp: Let's get rid of this line.

multinet/multinet/db.py
```diff
@@ -1,11 +1,30 @@
 """Low-level database operations to fulfill GraphQL queries."""
 import os
+import sys
 
 from arango import ArangoClient
+from requests.exceptions import ConnectionError
+
+from flask import current_app as app
```
waxlamp: If this is only needed for the logging line below, let's get rid of this as well.

multinet/multinet/db.py
```diff
@@ -1,11 +1,30 @@
...
+def test_db(arango):
+    app.logger.info("in test_db")
+
+    try:
+        arango.db("_system",
+                  username="root",
+                  password=os.environ.get("ARANGO_PASSWORD", "letmein")
+                  ).has_database('test')
+    except ConnectionError:
+        app.logger.error("db connection issues")
```
waxlamp: Get rid of this line.

multinet/multinet/db.py
```diff
@@ -52,12 +71,15 @@ def delete_workspace(name, arango=None):
 @with_client
 def get_workspaces(name, arango=None):
     """Return a list of all workspace names."""
-    sys = db("_system", arango=arango)
-    if name and sys.has_database(name):
-        return [name]
+    if test_db(arango):
+        sys = db("_system", arango=arango)
```
waxlamp: Let's change this variable from `sys` to something less ambiguous (so that it doesn't conflict with the Python standard module of the same name).

multinet/multinet/multinet.py
waxlamp: Can you change this to a 400?

Also, instead of using `abort()` I think it's better to use the ordinary return convention that Flask supports. That would be something like `return (result, 400)`. Since the next line returns a successful response, you could combine them like this:

```python
status_code = 400 if result["errors"] else 200
return (result, status_code)
```

waxlamp: Also, since this location isn't where the DB connection issue is going to be detected, we may just want to revert the changes to this file altogether, or handle them in a separate PR.

multinet/multinet/resolvers/query.py
```diff
@@ -5,11 +5,14 @@
...
-        if not name or workspace == name
-    ]
+    if db.get_workspaces(name) is not None:
+        return [
+            workspace
+            for workspace in db.get_workspaces(name)
+            if not name or workspace == name
+        ]
+    else:
+        return
```
waxlamp: This would be the place to raise an error if the database is not live. The way you have it now, you would want to replace this with something like `return (None, "500 Database Not Running")`.

multinet/multinet/db.py
```diff
@@ -1,11 +1,30 @@
...
 
 from arango import ArangoClient
+from requests.exceptions import ConnectionError
+
+from flask import current_app as app
 
 from multinet.types import Row, Entity, EntityType, Cursor
 
 
+def test_db(arango):
```
waxlamp: Let's use a more descriptive function name, something like `db_required`.

Then this function could be used as a decorator for functions that require the database, and it could automatically return the right status code if the database isn't running.

Or, perhaps more likely, this function could be called within `with_client()`, since that decorator already implies the idea of "database required".

waxlamp:
@JackWilb: by the way, the structure of the project has changed a bit, so you'll likely want to rebase this on master, or merge master in. See #117 for details on how files moved around.

multinet/multinet/db.py
JackWilb: Removed.

multinet/multinet/db.py
JackWilb: Removed.

multinet/multinet/db.py
JackWilb: Removed.

multinet/multinet/db.py
JackWilb: Removed.

multinet/multinet/db.py
JackWilb: This is what was in the file before and is in lots of other places. Should this be a new issue @waxlamp ?

multinet/multinet/db.py
JackWilb: I changed the name to the one you suggested. Are you imagining that the db_required function will live inside the with_client() function and then we'd call it in the wrapper function? If we're good, call fun(), otherwise just return?

multinet/multinet/db.py
waxlamp: Ah you're right. Yes, let's open an issue to fix that.

multinet/multinet/db.py
waxlamp: I think so, but rather than "just return", the cool thing would be to raise the 500 error at that point. I think that could just be done with `abort()` since there's no data payload in that case, but I'm also not entirely certain that will work, since we're not obviously within a flask context there.

JackWilb: Aborting there is giving the same graphql error where it prints the whole stack trace and is un-catchable. I think we should `return None` here and then in multinet.py if the data is None, we can `abort(500)`. That way if the db is up, we don't get a 500 and if the db is down we do. What do you think?

waxlamp: I want to avoid doing this so we don't have to check for `None` in a bunch of places throughout the API. Let me think about this a bit.

waxlamp: I want to avoid doing this so we don't have to check for None in a bunch of places throughout the API. Let me think about this a bit.

waxlamp: Ok, so I think I've figured this out.

We don't want to check if the DB is live on every single DB operation (which is what we'd be doing if we rolled this check into `with_client`); instead we just want to check once on any endpoint that will engage the DB. So I think `db_required` should become a decorator that performs the DB liveness test, does `return (500, "Database Not Live")` if the test fails, and runs the wrapped function if the test passes. Since the context is a Flask route, we should be able to do this. And since that liveness testing operation requires the use of the database, it can be wrapped in the `with_client` decorator.

So in the end, our mistake was in thinking that `with_client` should always include `db_required`, whereas it's actually the other way around.

Let me know what you think of this.

multinet/multinet/multinet.py
JackWilb: Opened #127

multinet/multinet/db.py
JackWilb: Opened #127

JackWilb:
I'm going to amend my solution to match the new strategy on Monday. I'll alert you when it's ready for review.

multinet/multinet/db.py
JackWilb: Edited the blueprints to run the require_db function before each route.

multinet/multinet/resolvers/query.py
JackWilb: We decided to move the error handling to the multinet.py file

JackWilb:
This is ready for review again, @waxlamp.

multinet/multinet.py
```diff
@@ -11,13 +11,24 @@
 bp = Blueprint("multinet", __name__)
 
 
+def require_db():
+    """Check if the db is live."""
+    if db.check_db():
+        return
+    else:
+        return Response(None, "500 Database Not Live")
```
waxlamp: We can simplify this block to

```python
if not db.check_db():
    return (None, "500 Database Not Live")
```

multinet/multinet.py
```diff
@@ -11,13 +11,24 @@
...
+
+
 def graphql_query(query, variables=None):
     """Perform a GraphQL query using optional variable definitions."""
     data = None
     errors = []
 
     result = graphql(schema, query, variables=variables or {})
-    if result:
+    if result.data:
```
waxlamp: What was the reason for this change?

multinet/multinet.py
```diff
@@ -53,6 +64,7 @@ def _graphql():
     app.logger.debug("variables: %s" % variables)
 
     result = graphql_query(query, variables)
+
```
waxlamp: What was the reason for this change?

multinet/resolvers/query.py
```diff
@@ -5,11 +5,14 @@
...
-        if not name or workspace == name
-    ]
+    if db.get_workspaces(name) is not None:
+        return [
+            workspace
+            for workspace in db.get_workspaces(name)
+            if not name or workspace == name
+        ]
+    else:
+        return
```
waxlamp: What was the reason for this change?

multinet/multinet.py
```diff
@@ -2,7 +2,7 @@
 from graphql import graphql
 import json
 
-from flask import Blueprint, request
+from flask import Blueprint, request, Response
```
waxlamp: Shouldn't need this if we use the tuple form (as shown in my comment below).

JackWilb: Without the response function we get this error: `TypeError: The view function did not return a valid response. The function either returned None or ended without a return statement.`

multinet/multinet.py
JackWilb: This was from one of my previous revision. I rolled it back to the original code (this is the same as the current code in master)

multinet/resolvers/query.py
JackWilb: It was a remnant from earlier attempts. I removed it.

multinet/multinet.py
JackWilb: Fixed in the next commit.

multinet/multinet.py
JackWilb: There's no reason for this extra line. Removed in the next commit

multinet/multinet.py
waxlamp: Ah, I thought you could send an empty response, but it's not allowed (so, I'm not actually sure why wrapping `None` in `Response` works, but it probably will causes headaches down the line).

Instead of `None`, passing an empty string actually does work, so let's go with that.

JackWilb: Fixed in the later commit.

PR #124, opened 2019-08-16
Change file input component to v-file-input

jtomeck:
@waxlamp - JS needs to be moved from FileInput.vue into WorkspaceDetail.vue.


jtomeck:
Looks good, thanks Roni!

PR #131, opened 2019-08-16
Add proper flask response to create_graph function

### LINKED ISSUE #129: create_graph function returns None on duplicate graph creation
The `create_graph` function is an endpoint exposed by Flask, and in it's current implementation returns `None` if the `db.create_graph` function returns a falsy value. This should be returning a well formed Flask response if the graph creation failed.

https://github.com/multinet-app/multinet/blob/8f6f08c59c1562ec34be98924a9dfe1c97bdc498/multinet/multinet.py#L110

AlmightyYakob:
Fixes #129

### DIFF
--- a/multinet/multinet.py
+++ b/multinet/multinet.py
@@ -107,4 +107,7 @@ def create_graph(workspace, graph):
         error = {"error": "undefined_tables", "detail": list(invalid_from | invalid_to)}
         return (error, "400 Edge Table Validation Failed")
 
-    return graph if db.create_graph(workspace, graph, node_tables, edge_table) else None
+    if db.create_graph(workspace, graph, node_tables, edge_table):
+        return graph
+    else:
+        return (f"Graph {graph} already exists", 400)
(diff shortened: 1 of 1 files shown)


multinet/multinet.py
```diff
@@ -107,4 +107,7 @@ def create_graph(workspace, graph):
         error = {"error": "undefined_tables", "detail": list(invalid_from | invalid_to)}
         return (error, "400 Edge Table Validation Failed")
 
-    return graph if db.create_graph(workspace, graph, node_tables, edge_table) else None
+    if db.create_graph(workspace, graph, node_tables, edge_table):
+        return graph
+    else:
+        return (f"Graph {graph} already exists", 400)
```
waxlamp: I think it would be better to return `graph` (to identify the problem), and use `"400 Graph Already Exists"` for the status code. (We could also consider using [409](https://www.restapitutorial.com/httpstatuscodes.html) for the error code, which might be good for helping the client disambiguate what's going on.)

Issue #105, opened 2019-08-01
Design a better type system and protocol for validation errors


waxlamp:

JackWilb:
I think this is done, right, @waxlamp ? We added in python typing and we're using typescript for the multinetjs. Is there any additional typing we can add to the vue components?

waxlamp:
This issue has more to do with designing better informational object types to return from the uploader validation routines. I don't believe this is complete yet.

waxlamp:
I believe @AlmightyYakob's (relatively) recent work on this means this issue should be closed.

PR #132, opened 2019-08-18
Use db iterator directly in graph creation

### LINKED ISSUE #111: Use db iterator directly in edge table validation
The code in question: https://github.com/multinet-app/multinet/blob/delete-endpoints/multinet/multinet/multinet.py#L90-L94

See discussion from #109 (https://github.com/multinet-app/multinet/pull/109#discussion_r312141814).

AlmightyYakob:
Fixes #111. This PR also introduces new code into the create_graph function. This code was intended to be merged into master before the move away from Girder, but got lost in the change. This also returns a response in direct contrast to #105. This is just temporary until that issue is resolved.

### DIFF
--- a/multinet/multinet.py
+++ b/multinet/multinet.py
@@ -128,8 +128,7 @@ def create_graph(workspace, graph):
 
     # TODO: Update this with the proper JSON schema
     if errors:
-        payload = {"error": "Graph Validation Failed", "detail": errors}
-        return (payload, 400)
+        return (errors, "400 Graph Validation Failed")
 
     if db.create_graph(workspace, graph, node_tables, edge_table):
         return graph
--- a/multinet/multinet.py
+++ b/multinet/multinet.py
@@ -72,6 +72,7 @@ def delete_workspace(workspace):
 @bp.route("/workspace/<workspace>/graph/<graph>", methods=["POST"])
 def create_graph(workspace, graph):
     """Create a graph."""
+
     # Get parameters from request body.
     body = request.data.decode("utf8")
     try:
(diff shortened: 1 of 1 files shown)


multinet/multinet.py
```diff
@@ -86,26 +87,49 @@ def create_graph(workspace, graph):
...
-            if edge["_to"].split("/")[0] not in node_tables
-        ]
-    )
-
-    if invalid_from or invalid_to:
-        error = {"error": "undefined_tables", "detail": list(invalid_from | invalid_to)}
-        return (error, "400 Edge Table Validation Failed")
+    loadedWorkspace = db.db(workspace)
+    if loadedWorkspace.has_graph(graph):
+        return (graph, "409 Graph Already Exists")
```
AlmightyYakob: Here I've added an early check for an existing graph of the same name. Since the workspace is loaded anyway, there's no inefficiency here, and can avoid unnecessary work. You can see that the same response is returned below, if the `db.create_graph` function fails. I wasn't sure if it was a good idea to get rid of this check entirely, as I'm not sure if the operation would fail for any other reason, but we are checking the condition already here. There's also the consideration of which error we want the user to see first, that the graph already exists, or that the graph is invalid. I'm interested to hear your input @waxlamp

multinet/multinet.py
```diff
@@ -86,26 +87,47 @@ def create_graph(workspace, graph):
...
+        if len(nonexistent_keys) > 0:
+            errors.append(
+                f"Nonexistent keys {', '.join(nonexistent_keys)} "
+                f"referenced in table: {table}"
+            )
+
+    # TODO: Update this with the proper JSON schema
+    if errors:
+        payload = {"error": "Graph Validation Failed", "detail": errors}
+        return (payload, 400)
```
waxlamp: I think the payload should be the list of errors, and the status code should be changed to `"400 Graph Validation Failed"`.

multinet/multinet.py
```diff
@@ -86,26 +87,47 @@ def create_graph(workspace, graph):
...
+            if table not in existing_tables:
+                invalid_tables.add(table)
+            elif table in valid_tables:
+                valid_tables[table].add(key)
+            else:
+                valid_tables[table] = {key}
+
+    if invalid_tables:
+        for table in invalid_tables:
+            errors.append(f"Reference to undefined table: {table}")
```
waxlamp: These should be replaced with structured dicts, or Python class instances of some sort. Could you open an issue about this (or reference an existing issue if there is one)?

multinet/multinet.py
```diff
@@ -86,26 +87,47 @@ def create_graph(workspace, graph):
...
+
+    loadedWorkspace = db.db(workspace)
+    existing_tables = set([x["name"] for x in loadedWorkspace.collections()])
+    edges = loadedWorkspace.collection(edge_table).all()
+
+    # Iterate through each edge and check for undefined tables
+    valid_tables = dict()
+    invalid_tables = set()
+    for edge in edges:
+        nodes = (edge["_from"].split("/"), edge["_to"].split("/"))
```
waxlamp: Do we need to validate that these are in the right format? If one of these reads something like `"/123"`, what would happen then?

waxlamp: Answer: we don't need to here, because `validate_csv()` in [the CSV uploader](https://github.com/multinet-app/multinet/blob/use_db_iterator/multinet/uploaders/csv.py#L32) does it when the table is created.

multinet/multinet.py
waxlamp: This is good because it's more explicit. We should get rid of the second check below this line, as it's redundant with the check here.

The "graph already exists" error should show up first, since that has to be fixed before anything else can happen. The user workflow would be to receive an error in the client that the graph name already exists, then select a different name, then try again, then possibly receive validation errors, fix those, and finally succeed in creating the graph.

multinet/multinet.py
AlmightyYakob: Dependent on #105

multinet/multinet.py
AlmightyYakob: Alright, I'll remove that second check.

PR #137, opened 2019-08-20
Add basic testing infrastructure and initial test

AlmightyYakob:
Note: This PR removes Tox. This is because it was hindering the use of Pipenv, and wasn't really providing much benefit to this project specifically.


AlmightyYakob:
@waxlamp the tests are failing due to the `test/data/table2csv.py` file. Does this belong in scripts?

waxlamp:
@AlmightyYakob, I fixed the docstring errors in `table2csv.py` but yeah that script and `PaulRevereAppD.csv` could possibly be placed elsewhere, but not in the scripts directory. For now let's just leave them where they are.

waxlamp:
I experimented with splitting apart the compound assertions in the test module. Let me know if there's a good reason not to do that, and then we can revert that commit.

AlmightyYakob:
> I experimented with splitting apart the compound assertions in the test module. Let me know if there's a good reason not to do that, and then we can revert that commit.

That's a good change, not really sure why I had it the original way tbh.

AlmightyYakob:
> This looks great--could you just add something to the documentation to explain how to run the testing and coverage?

@waxlamp I've done that, lmk what you think

PR #138, opened 2019-08-21
Remove preceding 'multinet' on url in createTable

AlmightyYakob:


JackWilb:
I'm having issues creating graphs after this update. I keep getting 404ed when hitting create

AlmightyYakob:
> I'm having issues creating graphs after this update. I keep getting 404ed when hitting create

That's a separate but very similar issue, which is addressed by #140

PR #140, opened 2019-08-21
Fix graph creation

AlmightyYakob:
One of the changes here wraps the `errors` list with a dict. I'm not sure how it got past us before, but at least on my local, when creating a graph (after fixing the issue addressed on the client) the server throws an error, stating that lists can't be directly returned. Lmk if you think there's a better way to deal with this @waxlamp.

### DIFF
--- a/client/src/components/GraphDialog.vue
+++ b/client/src/components/GraphDialog.vue
@@ -112,7 +112,7 @@ export default {
   methods: {
     async createGraph () {
       const { workspace, newGraph } = this;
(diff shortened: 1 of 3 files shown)


waxlamp:
The other way to handle this is to convert the list to JSON and return that string. This is implicitly what happens when you return a dict as well. The main thing is for us to design and document the API in the most efficient way we can.

I like just returning a list if the main value we're dealing with is in fact a list. Of course, I also like the idea of using a dict to sort of "describe" what's being returned. We'll need to discuss what to standardize on in the end. For now, though, I think this is fine.

PR #146, opened 2019-08-22
Fix pre-commit hooks to be in sync with CI

AlmightyYakob:
This solves the issue we had discussed, regarding the pre-commit hooks and the CI tests being out of sync. This also removes the need for the black hook entirely, as we were only running it with `--check`, and that's what flake8-black does. You'll notice there's an extra `excludes` field being used, which uses regex. Ideally, pre-commit would read this from the `.flake8` , but the flag that controls this is actually ignored, so this is necessary. However, the hook does read the other options from `.flake8`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,4 +5,6 @@ repos:
     - id: flake8
       additional_dependencies: ['flake8-black', 'flake8-bugbear', 'flake8-docstrings', 'flake8-quotes']
       args: [--config=.flake8]
-      exclude: (__pycache__)|(build)|(^scripts)
+      # Need to have this both here and in flake8, because pre-commit
+      # doesn't read the `exclude` config option in .flake8
+      exclude: (__pycache__)|(^build)|(^scripts/)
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -1,12 +1,8 @@
...
-    - id: black
-      language_version: python3.7
-      args: [--check]
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v1.2.3
     hooks:
     - id: flake8
+      additional_dependencies: ['flake8-black', 'flake8-bugbear', 'flake8-docstrings', 'flake8-quotes']
       args: [--config=.flake8]
+      exclude: (__pycache__)|(build)|(^scripts)
```
waxlamp: In order to *exactly* match what's in the .flake8 file, I think each term should be surrounded by `^` and `$`.

waxlamp: Can you also add a comment line right above this to explain why we're having to do this both here and in the config file?

AlmightyYakob: Discussed IRL: The regex in the `exclude` field behaves weirdly, so adding the anchor doesn't match the desired files correctly.

PR #148, opened 2019-08-22
Replace GraphQL with AQL-based REST API

waxlamp:
This works with the client application, but needs some serious review, cleanup, and in-place updates.

Some TODOs:
- [ ] overhaul the form of all endpoints, and their return types, to be as consistent as possible
- [ ] Eliminate GraphQL support code from db.py
- [ ] Move as much as possible from db.py directly into the view functions in api.py
- [ ] Try to use more AQL in implementing these endpoints
- [ ] Document the API in the Sphinx docs
- [ ] Document the API via Swagger
- [ ] Write tests

Some of these may just become new issues, in which case they'll be considered done for the purposes of the TODO list above.

@AlmightyYakob and @JackWilb, please take a look.


multinet/api.py
```diff
@@ -22,76 +31,168 @@ def require_db():
...
-        data = result.data
 
-        if errors:
-            app.logger.error("Errors in request: %s" % len(errors))
-            for error in errors[:10]:
-                app.logger.error(error)
+def lookup_workspace(workspace):
+    """Retrieve a single workspace, with a status code."""
+    result = db.get_workspace(workspace)
+    return (workspace, 404 if result is None else 200)
```
AlmightyYakob: It seems here we're just returning the name of the workspace, and not the result itself. I think here we should do something like:
```
result = db.get_workspace(workspace) or None
status = 200 if result else 404
return (result, status)
```
@waxlamp Lmk what you think

JackWilb:
db.py contains this line `from multinet.types import Row, Entity, EntityType, Cursor` but the types.py has been deleted

AlmightyYakob:
> db.py contains this line `from multinet.types import Row, Entity, EntityType, Cursor` but the types.py has been deleted

I've also noticed this, and the file still uses those types all over. However, none of the functions that use these types seem to be called anywhere. Specifically, the functions (all within `db.py`):
```
nodes
edges
fetchRows
countNodes
countEdges
graph_node_types
graph_edge_types
source
target
outgoing
incoming
```
all use these types but are never called. So if all of the behavior of these functions has been replicated somewhere else, I think we can remove them.

multinet/api.py
waxlamp: If `result` is `None` then you still want to return the name of the workspace as the response, just with the 404 code indicating which workspace was Not Found. The target use case here is if you sent out a bunch of requests in parallel for workspaces, then when they resolve eventually you'll want to make sure you know which one a 404 code is talking about.

waxlamp:
Yeah, one of the things left to do on this branch is to delete all of that dead code, something I'll do in my next pass over db.py.

(Also, in general, if you're commenting on a specific line, it's best to put the comment on that line so that we can have individual threads about these issues.)

PR #186, opened 2019-10-18
Fix transitions falling under v-slot:activator usages

waxlamp:
UPDATE: the bug is actually fixed in a newer version of Vuetify, so the fix here is to update to that version.

@jtomeck: could you please confirm that the application still behaves well under the new version (2.1.5)?

~@subdavis helped me figure this out: there's a bug in Vuetify that
prevents v-slot:activator components from taking a functional component
as the top level of their <template> argument.~

~The problem is in the following lines of code:
https://github.com/vuetifyjs/vuetify/blob/master/packages/vuetify/src/mixins/activatable/index.ts#L144-L152.
Functional components do not have a `mixins` property, and so this
if-block codes them as `HTMLElement`s, which leads to the problem.~

~@jtomeck, can you build this branch and confirm that the tooltips are working now? Can you also point out any that I missed?~

~I'm leaving this in draft mode until I can file a proper bug report with Vuetify.~


waxlamp:
Just FYI, I opened a Vuetify bug report about this: https://github.com/vuetifyjs/vuetify/issues/10037

PR #192, opened 2019-10-23
New table button improvement

jtomeck:
The button has been moved over, as well as some of the scripts for the button and dialog to work properly (although the script moving is not currently working properly). @waxlamp 

**To do:**
- [x] Figure out how to add GraphDialog into ItemPanel
- [x] Fix the scripts so that the dialogs work like they're supposed to


waxlamp:
@jtomeck, this looks good to me now. Since you opened this, I can't request you as a reviewer, but take a look at my changes and let me know if the app is behaving properly, and I'll provide the review, and you can press the merge button :)

waxlamp:
@subdavis, I could use some advice on this PR. There are two distinct problems:
1. The "grandchild" issue is appearing here: I need a dialog to signal success to the parent component of its parent component, and I've opted to just thread an event emission up through the genealogy. Is there a better way in this case, short of Vuex?
2. I also need to select either `<table-dialog>` or `<graph-dialog>` based on which instance of `<item-panel>` it is, and I'm doing so in a decidedly kludgey sort of way. Can you take a quick look and see if there's something better we could be doing? Maybe a slot of some kind?

waxlamp:
I approved this but I'd like you to verify that everything seems good to you as well before you merge.

subdavis:
@waxlamp for both those points, I think I would push you toward a slot.  Not even a scoped slot, because I don't see anything that ItemPanel provides that doesn't come directly from the parent.  I'm inclined to push you in that direction primarily because I see `nodeTables` and `edgeTables` are "optional" parameters that only get passed if the graph-dialog is created.

I would drop the following props from ItemPanel: `workspace, nodeTables, edgeTables`

```html
<item-panel
              title="Tables"
              :items="tables"
              :workspace="workspace"
              route-type="table"
              icon="table_chart",
            />
  <table-dialog
        :workspace="workspace"
        @success="update"
      />
  <!-- OR graph-dialog -->
```

There are also ways you can clean up the routing to avoid absolute routes and further eliminate the `workspace` and `route-type` props (I just learned of the perils of absolute routes myself) but that's another discussion.

waxlamp:
Thanks @subdavis! I will try out the slot idea, and follow up with you about routing!

PR #193, opened 2019-10-23
Use multinetjs in client application

waxlamp:
This PR does a bunch of coordinated things.

1. Replaces direct API calls in the client application with equivalent calls into the multinet client library.
2. Refines the multinetjs client library function signatures.
3. Updates the multinetjs test suite to reflect function signature changes.
4. Adds publishing infrastructure to the multinetjs directory (and, to support (1) above, also published [multinet 0.1.0](https://www.npmjs.com/package/multinet) to NPM.

@jtomeck, could you verify that the client application is working as it did before?

@jeffbaumes and @JackWilb, could you look at the library API and see if it makes sense? Please note that I have not folded in all of the suggestions I've gotten from you in meatspace--Jeff, I haven't created separate "all node" and "all edge" functions, and Jack, I haven't written that utility function that will serially retrieve all the nodes of a graph or table. Planning to add those as needed in followup PRs.

### DIFF
--- a/client/src/components/WorkspaceDialog.vue
+++ b/client/src/components/WorkspaceDialog.vue
@@ -60,7 +60,7 @@
 <script lang="ts">
 import Vue from 'vue';
 
-import api, { apix } from '@/api';
+import api from '@/api';
 
 export default Vue.extend({
   data() {
@@ -75,7 +75,7 @@ export default Vue.extend({
   methods: {
     async create() {
       if (this.newWorkspace) {
-        const response = apix.createWorkspace(this.newWorkspace);
+        const response = api.createWorkspace(this.newWorkspace);
 
         if (response) {
           this.$router.push(`/workspaces/${this.newWorkspace}`);
--- a/client/src/components/WorkspaceDialog.vue
+++ b/client/src/components/WorkspaceDialog.vue
@@ -60,7 +60,7 @@
 <script lang="ts">
 import Vue from 'vue';
 
-import api from '@/api';
+import api, { apix } from '@/api';
 
 export default Vue.extend({
   data() {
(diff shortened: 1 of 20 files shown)


client/src/components/WorkspaceDialog.vue
```diff
@@ -75,7 +75,7 @@ export default Vue.extend({
   methods: {
     async create() {
       if (this.newWorkspace) {
-        const response = await api().post(`/workspaces/${this.newWorkspace}`);
+        const response = api.createWorkspace(this.newWorkspace);
```
JackWilb: Should we await here? Looks like this function returns a promise

docker-compose.yaml
```diff
@@ -2,7 +2,6 @@ version: '3'
 services:
   arangodb:
     image: arangodb/arangodb:3.4.3
-    container_name: arangodb
```
JackWilb: Why did you remove this? It makes the name much longer and you changed the name of the test one

client/src/components/WorkspaceDialog.vue
waxlamp: Good catch. If I had explicitly typed the `response` variable, typescript would have caught this for me :open_mouth:

Fixed in 1a82a7d76c7ccc058e42a1dbd85ce7c4ff1f639d.

docker-compose.yaml
waxlamp: Including the container name overrides the use of the `-p` flag. Even when I supplied a name for the testing container, for some reason that I don't fully get, starting that container would kill the dev container. In the end, not setting the container name this way and letting the system autogenerate names based on `-p` was the cleanest way to go.

JackWilb: Ok, works for me

PR #197, opened 2019-11-04
Add end to end testing with puppeteer

### LINKED ISSUE #195: Add end to end tests for the client app functionality
Test creating and deleting workspaces, table, and graphs. Use the tape testing library with puppeteer to run in headless chrome

JackWilb:
Closes #195 

TODO:
- [x] Decide on testing framework (tape vs jest). We settled on tape.
- [x] Write tests


waxlamp:
Rather than include a lot of tests in this PR, I think it's a huge step just to put down the infrastructure to enable testing via puppeteer. (Future PRs can be smaller, aimed at adding batches of tests, or just individual ones.)

If you think this is ready for review on that basis, go ahead and mark it as ready and we can go over it and get it merged.

JackWilb:
> If you think this is ready for review on that basis, go ahead and mark it as ready and we can go over it and get it merged.

Let me write a couple more checks. I want to check the things that shouldn't work in the pieces I've already written tests for

JackWilb:
This is ready for review @waxlamp. There are things that are tested implicitly, without writing assertions such as clicking in the workspace-name box but this does the job.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.setViewport({ width, height });
+    await p.goto("http://127.0.0.1:8080/");
+    
+    // Act: Test creating a workspace
+    await p.waitForSelector("#add-workspace");
+    await p.click("#add-workspace")
+    await p.waitForSelector("#workspace-name");
+    await p.focus('#workspace-name')
+    await p.keyboard.type('puppeteer')
+    await p.click("#create-workspace")
```
waxlamp: This block (and others like it) could use some spacing/formatting to indicate the separate actions being performed. In fact, I would go one further and say that definite, semantic actions like "create a workspace" should be pulled out into a function at the top of the testing module, so there isn't so much repetition of the individual actions throughout the tests.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+    // Assert: Check that the new workspace exists with no tables
+    await p.waitForSelector(".v-list-item");
+    await p.waitForSelector(".v-list-item__title");
+    const workspace_name = await p.evaluate(() => document.querySelector('.v-list-item__title').innerText);
```
waxlamp: This evaluation is repeated verbatim in the other test--I think it makes sense to abstract it into a function.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    t.equal(workspace_name, "puppeteer", "The new workspace is created and called the right thing.")
```
waxlamp: The testing descriptions should be a bit more precise. In this case, `workspace "puppeteer" is successfully created` would be perfect. This is to reflect the fact that when actually running the test, this message is basically all the user has to go on.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+    // Assert: Check that there are no tables or graphs yet
+    await p.waitForSelector(".ws-detail-empty-list");
+    let tables = await p.evaluate(() => document.querySelectorAll('.ws-detail-empty-list')[0].innerText.split("info ")[1]);
```
waxlamp: Since this is repeated *almost* verbatim below, it might make sense to create a pair of functions to actually describe what this is doing.

Furthermore, this evaluation will likely cause a crash if for some reason there were already tables in the workspace. To improve this, you can first check the `length` property of the result of `querySelectorAll()` before pulling out the content.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    t.equal(tables, "There's nothing here yet...", "The new workspace has no tables.")
+
+    let graphs = await p.evaluate(() => document.querySelectorAll('.ws-detail-empty-list')[1].innerText.split("info ")[1]);
+    t.equal(graphs, "There's nothing here yet...", "The new workspace has no graphs.")
+
+    // Act: Add a node table, an edge table, and a graph
+    await p.waitForSelector("#add-table");
+    await p.click("#add-table")
+    await p.focus('#table-name')
+    await p.keyboard.type('nodes')
```
waxlamp: Similarly as above, I would abstract this action out into a function called `addNodeTable()` or something.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    // await p.click("#file-selector")
+
+    // Assert: Check the tables and graph exist
+
+
+    // Act: Check that the tables imported correctly, the data should show what we expect
```
waxlamp: This looks incomplete--I would prefer if you could move this WIP code to a new branch based off of this one, then work on it in a future PR.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.keyboard.type('123')
+    await p.click("#create-workspace")
+    await p.click("#workspace-name", { clickCount: 3 })
+    await p.focus('#workspace-name')
+    await p.keyboard.type('++--==__')
+    await p.click("#create-workspace")
+    await p.click("#workspace-name", { clickCount: 3 })
+    await p.focus('#workspace-name')
+    await p.keyboard.type('a')
+    await p.click("#create-workspace")
```
waxlamp: Similar advice as above--this could benefit from formatting and abstracting repeated actions into functions.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.goto("http://127.0.0.1:8080/");
+    
+    // Act: Test creating invalid workspaces
+    await p.waitForSelector("#add-workspace");
+    await p.click("#add-workspace")
+    await p.waitForSelector("#workspace-name");
+    await p.focus('#workspace-name')
+    await p.keyboard.type('123')
+    await p.click("#create-workspace")
+    await p.click("#workspace-name", { clickCount: 3 })
```
waxlamp: I did a bit of googling, but couldn't figure out what the `clickCount` does. What is it for?

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.focus('#workspace-name')
+    await p.keyboard.type('++--==__')
+    await p.click("#create-workspace")
+    await p.click("#workspace-name", { clickCount: 3 })
+    await p.focus('#workspace-name')
+    await p.keyboard.type('a')
+    await p.click("#create-workspace")
+
+    // Assert: Check that the new workspace exists with no tables
+    await sleep(200)
```
waxlamp: If the purpose of this delay is to allow the server to create the workspaces and for the client to catch up, we may want to make it longer (maybe 1 second) just to be safe.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.waitForSelector(".v-list-item");
+    await p.waitForSelector(".v-list-item__title");
+    const workspace_name = await p.evaluate(() => document.querySelectorAll('.v-list-item__title')[0].innerText);
+    t.equal(workspace_name, "a", "Invalid workspaces weren't created but the last valid one was.")
```
waxlamp: I am not entirely sure this test really does what the description states. I know you mentioned you wanted to keep some of this implicit, but this test specifically looks at the first workspace in the list and makes sure it's the only valid one from the act sequence before. What happens if the bad workspaces were created and appear below the first position?

I think a more comprehensive test here will not take up too much more code or mind space: instead of pulling just the first workspace and examining its name, it's about as simple to write a function that returns all the workspace names. Then, the test can assert that `a` is present but the badly-named workspaces are not. That would also make the test descriptions more precise and informative.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+    await p.waitForSelector("#add-table");
+    await p.click("#add-table")
+    await p.focus('#table-name')
+    await p.keyboard.type('nodes')
+    // await p.click("#file-selector")
+
+    // Assert: Check the tables and graph exist
+
+
+    // Act: Check that the tables imported correctly, the data should show what we expect
```
waxlamp: Same advice as above: pull this stuff into another branch since it's not completed here.

client/vue.config.js
```diff
@@ -5,9 +5,10 @@ const dotenv = require('dotenv');
 
 const VuetifyLoaderPlugin = require('vuetify-loader/lib/plugin');
 
-// Read in .env file.
+// Read in the local env and the .env file.
+const local_env = process.env;
```
waxlamp:
```suggestion

```

client/vue.config.js
```diff
@@ -5,9 +5,10 @@ const dotenv = require('dotenv');
...
 const env = dotenv.parse(fs.readFileSync(path.resolve('..', '.env')));
-process.env.FLASK_SERVE_PORT = env.FLASK_SERVE_PORT || 5000;
+process.env.FLASK_SERVE_PORT = local_env.FLASK_SERVE_PORT || env.FLASK_SERVE_PORT || 5000;
```
waxlamp:
```suggestion
process.env.FLASK_SERVE_PORT = process.env.FLASK_SERVE_PORT || env.FLASK_SERVE_PORT || 5000;
```

client/vue.config.js
waxlamp: (This and the following suggestion are just to make the code shorter and easier to understand on first read.)

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+function browser(width, height) {
+    return puppeteer.launch({
+    headless: true,
+    args: [`--window-size=${width},${height}`],
+    // slowMo: 20 // For testing
+    });
+}
+
+test('e2e-client-test-valid-actions', async (t) => {
```
waxlamp: The name of the test can and should be human readable. Can you use a descriptive natural language phrase here?

It's also worth considering whether to meld all the assertions into a single test. I'm not sure "valid" vs. "invalid" is a super useful distinction for client testing.

client/test/e2e.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+    // Act: Check that the tables imported correctly, the data should show what we expect
+
+
+    await b.close();
+
+    t.end();
+});
+
+test('e2e-client-test-invalid-actions', async (t) => {
```
waxlamp: Natural language test name here too.

client/test/e2e.test.js
JackWilb: It clicks 3 times. 1 click focuses the input element, 2 clicks select the word, 3 clicks select the whole string in the input element.

client/test/e2e.test.js
JackWilb: It allows the modal to close. Without it, it would pass in the slowmo non-headless version but fail in the fast headless mode. I think the selector is rendered and available but it technically can't be clicked because the modal is still taking over the whole screen so this wait just lets it close. We could look at making the delay longer but the test will fail if it takes more than 5 seconds (although that's configurable).

client/test/e2e.test.js
JackWilb: Fixed in 9f65f3d

client/test/e2e.test.js
waxlamp: Ah I see! Is that because, on a failed workspace creation, the bad workspace name is still in the form when we open it again? When that's not the case, just one click will do right?

client/test/e2e.test.js
waxlamp: Gotcha. If the delay is just to let the modal close, then 200ms is fine (proven by the fact that it works in practice).

client/test/e2e.test.js
JackWilb: Fixed in 7087cd9

client/test/e2e.test.js
JackWilb: Moving this into a function for cleaner code, but under what circumstances would there be tables that exist in the workspace? The testing code uses the test environment from the client and the API and wipes it all out before it loads. Do you mean in case we add more tests first?

client/test/e2e.test.js
JackWilb: Fixed in 024a5f8

client/test/e2e.test.js
JackWilb: fixed in 7087cd9

client/vue.config.js
JackWilb: Fixed in 1787357

client/vue.config.js
JackWilb: Fixed in 1787357

JackWilb: typo fixed in 8dd025e

client/vue.config.js
JackWilb: typo fixed in 8dd025e

client/test/e2e.test.js
JackWilb: Fixed in 8dd025e

client/test/e2e.test.js
JackWilb: Fixed in 8dd025e

client/test/e2e.test.js
JackWilb: Fixed in 7087cd9

client/test/e2e.test.js
JackWilb: Yeah, that's right. In this outdated case, the modal never closed but in the new code it does. Either way it clears out the invalid workspace name

client/test/e2e.test.js
JackWilb: Fixed in 5b3ea58

client/test/e2e.test.js
JackWilb: Fixed in 5b3ea58

client/test/e2e.test.js
waxlamp: 9f65f3d

client/test/e2e.test.js
waxlamp: "In case we add more tests first" would be a definite way this could happen, but my general attitude about this is more defensive in nature: when testing a system, you should make as few assumptions as possible. So asserting that the table list is empty is a nice way to do that, and it prepares the formal precondition for the test that follows.

I agree it's an unlikely scenario, and one that we could fix if it ever did come up. But since it's such a small fix now, it's one of those things we could just consider doing and forgetting about.

client/test/e2e.test.js
JackWilb: This WIP code so has been moved to a new branc

client/test/e2e.test.js
JackWilb: Gotcha. This is the section I'm having issues with

JackWilb: Fixed in 1e4c955

client/test/e2e.test.js
```diff
@@ -0,0 +1,131 @@
...
+
+  exists = workspaces.includes(name);
+
+  return exists;
+}
+
+async function element_exists(empty = false, element_type, p, name) {
+  let exists, list_is_empty, tables;
+
+  await p.waitForSelector('[data-title="Tables"] .ws-detail-empty-list ');
```
waxlamp:
```suggestion
  await p.waitForSelector(`[data-title="${element_type}"] .ws-detail-empty-list`);
```

This should use the `element_type` argument right?

client/test/e2e.test.js
```diff
@@ -0,0 +1,131 @@
...
+  tables = await p.evaluate(element_type => {
+    let titles = [];
+    let doc_nodes = document.querySelectorAll('[data-title="' + element_type + '"] .ws-detail-empty-list ');
```
waxlamp:
```suggestion
    let doc_nodes = document.querySelectorAll(`[data-title="${element_type}"] .ws-detail-empty-list`);
```

client/test/e2e.test.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+    return titles;
+  });
+
+  exists = workspaces.includes(name);
+
+  return exists;
+}
+
+async function element_exists(empty = false, element_type, p, name) {
```
waxlamp: JS allows you to set an argument's default value without setting defaults for the ones after it, but it seems a bit strange. What's the intent here?

client/test/e2e.test.js
```diff
@@ -0,0 +1,131 @@
...
+    return titles;
+  }, element_type);
+
+  if (empty) {
+    list_is_empty = tables.includes('info There\'s nothing here yet...');
+    return list_is_empty;
+  } else {
+    exists = tables.includes(name);
+    return exists;
+  }
```
waxlamp: It's a bit of an antipattern to pack two different tests into one function, dispatching between them based on a boolean argument. A better approach might be to abstract everything before this if-block into a helper function, then split this if-block into two functions, each of which calls the helper function, then includes one branch or the other. That will make this code block easier to read, and the usage of these functions much clearer at their point of use.

client/test/e2e.test.js
JackWilb: Yes, good catch

client/test/e2e.test.js
JackWilb: My intent was for the default to be searching for a table that exists, not checking if it's empty. I guess arguments don't usually come in with:

```
element_exists(element_type = "Tables")
```

Though right?

client/test/e2e.test.js
JackWilb: I agree and I knew it as I wrote it, I was just having a mental block. I've split it out into 3 functions in b3f9d38

client/test/e2e.test.js
JackWilb: Splitting this function into 3 different ones gets rid of this problem. Fixed in b3f9d38

client/test/e2e.test.js
JackWilb: Fixed in b3f9d38.

client/test/e2e.test.js
JackWilb: Fixed in b3f9d38.

client/test/e2e.test.js
```diff
@@ -0,0 +1,150 @@
...
+  t.notOk(exists, 'Workspace called "123" wasn\'t created.');
+
+  exists = await workspace_exists(p, '++--==__');
+  t.notOk(exists, 'Workspace called "++--==__" wasn\'t created.');
+
+  exists = await workspace_exists(p, 'a');
+  t.ok(exists, 'Workspace called "a" was created.');
+
+  exists = await element_exists('Tables', p, 'broken');
+  t.notOk(exists, 'New workspaces don\t have tables');
```
waxlamp:
```suggestion
  t.notOk(exists, 'New workspaces don\'t have tables');
```

PR #199, opened 2019-11-05
Add an uploader for D3 graph format

### LINKED ISSUE #198: Add an uploader to handle the d3/js standard graph format
Since this format is popular around the web, it makes sense for us to be able to natively handle it. Here's an example of the format:

```
{
  nodes: [
    {"id": 1}
    {"id": 2}
    {"id": 3}
  ],
  links: [
    {"source": 1, "target": 2}
    {"source": 2, "target": 3}
    {"source": 3, "target": 1}
  ]

```

JackWilb:
Closes #198. 

Add a new uploader for the common js filetype and a test to validate it's function


JackWilb:
@waxlamp This is functional now. I'm having a couple issues that I think you might be able to help me solve. 

The first is that the linter is throwing some cryptic messages about errors that it finds in the first line of the files. Based on [this](https://pypi.org/project/flake8-mypy/) page, it looks like it's an internal flake8 error? Maybe you've dealt with this before. 

The second issue is that I used pandas for part of the maniuplation and loading the json into a list of OrderedDicts, to mirror the csv code. It looks like there are no typing stubs for pandas from what I can tell. We could get around that by adding `# type: ignore` after the pandas import by the looks of things but that doesn't seem ideal.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,99 @@
...
+    # Extract each table to pandas dataframes and change column names
+    nodes = pd.DataFrame(data["nodes"])
+    nodes["_key"] = nodes["id"]
+    del nodes["id"]
+
+    links = pd.DataFrame(data["links"])
+    links["_from"] = links["source"].apply(lambda x: table + "_nodes/" + x)
+    links["_to"] = links["target"].apply(lambda x: table + "_nodes/" + x)
+    del links["source"]
+    del links["target"]
```
waxlamp: I'd *really* prefer not to pull in Pandas, which is HUGE, just to do some table manipulation. If you like, I can take a crack at doing this part using standard library operations instead. I suspect it won't be much trouble. Let me know how you'd like to proceed.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,99 @@
...
+    if len(ids) != len(set(ids)):
+        data_errors.append({"error": "node_duplicates"})
+
+    # Duplicated edges are okay, they might encode different relationships, but
+    # that should be specified in the attributes. Thus duplicated dicts are not okay
+    if len(data["links"]) != len(set([tuple(x.items()) for x in data["links"]])):
+        data_errors.append({"error": "link_duplicates"})
+
+    if len(data_errors) > 0:
+        raise ValidationFailed(data_errors)
```
waxlamp: I think instead of raising the exception here, it's better to return the list of errors. Then, at the point where validation is needed, the caller can look at the result and decide what to do with it (in this case, simply raise `ValidationFailed()` as you do here).

This has two major benefits:
1. It makes the usage of this function more explicit where it's used, rather than making the reader of the code guess or have to look up that it raises an exception.
2. It makes testing this function simpler, as you don't have to catch exceptions to validate this function.

If you're copying the validation pattern from the other uploader modules, then I think you should still make this change here as suggested, and file an issue to fix the pattern in the other validators.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Check for duplicated nodes
+    ids = [row["id"] for row in data["nodes"]]
+    if len(ids) != len(set(ids)):
+        data_errors.append({"error": "node_duplicates"})
+
+    # Duplicated edges are okay, they might encode different relationships, but
+    # that should be specified in the attributes. Thus duplicated dicts are not okay
+    if len(data["links"]) != len(set([tuple(x.items()) for x in data["links"]])):
+        data_errors.append({"error": "link_duplicates"})
```
waxlamp: I'm not sure I understand this check--I think it's checking for identical attributes among different links, right? But why wouldn't that be allowed?

JackWilb: This is checking for duplicate links by converting the dict into a tuple and comparing them all. Duplicate links shouldn't be allowed, they should be collapsed down into one object

waxlamp: Right, but why can't we have multiple links with the same data on them?

In the simplest case, there wouldn't be any data on the link at all, and the multiplicity itself is the desired state.

JackWilb: I was thinking that in the case that you have the following link:

```
{source: a, target: b, likes: 5}
```

There should not be duplicates allowed. I do see your point about potentially allowing the duplicates, too. Should I remove this code and we can leave it to the user to properly format their data?

multinet/uploaders/d3_json.py
JackWilb: I think I've fixed this. It turns out you can specify in json.load that you want OrderedDicts, so I did and renamed the keys by looping through in 62dac8d. Let me know if this is what you meant.

multinet/uploaders/d3_json.py
JackWilb: I agree with all of your points here. I've updated the code for this uploader and have opened a new issue to update the other uploaders #200.

multinet/uploaders/d3_json.py
waxlamp: I'm almost certain that if you load a d3 json file with a repeated link record like that, it will parse just fine in, e.g., d3's own graph vis libraries. There may certainly be applications where you would want to restrict this sort of thing, but there are definitely applications where this would be expected. For that reason I don't believe we should disallow it at this most general level.

JackWilb: Removed this code in af0ee12

waxlamp:
I'm planning to take a deeper look at this today. If it's ready for review, please mark it as such. (If not, that's fine, but I'll still plan to start a review of it today.)

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+    ):
+        data_errors.append({"error": "invalid_link_keys"})
+
+    # Check that the keys for each dictionary match
+    if not all(data["links"][0].keys() == row.keys() for row in data["links"]):
+        data_errors.append({"error": "inconsistent_link_keys"})
+
+    # Check for duplicated nodes
+    ids = [row["id"] for row in data["nodes"]]
+    if len(ids) != len(set(ids)):
```
waxlamp: Instead of `len(ids)`, you should be able to use `len(data["nodes"])`. Then, line 39 can read `ids = set(row["id"] for row in data["nodes"])` (not the lack of square brackets around the list comprehension expression).

I think this will yield the same results, but use less memory at runtime.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+        data_errors.append({"error": "node_duplicates"})
+
+    return data_errors
+
+
+@bp.route("/<workspace>/<table>", methods=["POST"])
+def upload(workspace: str, table: str) -> Any:
+    """Store a d3-esque json file into the database as a node and edge table.
```
waxlamp:
```suggestion
    """Store a d3 json-encoded graph into the database as a node and edge table.
```

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+
+    `workspace` - the target workspace
+    `table` - the target table
+    `data` - the json data, passed in the request body. The json data should contain
+    nodes: [] and links: []
+    """
+    app.logger.info("Bulk Loading D3 Json Data")
```
waxlamp:
```suggestion
```

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+    # Get data from the request and load it as json
+    body = decode_data(request.data)
+    data = json.load(StringIO(body), object_pairs_hook=OrderedDict)
+
+    # Check file structure
+    errors = validate_d3_json(data)
+    if len(errors) > 0:
+        raise ValidationFailed(errors)
+
+    # Extract each table to pandas dataframes and change column names
```
waxlamp: Update this comment to reflect that pandas isn't being used.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+    nodes = data["nodes"]
+    for i in range(0, len(nodes)):
+        nodes[i]["_key"] = nodes[i]["id"]
+        del nodes[i]["id"]
```
waxlamp: I don't think you need the index parameter here; you should be able to change this to `for node in nodes:` and manipulate `node` directly in the body.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+
+    links = data["links"]
+    for i in range(0, len(links)):
+        links[i]["_from"] = table + "_nodes/" + links[i]["source"]
+        links[i]["_to"] = table + "_nodes/" + links[i]["target"]
+        del links[i]["source"]
+        del links[i]["target"]
```
waxlamp: Same thing here (use `for link in links:`).

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+    # Extract each table to pandas dataframes and change column names
+    nodes = data["nodes"]
+    for i in range(0, len(nodes)):
+        nodes[i]["_key"] = nodes[i]["id"]
+        del nodes[i]["id"]
+
+    links = data["links"]
+    for i in range(0, len(links)):
+        links[i]["_from"] = table + "_nodes/" + links[i]["source"]
+        links[i]["_to"] = table + "_nodes/" + links[i]["target"]
```
waxlamp: Use string interpolation for these two lines.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+        del links[i]["source"]
+        del links[i]["target"]
+
+    print("nodes" + str(nodes))
+    print("links" + str(links))
```
waxlamp: Remove these lines.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # Create the workspace
+    space = db.db(workspace)
+    if space.has_collection(table + "_nodes") or space.has_collection(table + "_links"):
+        nodes_coll = space.collection(table + "_nodes")
+        links_coll = space.collection(table + "_links")
```
waxlamp: The logic here doesn't look right--both collections must exist in order to pick them up like this.

Also, please use string interpolation to create the names of the collections, and factor out the sum expression from here and from lines 88 and 89 below.

multinet/uploaders/d3_json.py
```diff
@@ -0,0 +1,95 @@
...
+    else:
+        nodes_coll = space.create_collection(table + "_nodes", edge=False)
+        links_coll = space.create_collection(table + "_links", edge=True)
+
+    # Insert data
+    nodes_coll.insert_many(nodes)
+    links_coll.insert_many(links)
+
+    return dict(count=len(nodes) + len(links))
```
waxlamp: Could you follow the example of the [nested_json uploader](https://github.com/multinet-app/multinet/blob/d3-json-uploader/multinet/uploaders/nested_json.py#L116-L118) in crafting this return value?

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,39 @@
...
+    outcome1 = validate_d3_json(good_data)
+    outcome2 = validate_d3_json(dup_node_data)
+    outcome3 = validate_d3_json(incon_keys)
+    outcome4 = validate_d3_json(inval_keys)
+
+    # Assert: Check the outcomes against what we expect
+    assert len(outcome1) == 0
+    assert {"error": "node_duplicates"} == outcome2[0]
+    assert {"error": "inconsistent_link_keys"} == outcome3[0]
+    assert {"error": "invalid_link_keys"} == outcome4[0]
```
waxlamp: Can you add assertions about the numbers of failures expected in each case?

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,39 @@
...
+from collections import OrderedDict
+
+from multinet.uploaders.d3_json import validate_d3_json
+
+TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "data"))
+
+
+def test_validate_d3_json():
+    """Tests the validate_csv function."""
+    # Arrange: Set file paths
```
waxlamp: What does this `Arrange:` prefix do in a comment?

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,39 @@
...
+    good_file = os.path.join(TEST_DATA_DIR, "miserables.json")
+    duplicated_nodes = os.path.join(TEST_DATA_DIR, "miserables_duplicate_nodes.json")
+    incon_keys = os.path.join(TEST_DATA_DIR, "miserables_inconsistent_link_keys.json")
+    inval_keys = os.path.join(TEST_DATA_DIR, "miserables_invalid_link_keys.json")
```
waxlamp: These path names should just be transplanted directly into the `open()` calls below. If you want to do a bit of refactoring, it might be worthwhile to write a function that prefixes `TEST_DATA_DIR` for you, but separating the filenames from the file objects into two blocks makes this test a lot slower to read/understand.

multinet/uploaders/d3_json.py
JackWilb: Do you mean string formatting here?

multinet/uploaders/d3_json.py
JackWilb: I was following a similar structure to the csv here. I understand it would be best to spilt this up but I think the labels from the nested json uploader wouldn't make much sense. Should I split it into nodecount and edgecount?

test/test_d3_json_uploader.py
JackWilb: Nothing really. It helps me structure my tests and is part of the [AAA](https://www.c-sharpcorner.com/UploadFile/dacca2/fundamental-of-unit-testing-understand-aaa-in-unit-testing/) testing methodology.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 097c5a6.

multinet/uploaders/d3_json.py
JackWilb: Looks like that's the case, I've just never heard it referred to as interpolations. Working on a fix

JackWilb: Fixed in 3b0e3ba.

multinet/uploaders/d3_json.py
JackWilb: Fixed in 3b0e3ba.

multinet/uploaders/d3_json.py
JackWilb: Fixed using my compromise in 3b0e3ba.

multinet/uploaders/d3_json.py
waxlamp: I guess different languages have different names for it. Python calls it "formatted strings", while JavaScript uses "string interpolation". Wikipedia has an article using the latter as a title :man_shrugging:

test/test_d3_json_uploader.py
waxlamp: Cool, hadn't seen that approach to testing before.

test/test_d3_json_uploader.py
JackWilb: Fixed in cff37b6.

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,42 @@
...
+    assert outcome2[0] == {"error": "node_duplicates"}
+    assert len(outcome2) == 1
+
+    assert outcome3[0] == {"error": "inconsistent_link_keys"}
+    assert len(outcome3) == 1
+
+    print(outcome4)
+    assert outcome4[0] == {"error": "invalid_link_keys"}
+    assert outcome4[1] == {"error": "inconsistent_link_keys"}
+    assert len(outcome4) == 2
```
waxlamp: The length assertions should come before the content assertions, to avoid the test itself crashing on a content assertion if there aren't enough validation errors in the array.

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,42 @@
...
+    # Assert: Check the outcomes against what we expect
+    assert len(outcome1) == 0
+
+    assert outcome2[0] == {"error": "node_duplicates"}
+    assert len(outcome2) == 1
+
+    assert outcome3[0] == {"error": "inconsistent_link_keys"}
+    assert len(outcome3) == 1
+
+    print(outcome4)
```
waxlamp: Remove this print statement please.

multinet/util.py
```diff
@@ -40,3 +43,10 @@ def decode_data(input: bytes) -> str:
         raise DecodeFailed(str(e))
 
     return body
+
+
+def test_data_path(file_name: str) -> str:
+    """Load data from the test directory."""
+    file_path = os.path.join(TEST_DATA_DIR, file_name)
+    print(file_path)
+    return file_path
```
waxlamp: I think this function should go directly in the file where it's used.

test/test_d3_json_uploader.py
```diff
@@ -0,0 +1,42 @@
+"""Tests functions in the CSV Uploader Flask Blueprint."""
+import json
+import os
+from collections import OrderedDict
+
+from multinet.util import test_data_path
```
waxlamp: See other comment about bringing this function back to this file.

test/test_d3_json_uploader.py
JackWilb: Fixed in 83a5c03.

test/test_d3_json_uploader.py
JackWilb: Fixed in 83a5c03.

multinet/util.py
JackWilb: My plan is to have this be usable by all the tests to reduce code clutter. Does that sound good or should I move it and duplicate it to the other tests in another PR?

waxlamp: No you're right, it should stay here. Just change the name of the function so pytest doesn't get confused by it. Maybe just `data_path()`?

JackWilb: Fixed in e7bca47

PR #201, opened 2019-11-14
Add autofocus attribute to create workspace dialog

danlamanna:


danlamanna:
I haven't actually tested this since I don't have a dev space up, but I hope the intent is clear. Clicking create workspace should let the user immediately start typing, instead of having to click the text dialog for focus.

waxlamp:
This works like a charm. If you mark it ready for review I'll approve it. Thanks @danlamanna!

JackWilb:
Only seems to work on chrome for me, not firefox. I also think we should expand this to the create table +  create graph dialogs too for completeness

waxlamp:
The problem with Firefox seems to be something Vuetify is actively trying to fix. I agree about adding `autofocus` to the other dialogs, but let's do them in followup PRs since Dan's busy with other stuff today.

PR #205, opened 2019-12-05
Add Swagger documentation for API

### LINKED ISSUE #143: Standardize and document REST API
We need to overhaul the developing REST API to make sure that
1. the endpoints are carefully designed to be clear, logical, and useful
2. the return values from each endpoint are carefully designed and documented

One impetus for this is the fact that Flask does not support returning Python list objects from view functions, though it does allow dict objects. However, list objects *can* serve as the return type if the view function takes care to convert them to JSON first. So one major question is: should be standardize on dicts or freely use dicts or lists (or other types) where each is appropriate?

Also, see discussion in #140.

### LINKED ISSUE #154: Add Swagger documentation
This will involve seeing if there's a Flask plugin or something like that for generating Swagger docs

waxlamp:
This PR adds documentation for every API endpoint, and enables Flasgger to display them via Swagger at `/apidocs` on the API server.

Closes #143 
Closes #154


multinet/swagger/create_graph.yaml
```diff
@@ -0,0 +1,33 @@
...
+  200:
+    description: Graph successfully created
+    schema:
+      $ref: "#/definitions/entity_name"
+
+  400:
+    description: Graph could not be created
+    schema:
+      $ref: "#/definitions/entity_name"
+
```
JackWilb: Could 409 if graph already exists

multinet/swagger/workspace_tables.yaml
```diff
@@ -0,0 +1,21 @@
...
+    schema:
+      type: array
+      items:
+        $ref: "#/definitions/entity_name"
+
+  404:
+    description: Specified workspace could not be found
+    schema:
+      $ref: "#/definitions/entity_name"
+
```
JackWilb: Can return a 400 from BadQueryArgument if type is wrong

multinet/uploaders/swagger/csv.yaml
```diff
@@ -0,0 +1,32 @@
...
+  200:
+    description: CSV uploaded to table
+    schema:
+      type: object
+      properties:
+        count:
+          type: integer
+      example:
+        count: 3
+
```
JackWilb: Can return 400 if the validation fails

multinet/uploaders/swagger/d3_json.yaml
```diff
@@ -0,0 +1,27 @@
...
+  200:
+    description: D3 data uploaded to tables
+    schema:
+      type: object
+      properties:
+        edgecount:
+          type: integer
+        nodecount:
+          type: integer
+
```
JackWilb: Can return 400 if the validation fails

multinet/uploaders/swagger/nested_json.yaml
```diff
@@ -0,0 +1,29 @@
...
+    schema:
+      type: object
+      properties:
+        edgecount:
+          type: integer
+        int_nodecount:
+          type: integer
+        leaf_nodecount:
+          type: integer
+
```
JackWilb: This doesn't return 400 but maybe it should. It looks like the nested json doesn't do any validation. Should we open an issue here?

multinet/uploaders/swagger/newick.yaml
```diff
@@ -0,0 +1,28 @@
...
+  200:
+    description: Newick data uploaded to tables
+    schema:
+      type: object
+      properties:
+        edgecount:
+          type: integer
+        nodecount:
+          type: integer
+
```
JackWilb: Can return 400 if the validation fails

multinet/uploaders/swagger/newick.yaml
```diff
@@ -0,0 +1,28 @@
...
+      example: ((raccoon:19.19959,bear:6.80041):0.84600,((sea_lion:11.99700, seal:12.00300):7.52973,((monkey:100.85930,cat:47.14069):20.59201, weasel:18.87953):2.09460):3.87382,dog:25.46154);
+
+responses:
+  200:
+    description: Newick data uploaded to tables
+    schema:
+      type: object
+      properties:
+        edgecount:
+          type: integer
```
JackWilb: In all the uploaders using this definition shows what would be 0 uploaded nodes and edges. Can we change these to match up with the example data and what you'd expect uploading that to a fresh workspace?

multinet/uploaders/swagger/nested_json.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - text/plain
+
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+  - name: data
+    in: body
+    description: Raw JSON content of file
+    schema:
+      type: object
```
JackWilb: Can we add a richer example here?

multinet/uploaders/swagger/d3_json.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - text/plain
+
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+  - name: data
+    in: body
+    description: Raw JSON text
+    schema:
+      type: string
```
JackWilb: Can we add a richer example here?

multinet/swagger/workspace_tables.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table_type"
+
+responses:
+  200:
+    description: A list of table names belonging to the specified workspace
+    schema:
+      type: array
+      items:
+        $ref: "#/definitions/entity_name"
```
JackWilb: You use "fizzbin" for both the workspace and the table which is technically allowed but might be confusing in the docs. Could we change one?

multinet/swagger/node_edges.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - $ref: "#/parameters/node"
+  - $ref: "#/parameters/direction"
+  - $ref: "#/parameters/offset"
+  - $ref: "#/parameters/limit"
+
+responses:
+  200:
+    description: A list of edges
+    schema:
+      $ref: "#/definitions/entity_names"
```
JackWilb: Can we do a richer example here?

multinet/swagger/graph_nodes.yaml
```diff
@@ -0,0 +1,21 @@
...
+  - $ref: "#/parameters/graph"
+  - $ref: "#/parameters/offset"
+  - $ref: "#/parameters/limit"
+
+responses:
+  200:
+    description: A list of nodes from the requested graph
+    schema:
+      $ref: "#/definitions/entity_names"
+
```
JackWilb: Can we do a richer example here?

multinet/uploaders/swagger/nested_json.yaml
waxlamp: Yes, let's open an issue. It's probably a low-priority thing unless we run into problems with processing nested_json files, which we probably won't since the process for getting data into evo-bio has changed (and no longer uses nested_json files).

multinet/uploaders/swagger/newick.yaml
waxlamp: The major reason I didn't add examples for these is that this response is almost certainly not what we should be sending back. A more useful response would be, for instance, the list of tables created, and what workspace they belong to. I think we should open an issue to discuss that, rather than spend the time on enriching these examples.

On the other hand, it's pretty simple to create examples for these too. Let me see about doing that as I respond to your review as a whole.

multinet/uploaders/swagger/nested_json.yaml
waxlamp: See https://github.com/multinet-app/multinet/pull/205#discussion_r354479906.

multinet/uploaders/swagger/d3_json.yaml
waxlamp: See https://github.com/multinet-app/multinet/pull/205#discussion_r354479906.

multinet/swagger/workspace_tables.yaml
waxlamp: This is just supposed to be an example that shows names coming back, so I'm not super concerned with whether these would be confusing. It will complicate the specs to have different examples for each thing (workspace, table, graph, node id, edge id) so I'd rather just leave it alone.

multinet/swagger/node_edges.yaml
waxlamp: I think this is connected to the discussion in https://github.com/multinet-app/multinet/pull/205#discussion_r354480703.

multinet/swagger/graph_nodes.yaml
waxlamp: Oops, I think this is a mistake actually. I removed `entity_names` because it's not markedly different from `entity_name` embedded in a list type to justify it as a separate model. Will fix (and will look at enriching the example value if possible).

multinet/swagger/create_workspace.yaml
```diff
@@ -0,0 +1,18 @@
...
+parameters:
+  - $ref: "#/parameters/workspace"
+
+responses:
+  200:
+    description: Workspace successfully created
+    schema:
+      $ref: "#/definitions/entity_name"
+
+  400:
```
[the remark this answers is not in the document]
waxlamp: ddd8b5360473fc1ae0034cda34099d55a7702180

multinet/swagger/create_graph.yaml
waxlamp: ddd8b5360473fc1ae0034cda34099d55a7702180

multinet/swagger/aql.yaml
```diff
@@ -0,0 +1,33 @@
...
+      items:
+        $ref: "#/definitions/any_type"
+      example:
+        - _id: table1/0
+          name: Picard
+        - _id: table1/1
+          name: Riker
+        - _id: table1/2
+          name: Troi
+
```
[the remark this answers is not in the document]
waxlamp: c9eab1b60380729522734c6000dccbdc1363957d

multinet/swagger/workspace_tables.yaml
waxlamp: e68172930be12b3026b1e72593c333e1f5e839c4

Also made this change to the `node_edges` endpoint.

multinet/uploaders/swagger/csv.yaml
waxlamp: 8d600b7fd656afab80ac9233b661e207677d07e4

multinet/uploaders/swagger/d3_json.yaml
waxlamp: 8d600b7fd656afab80ac9233b661e207677d07e4

multinet/uploaders/swagger/newick.yaml
waxlamp: b072389ce9c73639630888fffdec8eb4d4bd0f65

multinet/swagger/graph_nodes.yaml
waxlamp: 66e11acd1cff83b31a8fe5d1dd4315f7c525a0f7

multinet/uploaders/swagger/nested_json.yaml
waxlamp: Looks like #159 already exists :)

multinet/swagger/node_edges.yaml
waxlamp: 66e11acd1cff83b31a8fe5d1dd4315f7c525a0f7

multinet/uploaders/swagger/d3_json.yaml
waxlamp: 50daf7efb8268f5f7a3970dcf620e9fec59df838

multinet/uploaders/swagger/nested_json.yaml
waxlamp: 50daf7efb8268f5f7a3970dcf620e9fec59df838

multinet/uploaders/swagger/newick.yaml
waxlamp: 50daf7efb8268f5f7a3970dcf620e9fec59df838

multinet/swagger/workspace_tables.yaml
waxlamp: 3ee5250f4708fe9427e2a9a5c1a5d93fe0446cf5

multinet/uploaders/swagger/d3_json.yaml
JackWilb: Sorry, I should have been more explicit here. I meant can we add an example of the d3 data format in the data (body) parameter? The same for the nested json

waxlamp: Ah ok. Yeah we can do that but the nested_json example is difficult to get together (because the examples I have are rather large, production-oriented files). Let me take a crack at this and add it in and we'll see what it looks like.

multinet/uploaders/swagger/nested_json.yaml
waxlamp: 8be5a8589f1f75f486416e91b8f484ff3c0be2f0

multinet/uploaders/swagger/d3_json.yaml
waxlamp: 8be5a8589f1f75f486416e91b8f484ff3c0be2f0

PR #213, opened 2019-12-11
Add UI for deleting workspaces

### LINKED ISSUE #78: Add ability to delete workspaces
Mentioned as a TODO in #76

waxlamp:
Depends on multinet-app/multinet-server#212 and multinet-app/multinet-server#216 
Closes multinet-app/multinet-server#78 

This PR adds functionality to the "delete workspaces" button.

It adds a confirmation dialog with a built-in delay of 2 seconds before the user can click the "confirm" button.

One issue is that the `GET /workspaces` endpoint doesn't seem to immediately update after the `DELETE /workspaces/<id>` call succeeds. To get around this, I have a one-second delay before the Sidebar component refreshes the workspaces list. One second works but is not guaranteed formally. A better solution might be to manually update the Sidebar component's workspace list (as we do when creating a new workspace), in which case the delay won't matter.

Another important task here is to review just the changes to the `multinet` directory, so that I can cut a release of that and make use of it in the client.

I'd like a bunch of reviews on this for various aspects:
- @JackWilb: please let me know if this is the right interaction for what we need in the Multinet project
- @jtomeck: please confirm that my use of Vuetify is good and correct (and that the style is reasonable)
- @subdavis: I can't formally request a review from you until you accept the invitation to the project, but I'd like to know if my Vue/TypeScript usage here is good

Thanks everyone!

TODO:
- [x] Release multinetjs 0.3.0 (which includes `deleteWorkspace()` function)
- [x] Update client to use multinetjs 0.3.0

### DIFF
--- a/client/src/components/DeleteWorkspaceDialog.vue
+++ b/client/src/components/DeleteWorkspaceDialog.vue
@@ -111,7 +111,7 @@ export default Vue.extend({
         selection,
       } = this;
 
-      selection.forEach(async ws => {
+      selection.forEach(async (ws) => {
         await api.deleteWorkspace(ws);
       });
 
--- a/client/src/components/DeleteWorkspaceDialog.vue
+++ b/client/src/components/DeleteWorkspaceDialog.vue
@@ -115,6 +115,7 @@ export default Vue.extend({
         await api.deleteWorkspace(ws);
       });
 
+      this.$emit('deleted');
       this.dialog = false;
     },
   },
--- a/client/src/components/DeleteWorkspaceDialog.vue
+++ b/client/src/components/DeleteWorkspaceDialog.vue
@@ -1,0 +1,122 @@
+<template>
+
+  <v-dialog
+    v-model="dialog"
+    width="700"
+    >
+
+    <template v-slot:activator="{ on }">
+      <v-tooltip right>
+        <template v-slot:activator="tooltip">
+          <v-scroll-x-transition>
+            <v-btn
+              icon
+              small
+              text
+              v-if="somethingChecked"
+              v-on="on"
+              >
+              <v-icon color="red accent-3" size="22px">delete_sweep</v-icon>
+            </v-btn>
+          </v-scroll-x-transition>
+        </template>
+        <span>Delete selected</span>
+      </v-tooltip>
+    </template>
+
+    <v-card>
+      <v-card-title
+        class="headline pb-0 pt-3"
+        primary-title
+        >
+        Delete Workspaces
+      </v-card-title>
+
+      <v-card-text class="px-4 pt-4 pb-1">
+        You are about to delete {{ selection.length }} workspace{{plural}}. <strong>Are you sure?</strong>
+      </v-card-text>
+
+      <v-divider />
+
+      <v-card-actions class="px-4 py-3">
+        <v-spacer />
+        <v-btn
+          depressed
+          color="error"
+          @click="execute"
+          :disabled="disabled"
+        >yes</v-btn>
+
+        <v-btn
+          depressed
+          @click="dialog = false"
+          >cancel</v-btn>
+      </v-card-actions>
+
+    </v-card>
+
+  </v-dialog>
+
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+import api from '@/api';
+
+export default Vue.extend({
+  props: {
+    somethingChecked: {
+      type: Boolean,
+      required: true,
+    },
+
+    selection: {
+      required: true,
+    },
+  },
+
+  data() {
+    return {
+      dialog: false,
+      disabled: true,
+      timeout: null as number | null,
+    };
+  },
+
+  computed: {
+    plural() {
+      return this.selection.length > 1 ? 's' : '';
+    },
+  },
+
+  watch: {
+    dialog() {
+      if (this.dialog) {
+        this.timeout = window.setTimeout(() => {
+          this.disabled = false;
+          this.timeout = null;
+        }, 2000);
+      } else {
+        window.clearTimeout(this.timeout);
+        this.disabled = true;
+        this.timeout = null;
+      }
+    },
+  },
+
+  methods: {
+    async execute() {
+      const {
+        selection,
+      } = this;
+
+      selection.forEach(async ws => {
+        await api.deleteWorkspace(ws);
+      });
+
+      this.dialog = false;
+    },
+  },
+});
+</script>
--- a/client/src/components/DeleteWorkspaceDialog.vue
+++ b/client/src/components/DeleteWorkspaceDialog.vue
@@ -60,18 +60,19 @@
 </template>
 
 <script lang="ts">
-import Vue from 'vue';
+import Vue, { PropType } from 'vue';
 
 import api from '@/api';
 
 export default Vue.extend({
   props: {
     somethingChecked: {
-      type: Boolean,
+      type: Object as PropType<boolean>,
       required: true,
     },
 
     selection: {
+      type: Object as PropType<string[]>,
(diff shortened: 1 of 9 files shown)


JackWilb:
Just about to pull it down and test it but I also noticed that, f7c69d5, comes from the other PR for the `no-checkbox-animation` branch. We should merge this branch after that one probably, right?

waxlamp:
> Just about to pull it down and test it but I also noticed that, [f7c69d5](https://github.com/multinet-app/multinet/commit/f7c69d5505bcebd597b091c29f304ac5a74ce3be), comes from the other PR for the `no-checkbox-animation` branch. We should merge this branch after that one probably, right?

Yes, I forgot to mention the dependency in the PR description; will add that now.

subdavis (dismissed):
Seems fine.  I haven't done a lot of typescript yet so I don't feel like I have a great eye for what is and isn't good practice.

Is void the assumed return type of methods where return value is undocumented?  If you return something, is that a compiler warning or does it work?

waxlamp:
> Seems fine. I haven't done a lot of typescript yet so I don't feel like I have a great eye for what is and isn't good practice.
> 

What do you think about the artificial delay before retrieving the workspaces again? Would I be better off just updating the local data from the known changes to it, rather than waiting an arbitrary time and asking the server directly?

> Is void the assumed return type of methods where return value is undocumented? If you return something, is that a compiler warning or does it work?

I think so... I think the idea is that if you add a return statement, the compiler will infer the return type if it can, and then warn if something else uses that return value in a type-incompatible way. But of course it's better to just be explicit.

client/src/components/DeleteWorkspaceDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+  },
+
+  methods: {
+    async execute() {
+      const {
+        selection,
+      } = this;
+
+      selection.forEach(async (ws) => {
+        await api.deleteWorkspace(ws);
```
JackWilb: Shouldn't this be taking a path as the argument, so `workspaces/${ws}`

JackWilb (changes requested):
Left a note and I'm also getting a bunch of errors when deleting, and the workspace remains.

Here are some of the errors:
```
TypeError: _api__WEBPACK_IMPORTED_MODULE_4__.default.deleteWorkspace is not a function

uncaught exception: Object

[Vue warn]: Invalid prop: type check failed for prop "selection". Expected Object, got Array 

found in

---> <DeleteWorkspaceDialog> at src/components/DeleteWorkspaceDialog.vue
       <VSubheader>
         <VList>
           <VNavigationDrawer>
             <Sidebar> at src/components/Sidebar.vue
               <VApp>
                 <App> at src/App.vue
                   <Root>

[Vue warn]: Invalid prop: type check failed for prop "somethingChecked". Expected Object, got Boolean with value false.

found in

---> <DeleteWorkspaceDialog> at src/components/DeleteWorkspaceDialog.vue
       <VSubheader>
         <VList>
           <VNavigationDrawer>
             <Sidebar> at src/components/Sidebar.vue
               <VApp>
                 <App> at src/App.vue
                   <Root>
```

client/src/components/DeleteWorkspaceDialog.vue
waxlamp: Since this is the client library, there's no need to force the caller to prepend the "workspaces/" part; if you look at the library source code, you'll see that the function takes care of preparing the correct API path to use to make the DELETE call.

client/src/components/DeleteWorkspaceDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+            </v-btn>
+          </v-scroll-x-transition>
+        </template>
+        <span>Delete selected</span>
+      </v-tooltip>
+    </template>
+
+    <v-card>
+      <v-card-title
+        class="headline pb-0 pt-3"
```
jtomeck: `class="headline pb-0 pt-3 px-5"`

client/src/components/DeleteWorkspaceDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+        primary-title
+        >
+        Delete Workspaces
+      </v-card-title>
+
+      <v-card-text class="px-4 pt-4 pb-1">
```
jtomeck: `class="px-5 py-4"`

client/src/components/DeleteWorkspaceDialog.vue
JackWilb: Ah, makes sense!

client/src/components/DeleteWorkspaceDialog.vue
```diff
@@ -0,0 +1,125 @@
...
+
+      <v-divider />
+
+      <v-card-actions class="px-4 py-3">
+        <v-spacer />
+        <v-btn
+          depressed
+          color="error"
+          @click="execute"
+          :disabled="disabled"
```
jtomeck: For some reason the button looks disabled initially, but after a couple seconds turns red. Is this delay intentional?

jtomeck (changes requested):

client/src/components/DeleteWorkspaceDialog.vue
JackWilb: It was:
> One issue is that the GET /workspaces endpoint doesn't seem to immediately update after the DELETE /workspaces/<id> call succeeds. To get around this, I have a one-second delay before the Sidebar component refreshes the workspaces list. One second works but is not guaranteed formally.

waxlamp: Yeah, it's my way of slowing the user down before they do something irreversible.

It's quite likely bad UX. The better solution is to not actually delete, but simply "archive" the workspaces so they no longer show up, but are not actually deleted. See #214.

client/src/components/DeleteWorkspaceDialog.vue
waxlamp: 5d744e22f20ef62335af0971663f0948c15080a3

client/src/components/DeleteWorkspaceDialog.vue
waxlamp: 5d744e22f20ef62335af0971663f0948c15080a3

jtomeck (approved):

JackWilb (approved):

waxlamp:
> Left a note and I'm also getting a bunch of errors when deleting, and the workspace remains.
> 
> Here are some of the errors:
> 
> ```
> TypeError: _api__WEBPACK_IMPORTED_MODULE_4__.default.deleteWorkspace is not a function
> 
> uncaught exception: Object
> 
> [Vue warn]: Invalid prop: type check failed for prop "selection". Expected Object, got Array 
> 
> found in
> 
> ---> <DeleteWorkspaceDialog> at src/components/DeleteWorkspaceDialog.vue
>        <VSubheader>
>          <VList>
>            <VNavigationDrawer>
>              <Sidebar> at src/components/Sidebar.vue
>                <VApp>
>                  <App> at src/App.vue
>                    <Root>
> 
> [Vue warn]: Invalid prop: type check failed for prop "somethingChecked". Expected Object, got Boolean with value false.
> 
> found in
> 
> ---> <DeleteWorkspaceDialog> at src/components/DeleteWorkspaceDialog.vue
>        <VSubheader>
>          <VList>
>            <VNavigationDrawer>
>              <Sidebar> at src/components/Sidebar.vue
>                <VApp>
>                  <App> at src/App.vue
>                    <Root>
> ```

Just for posterity: the webpack error was appearing because the local build was still trying to use an old version of the multinet library, and the prop types error was fixed in b341d35.

PR #220, opened 2019-12-17
Node detail view

jtomeck:


jtomeck:
> Can we try reversing the sizes for the column name and the value?

I normally wouldn't push back on something this small, but I feel very strongly that values should be emphasized over labels. Instead of swapping the sizes, do you think it would help to just show the label on top of the value, keeping the same size?

> what happens if there are more and more attributes?

I wasn't sure there'd be more than 3, so I did not take that into account. I'll look into changing this!

waxlamp:
> > Can we try reversing the sizes for the column name and the value?
> 
> I normally wouldn't push back on something this small, but I feel very strongly that values should be emphasized over labels. Instead of swapping the sizes, do you think it would help to just show the label on top of the value, keeping the same size?
> 

Reversing is good. Maybe we can also play with making the values a bit smaller to see how that looks--they just feel huge to me right now.

> > what happens if there are more and more attributes?
> 
> I wasn't sure there'd be more than 3, so I did not take that into account. I'll look into changing this!

The general use case is that each node will have some arbitrary amount of data on it, so we should think some about how best to display that.

It also occurs to me that the key and id values are special--every node in the system will have them, so it may make more sense to simply elevate the id value itself to a sort of header or something like that.

I will discuss this with you in person, as there's a lot of potential for misunderstanding over the internets.

jtomeck:
> I added a commit updating the header title slightly; @jtomeck, please let me know if you think it's an improvement.

It seems redundant to have it there, when it's already in the main header. Do you think changing it to "node/key" format instead of "node   >   key" like it is now would help?

> Also, the blue of the attribute panel seems really harsh--is there a lighter or softer version that we could try out?

Sure thing!

waxlamp:
> > I added a commit updating the header title slightly; @jtomeck, please let me know if you think it's an improvement.
> 
> It seems redundant to have it there, when it's already in the main header. Do you think changing it to "node/key" format instead of "node > key" like it is now would help?
> 

I'm fine with redundancy as long as it's helpful. Here I think it is: the breadcrumb in the main header is useful for navigation purposes, and if you think of the node detail view itself as a standalone thing, then I think it's important to have the node id displayed in its own authoritative header as well. Note that the id is shown once more as part of the attribute display, but even there the redundancy would be important in order to highlight that that value is special for the node.

That being said, I do think it would be an improvement in the breadcrumb to display the id as "node/key" in place of "node > key".

> > Also, the blue of the attribute panel seems really harsh--is there a lighter or softer version that we could try out?
> 
> Sure thing!

Excellent, thanks!

PR #223, opened 2019-12-26
Fix tooltip positioning bug

waxlamp:
According to https://github.com/vuetifyjs/vuetify/issues/10037 we can fix the tooltip positioning issue by hoisting the v-if upward. Unfortunately, tooltips just don't work well with transitions, though they are working on fixing that in a later release. Consequently, I'm leaving the transition components in the code, but they don't actually work.

I figured it was better to have correctly positioned tooltips and put the transition animations on hold, if that's the cost of doing so.

### DIFF
--- a/client/src/components/ItemPanel.vue
+++ b/client/src/components/ItemPanel.vue
@@ -9,14 +9,13 @@
 
       <v-spacer />
 
-      <v-tooltip top>
+      <v-tooltip top v-if="anySelected">
         <template v-slot:activator="{ on }">
           <v-scroll-x-transition>
             <v-btn
               icon
               small
               text
-              v-if="anySelected"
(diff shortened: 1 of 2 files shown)


client/src/components/ItemPanel.vue
```diff
@@ -9,14 +9,13 @@
 
       <v-spacer />
 
-      <v-tooltip top>
+      <v-tooltip top v-if="anySelected">
         <template v-slot:activator="{ on }">
           <v-scroll-x-transition>
```
KaelWD: Does it work if the transition is above the tooltip?

waxlamp: Unfortunately, no :crying_cat_face:

Any other ideas?

PR #224, opened 2019-12-26
Add shell.nix file

waxlamp:


waxlamp:
Yeah, this is just needed for people wanting to use Nix to develop (like I do on my laptop). It's not appropriate to gitignore it because it may need to be updated along with the project dependencies. However, this also doesn't affect anyone not using Nix, so it shouldn't cause any undue burden on the developers.

PR #225, opened 2019-12-27
Add UI for deleting graphs and tables

waxlamp:
This adds several pieces to support graph and table deletion in the client:
- API pieces for deleting graphs and deleting tables
- client library functions for accessing those new API routes
- dialogs in the client to implement graph deletion and table deletion
  - a safety interlock on table deletion: if an existing graph uses any of the tables slated for deletion, the UI will not allow the deletion to proceed until the affected graphs are deleted first

Because this PR updates the client library, it's necessary to use `yarn link` in order to test locally:
1. Go into the `multinetjs` directory, run `yarn install` and then `yarn build` there.
2. Run `yarn link` and verify that the system has registered multinet as a linkable module.
3. Go into the `client` directory and run `yarn link multinet` to use the linked version of the multinet library.
4. Run `yarn install` and `yarn serve` again if necessary.

These steps will let you use the modified version of multinetjs for testing.

That also means there are TODO items before merging this PR:
- [x] Cut a release of multinetjs
- [x] Update the client code to make use of the new version of multinetjs

I will take care of these TODOs after review is done, and just before I merge.

### DIFF
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -283,6 +283,16 @@ def table_fields(workspace: str, table: str, arango: ArangoClient) -> List[str]:
 
 
 @with_client
+def delete_table(workspace: str, table: str, arango: ArangoClient) -> str:
+    """Delete a table."""
+    space = db(workspace, arango=arango)
+    if space.has_collection(table):
+        space.delete_collection(table)
+
+    return table
+
+
+@with_client
 def aql_query(
     workspace: str, query: str, arango: ArangoClient
 ) -> Generator[dict, None, None]:
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -317,6 +317,16 @@ def create_graph(
 
 
 @with_client
+def delete_graph(workspace: str, graph: str, arango: ArangoClient) -> str:
+    """Delete graph `graph` from workspace `workspace`."""
+    space = db(workspace, arango=arango)
+    if space.has_graph(graph):
+        space.delete_graph(graph)
+
+    return graph
+
+
+@with_client
 def graph_node_tables(
     workspace: str, graph: str, arango: ArangoClient
 ) -> List[StandardCollection]:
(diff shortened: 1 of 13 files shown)


waxlamp:
> This looks good visually. My only recommendation is mentioned in #226. Your call whether that should be a different PR, but I'll approve for now.

I'll handle that in a separate PR--there are a few places where that delay mechanism is used, which means we might want to do some refactoring. Also, the PRs that are in flight now are going to cause conflicts, and can't all be reviewed until probably next week.

multinet/db.py
```diff
@@ -282,6 +282,16 @@ def table_fields(workspace: str, table: str, arango: ArangoClient) -> List[str]:
...
 
 
+@with_client
+def delete_table(workspace: str, table: str, arango: ArangoClient) -> str:
+    """Delete a table."""
+    space = db(workspace, arango=arango)
+    if space.has_collection(table):
+        space.delete_collection(table)
+
+    return table
```
JackWilb: We should check if the table is in a graph here since it's currently possible to break the dependence with the api

waxlamp: That's not a bad idea.

My thinking around allowing it in the API is that it's not technically an error to delete a table while a graph depends on it. Imagine, for instance, that you want to replace a bad table with a good one. If we disallow deleting tables that a graph depends on, then the only way to do that is to delete the graph, delete the table, re-upload the table, then re-form the graph. I'd argue that's bad UI.

But, we could definitely have the default be for this delete method to check for a conflict and send back an error code, but also have a `force` parameter that skips the check.

In our use case, however, we still need to check manually for a conflict because of the logic of the delete tables dialog box: that dialog shouldn't allow you to try to delete the tables before it tells you why you can't; it ought to tell you right away. So, I'm for adding a guard to this method, though it doesn't affect our current use case.

JackWilb: This sounds like a suitable approach

waxlamp: See #228

PR #232, opened 2020-01-02
Add openflights dataset

waxlamp:
This adds the "flights" dataset from [openflights](https://openflights.org/data.html).

The script `fetchdata.sh` uses `curl` to get the raw data from the OpenFlights website, and then the following commands transform it to produce CSV files suitable for multinet. The PR also adds the final CSV files as a convenience, with the scripts themselves standing in as a provenance for the data.


waxlamp:
> Looks good. I assume the csv files are the fully cleaned version using the process.py script?

Yep--I generated and tested the output and checked them into the repo as a convenience. The scripts are mainly for provenance.

I will update the PR description.

PR #235, opened 2020-01-02
Table detail improvements

jtomeck:


jtomeck:
@waxlamp - I have `white-space: nowrap;` on all the cells, which basically forces them to stay single line and not break if they get longer. Do you see this being a problem in the future?

waxlamp:
> @waxlamp - I have `white-space: nowrap;` on all the cells, which basically forces them to stay single line and not break if they get longer. Do you see this being a problem in the future?

As you know, users will always find a way to break everything :smirk_cat: 

But this makes the table display as it used to before, right? Seems like a good place to be right now.

I think if we come upon a dataset with really long data values, we can reconsider what to do. Perhaps something like `text-overflow: ellipses` together with a tooltip showing the full value.

waxlamp:
Is this ready for review?

PR #239, opened 2020-01-07
Env file change

### LINKED ISSUE #238: Change including .env file
I think it would be beneficial to instead of checking the `.env` file directly into the repository, including something like `.env.default`, (and ignoring `.env`) which contains the default values currently in `.env`. Then, the user can just copy `.env.default` to `.env`, and change any variables that need to be changed. This is a pattern I've used before in personal project and it seems to work well.

This has the benefit of working automatically with docker-compose, pipenv, etc. An example of something I currently use this for is the `ARANGO_DATA` environment variable. So instead of having to specify this everytime I run `docker-compose up`, it's just specified in the `.env` file. If this sounds good I can make these changes.

AlmightyYakob:
This fixes #238


AlmightyYakob:
@JackWilb Looking at d33c786, will this PR affect the deploy process?

Pipfile
```diff
@@ -20,7 +20,7 @@ flake8-black = "==0.1.0"
 flake8-bugbear = "==19.8.0"
 flake8-docstrings = "==1.3.1"
 flake8-quotes = "==2.1.0"
-sphinx = "==2.1.2"
+sphinx = "==2.3.1"
```
waxlamp: Why this change?

docs/installation.rst
```diff
@@ -22,9 +30,9 @@ Set Up Multinet
 
    and then launch the Docker container using::
 
-       $ ARANGO_DATA=~/.local/multinet/arango docker-compose up
+       $ ARANGO_DATA=~/.local/multinet/arango docker-compose up -d
 
-3. Inspect the ``.env`` file, which contains a few useful environment variable
+5. Inspect the ``.env`` file, which contains a few useful environment variable
```
waxlamp: Since the .env file is more of an important configuration device now, I think this step should be moved above the "run docker-compose" step.

Pipfile
AlmightyYakob: It seems that there was an update to a dependency of sphinx (probably unpinned) that broke the doc build process. It's fixed in this update

docs/installation.rst
AlmightyYakob: Will do

JackWilb:
> @JackWilb Looking at [d33c786](https://github.com/multinet-app/multinet/commit/d33c78699bab79cc3b0098115d142c6b280dc56f), will this PR affect the deploy process?

It shouldn't @AlmightyYakob. When we're deploying to production we're using uwsgi and building the client apps (not even touching .env from what I can tell). Therefore there shouldn't be an issue with the .env being renamed.

[Here](https://github.com/multinet-app/multinet-deploy/blob/81525950d19dc0b898313a8c8f7afb3c5214eeb2/ansible/playbook.yml#L53) is the start of the relevant code section. You can also look in the tasks folder and see we're never

JackWilb:
If Roni, agrees I think we can merge

Issue #185, opened 2019-10-17
404 errors + others don't work for the main app and for the sub-components


JackWilb:
If you navigate to multinet.app/123/123 it just redirects back to the homepage which may be the desired pathing (can you confirm this, please, @waxlamp?) Also, navigating to multinet.app/nodelink/123/123 returns a blank page. Is this something that needs to be fixed on a per app basis or can we handle this elsewhere?

waxlamp:
I'm not clear on what the exact problem is. Can you clarify the issue description to state what you think should be happening (in addition to what does in fact happen)?

JackWilb:
@waxlamp, When you navigate to a page the technically shouldn't exist, for example trying to navigate to multinet.app/foo/bar, the app should alert users to the fact there is an error. Right now, it just redirects you to the homepage. This would usually be a 404 error so we should make a 404 error page. 

Also if you navigate to https://multinet.app/nodelink/123/ it returns a 404 error but does not return a 404 error page to alert users.

waxlamp:
Ok I gotcha now. I'll edit the title/description to reflect this.

PR #241, opened 2020-01-08
Add simple redirect on invalid workspaces

### LINKED ISSUE #240: Display new page or redirect if invalid workspace name entered
When navigating multinet, if you enter the url of a workspace that does not exist (or perhaps that you just deleted but still have in your browser url), it will display the page as if the workspace is just empty. This could be confusing and causes problems if the user then tries to add tables (as the requests fail). A solution to this would be to display a workspace not found page, but a more immediate solution would be to simply redirect to the home page, like is already done for catching incorrect URLs.

Screenshot from 2020-01-08 11-03-40

AlmightyYakob:
Fixes #240. This is a simple redirect back to the home page. If it's thought that more needs to be done we can revise this.


AlmightyYakob:
> Can you run the linter to see if it makes any changes? (I see missing semicolons, for instance.) If you go into the `client` directory and run `yarn lint`, that should do the trick.

My bad, I guess I still need to get vscode completely setup. I've fixed the issues.

waxlamp:
The only remaining issue is the use of `router.push()` -- I just tested and it degrades the back button UX quite a bit. Since `push()` creates a new history entry, the back button now goes back to the bad URL, then immediately navigates to the home page. This essentially makes it impossible to go back to a known good state prior to the bad URL.

I believe `router.replace()` is more appropriate here. What do you think?

In the future, I think we should handle this case with a custom 404 view rather than simply redirecting to the home page. We've discussed this in #185, so let's keep an eye out for how we can upgrade to that approach in the future.

AlmightyYakob:
> The only remaining issue is the use of `router.push()` -- I just tested and it degrades the back button UX quite a bit. Since `push()` creates a new history entry, the back button now goes back to the bad URL, then immediately navigates to the home page. This essentially makes it impossible to go back to a known good state prior to the bad URL.
> 
> I believe `router.replace()` is more appropriate here. What do you think?
> 
> In the future, I think we should handle this case with a custom 404 view rather than simply redirecting to the home page. We've discussed this in #185, so let's keep an eye out for how we can upgrade to that approach in the future.

Yes I see the issue, good catch. I'll fix that now

AlmightyYakob:
I agree with the thought on a 404 page. When the time comes, it should be fairly easy to replace this redirection, as both here and in the routes page that redirect this, we can instead of pointing to the home page just point to the new 404 page.

PR #244, opened 2020-01-09
Return all attributes when querying for a graph's nodes

waxlamp:
TODO:
- [x] Either add field filtering option, or open an issue around doing the same
- [x] Release multinetjs
- [x] Update multinetjs version number in client

### DIFF
--- a/multinetjs/src/index.ts
+++ b/multinetjs/src/index.ts
@@ -1,5 +1,10 @@
 import { Client } from './client';
 
+export interface TableRow {
+  _key: string;
+  _id: string;
+}
+
 export interface GraphSpec {
   edgeTable: string;
   nodeTables: string[];
@@ -7,7 +12,7 @@ export interface GraphSpec {
 
 export interface NodesSpec {
   count: number;
-  nodes: string[];
+  nodes: TableRow[];
 }
 
 export interface Edge {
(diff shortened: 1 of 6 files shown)


multinetjs/src/index.ts
```diff
@@ -1,13 +1,18 @@
 import { Client } from './client';
 
+export interface TableRow {
+  _key: string;
+  _id: string;
+}
+
```
JackWilb: Would this ever break? Can the TableRow have more than just _key and  _id? Will it have attributes?

waxlamp: The beauty of TypeScript "interfaces" is they're composable. That is, saying that a value is a `TableRow` means that it must have `_key` and `_id` keys, but can also have other things in it as well. If you had a specific type that just encoded someone's name and age, you could do:

```typescript
interface Person {
  name: string;
  age: number;
}
```

and then declare a specific table's rows as having type `TableRow & Person`. But in general, a `TableRow` can indeed have other attributes on it.

JackWilb: Awesome, thanks for the clarification

waxlamp:
Hey guys, after the multinetjs updates, I need another review here. Sorry about that.

PR #246, opened 2020-01-09
Fixed table header

### LINKED ISSUE #245: Add fixed-header to table on TableDetail view
`fixed-header` prop is not working on `v-data-table`

jtomeck:
Fixes #245 

This also tweaks the paging options a bit. Now that the table is full height, if there are less items in the table there is just white space. @waxlamp - Do you think it makes sense to update the "rows per page" options to have a minimum of 15 and go `[15, 20, 25, 30, -1]`?

### DIFF
--- a/client/src/views/TableDetail.vue
+++ b/client/src/views/TableDetail.vue
@@ -66,6 +66,8 @@
       </v-app-bar>
       <div class="wrapper">
         <v-data-table
+          fixed-header
+          height="calc(100vh - 123px)"
           class="table-details"
           :headers="dataTableHeaders"
           :items="dataTableRows"
--- a/client/src/views/TableDetail.vue
+++ b/client/src/views/TableDetail.vue
@@ -71,7 +71,11 @@
           class="table-details"
           :headers="dataTableHeaders"
           :items="dataTableRows"
-          :items-per-page="15"
+          :items-per-page="20"
+          :footer-props="{
+            itemsPerPageOptions: [5, 10, 15, 20, -1],
+            showFirstLastPage: true,
+          }"
         />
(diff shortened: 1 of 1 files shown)


waxlamp:
I am actually working on the serverside pagination fixes right now, and I was literally just about to look at how to adjust the rows per page options. I am thinking something more like `[10, 20, 50, 100]` (and skip the `-1` since that will cause problems for the typical use case for Multinet).

So I'd actually prefer if you kept the changes on pagination options to a minimum to avoid conflicts with my developing branch.

client/src/views/TableDetail.vue
```diff
@@ -66,10 +66,16 @@
...
+          height="calc(100vh - 123px)"
           class="table-details"
           :headers="dataTableHeaders"
           :items="dataTableRows"
-          :items-per-page="15"
+          :items-per-page="20"
+          :footer-props="{
+            itemsPerPageOptions: [5, 10, 15, 20, -1],
+            showFirstLastPage: true,
+          }"
```
waxlamp: Can you revert this? I am taking care of this paging functionality in #247, which will likely conflict with these changes.

What does `showFirstLastPage` do?

jtomeck:
I reverted the wrong commit initially, but I just corrected it in my last commit so that you don't have conflicts when rebasing.

client/src/views/TableDetail.vue
waxlamp: Just for completeness, I spoke to @jtomeck IRL and learned that `showFirstLastPage` simply includes the first and last page buttons in addition to previous and next in the paginator.

PR #249, opened 2020-01-09
Add script that populates example data into multinet instance

AlmightyYakob:
This is a script that automatically adds example data from the `data` directory to a multinet instance, creating both tables and graphs. It's invoked by running 

`python scripts/data.py populate [address]`

If no address is provided at the cli, it checks the environment variables `MULTINET_HOST` and `MULTINET_PORT`, defaulting to `localhost` and `5000` respectively.


scripts/data.py
```diff
@@ -0,0 +1,216 @@
...
+    global server_address
+    log_tabstop = 4
+    log_indent = 0
+
+    if address is not None:
+        address_parts = address.split(":")
+        address_parts = [x for x in address_parts if x]
+
+        if len(address_parts) != 2:
+            raise Exception("Invalid passed address.")
```
waxlamp: Probably want a more detailed error message, something like

```suggestion
            raise Exception("Address must be of the form HOSTNAME:PORT")
```

scripts/data.py
```diff
@@ -0,0 +1,216 @@
...
+
+        # Create Tables
+        for file in files:
+            table_name = file.stem
+
+            if table_exists(workspace, table_name):
+                log(
+                    f'Table "{table_name}" already exists, skipping...',
+                    indent=log_indent,
+                )
```
waxlamp: The use case we're targeting is that this very script created the workspace, which means it should be empty. If this error condition occurs, something has gone very wrong, and we should abort with a fatal error message, rather than simply skip.

scripts/data.py
```diff
@@ -0,0 +1,216 @@
...
+
+            associated_node_tables = set()
+
+            for row in rows:
+                associated_node_tables.add(row["_from"].split("/")[0])
+                associated_node_tables.add(row["_to"].split("/")[0])
+
+            create_graph(
+                workspace,
+                f"{'-'.join([edge_table, *associated_node_tables])}",
```
waxlamp: Can we instead have the graph named the same as the workspace? That will give us better standard names for the graphs, rather than having to name them by the tables, which will inevitably lead to confusion.

AlmightyYakob: The reason that I did this is to allow for the creation of multiple graphs if that's desired. However, if that's not a realistic use case then I'd be happy to make he change.

scripts/data.py
AlmightyYakob: Good point. I'll change that.

scripts/data.py
waxlamp: Let's go with the simple case where we have one graph per dataset. As soon as we have a more complex dataset, we can (and will have to) revisit this.

PR #253, opened 2020-01-13
Automatically detect node tables in graph creation

### LINKED ISSUE #251: Automatically detect the required node tables for creating a graph
Currently, to create a graph, you have to specify the edge table and the node tables you want. However, selecting the node tables is redundant, as this information is already encoded in the edge table. Making this change would remove complexity in the api/backend, as well as on the client side.

AlmightyYakob:
Fixes #251. Currently, running `yarn serve` will throw a warning due to the new unpublished changes in `multinetjs`. Until the packages are updated I believe [`yarn link`](https://yarnpkg.com/lang/en/docs/cli/link/) can be used, but I've not been able to use it successfully.


client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "client",
-  "version": "0.1.0",
+  "version": "0.2.0",
```
waxlamp: Let's leave the client version as-is. We don't have a release process around the client application yet, so I'd like to leave the version number unchanged until we do.

multinet/api.py
```diff
@@ -138,72 +138,48 @@ def delete_workspace(workspace: str) -> Any:
...
     loaded_workspace = db.db(workspace)
     if loaded_workspace.has_graph(graph):
         raise AlreadyExists("Graph", graph)
 
-    existing_tables = set([x["name"] for x in loaded_workspace.collections()])
-    edges = loaded_workspace.collection(edge_table).all()
+    edge_table_properties = util.get_edge_table_properties(workspace, edge_table)
+    node_tables = edge_table_properties["table_keys"]
+    from_tables = edge_table_properties["from_tables"]
+    to_tables = edge_table_properties["to_tables"]
```
waxlamp: Could you add a comment to this block of code explaining what the util function is doing?

multinet/db.py
```diff
@@ -307,21 +310,47 @@ def aql_query(
...
+            e_def
+            for e_def in edge_defs
+            if e_def["edge_collection"] == edge_table and e_def != new_edge_def
+        ]
+
+        if conflicts:
+            # THIS SHOULDN'T HAPPEN
+            # TODO: Add error
+            print("CONFLICTING EDGE DEFINITIONS:", conflicts)
+            return False
```
waxlamp: If this is truly an exceptional condition (meaning that something is seriously wrong with the setup, or something we never anticipated has happened), then this is a good place to raise a sort of fatal error. In an ordinary Python program, I would do something like `raise RuntimeError(f"Conflicting edge definitions: {conflicts}")`; in this Flask server, I'm not completely sure what to do, but it's probably a new class of exception mapping to a 500-level error.

Let me know if you want to discuss this further. This is a small but important change to make since we want to know immediately what happened if we ever run into this error condition.

multinet/types.py
```diff
@@ -1,6 +1,15 @@
...
 EdgeDirection = Literal["all", "incoming", "outgoing"]
 TableType = Literal["all", "node", "edge"]
+
+
+class EdgeTableProperties(TypedDict):
+    """Describes gathered information about an edge table."""
+
+    table_keys: Dict[str, Set[str]]
+    from_tables: Set[str]
+    to_tables: Set[str]
```
waxlamp: Some comments describing exactly what these mean would be good to have here.

multinet/util.py
```diff
@@ -4,14 +4,39 @@
...
+from .types import EdgeTableProperties
 
 TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../test/data"))
 
 
+def get_edge_table_properties(workspace: str, edge_table: str) -> EdgeTableProperties:
+    """Return extracted information about an edge table."""
+
+    loaded_workspace = db.db(workspace)
+    edges = list(loaded_workspace.collection(edge_table).all())
```
waxlamp: Instead of coercing to a list here, it would be safer to treat this as an iterable sequence. I don't think you need the whole list in the for-loop in line 26 below; correct me if I'm wrong.

multinet/util.py
```diff
@@ -4,14 +4,39 @@
...
+
+    tables_to_keys: Dict[str, Set] = {}
```
waxlamp: Might want to specify `Set[str]` here (as you do in types.py)?

multinet/util.py
```diff
@@ -4,14 +4,39 @@
...
+    from_tables = set()
+    to_tables = set()
+
+    for edge in edges:
+        nodes = (edge["_from"].split("/"), edge["_to"].split("/"))
```
waxlamp: The code in this block might be a bit more readable if you try

```suggestion
        from_node, to_node = (edge["_from"].split("/"), edge["_to"].split("/"))
```

Of course you'd have to modify the remainder of the block to follow suit.

multinet/util.py
```diff
@@ -4,14 +4,39 @@
...
+from typing import Sequence, Any, Generator, Dict, Set
 
 from . import db
 from .errors import DatabaseNotLive, DecodeFailed
+from .types import EdgeTableProperties
 
 TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../test/data"))
 
 
+def get_edge_table_properties(workspace: str, edge_table: str) -> EdgeTableProperties:
```
waxlamp: In general throughout the function, a few brief comments explaining the large scale structure of what's happening would be helpful.

waxlamp: And now that I'm thinking about it, perhaps unit tests for this would be a good idea too.

We can add those in a future PR, but please open an issue around it.

multinetjs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "multinet",
-  "version": "0.5.0",
+  "version": "0.6.0",
```
waxlamp: Usually I like to defer the multinetjs version number bump until the moment I actually release it to NPM, but you can leave this instance of it intact.

client/package.json
AlmightyYakob: Sounds good

multinet/db.py
AlmightyYakob: Following our discussion offline I've pushed these changes, let me know what you think.

multinet/db.py
```diff
@@ -307,23 +308,31 @@ def aql_query(
 def create_graph(
     workspace: str,
     graph: str,
-    node_tables: List[str],
     edge_table: str,
     arango: ArangoClient,
+    from_vertex_collections: Set[str],
+    to_vertex_collections: Set[str],
```
waxlamp: I think the `arango` argument should go last in the list (to match the way the other `@with_client` functions are).

multinet/db.py
```diff
@@ -3,10 +3,10 @@
 
 from arango import ArangoClient
 from arango.database import StandardDatabase, StandardCollection
-from arango.exceptions import DatabaseCreateError
+from arango.exceptions import DatabaseCreateError, EdgeDefinitionCreateError
```
waxlamp: It looks like this is causing a typechecking error; the fix is to update the manual definitions of arango types in the top level `mypy_stubs` directory.

multinet/api.py
```diff
@@ -138,72 +138,50 @@ def delete_workspace(workspace: str) -> Any:
...
     if errors:
         raise ValidationFailed(errors)
 
-    db.create_graph(workspace, graph, node_tables, edge_table)
+    db.create_graph(
+        workspace,
+        graph,
+        edge_table,
+        from_vertex_collections=from_tables,
+        to_vertex_collections=to_tables,
```
waxlamp: If you change the argument declarations in the `create_graph` function, then you can pass these two arguments directly as positional arguments, rather than having to use the kwarg style.

AlmightyYakob: Alright. I kept them that way for readability, but I'm not really concerned about it.

waxlamp: I'm definitely more concerned about `arango` not being the last argument in the signature, and when that's fixed I'm not even sure we'd be able to pass these args as keywords. But even then, I think it actually degrades readability to only pass some of the arguments as keywords without a good reason.

PR #254, opened 2020-01-13
Improve newick uploader efficiency

AlmightyYakob:
While working on other tasks I noticed that in the `validate_newick`, we were tracking duplicates by appending to lists and checking if new items were in these lists. Since checking for the existence of an item in a list has average O(n) time complexity, if this list grows significantly this would become slow. This PR replaces that usage by using sets, which have average O(1) complexity. I also substituted f-strings in a section that used the old string formatting style.

Please let me know what you think!

### DIFF
--- a/multinet/uploaders/newick.py
+++ b/multinet/uploaders/newick.py
@@ -10,7 +10,7 @@ from ..util import decode_data
 from flask import Blueprint, request
 from flask import current_app as app
 
-from typing import Any, Optional, List, Dict
+from typing import Any, Optional, List, Dict, Set, FrozenSet, Tuple
 
 bp = Blueprint("newick", __name__)
 bp.before_request(util.require_db)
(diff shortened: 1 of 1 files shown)


multinet/uploaders/newick.py
```diff
@@ -19,41 +19,45 @@
...
     """Validate newick tree."""
     data_errors: List[Dict[str, Any]] = []
-    unique_keys: List[str] = []
-    duplicate_keys: List[str] = []
-    unique_edges: List[dict] = []
-    duplicate_edges: List[dict] = []
+    unique_keys: Set[str] = set()
+    duplicate_keys: Set[str] = set()
+    unique_edges: Set[FrozenSet[Tuple[str, object]]] = set()
+    duplicate_edges: Set[FrozenSet[Tuple[str, object]]] = set()
```
waxlamp: What's a `FrozenSet`?

AlmightyYakob: It's an immutable and hashable type which can be constructed from iterables. Since dictionaries themselves can't be hashed (because they're mutable), a frozenset is required to maintain data integrity while getting the benefits of hashing.

Here's an example of the conversion:

```
>>> a = {"a": 1, "b": 2, "c": 3}
>>> frozenset(a.items())
frozenset({('b', 2), ('a', 1), ('c', 3)})

```

waxlamp: Cool. Did not really know about these before, thanks!

PR #256, opened 2020-01-14
Standardize validation response

AlmightyYakob:
This is an overhaul to the way we currently handle the validation response. This both defines a new general structure for the validation response, as well as the types for each different validation error. Some of these errors I did no work on prior to this change but I tried my best to keep the semantics. I'm starting this as a draft PR, as there's lots of things to discuss and iterate on here.

You'll notice in several places we instantiate a named type, as well as the `type` parameter during construction. An example is shown below:

https://github.com/multinet-app/multinet/blob/a2c770b8a1cbf25d4484992a2bed937e2f3c90ee/multinet/api.py#L206-L209

Ideally here the `type` parameter would not need to be supplied, as it's redundant, but I couldn't find a way to natively do this with a `TypedDict`. One thought I had was making constructors that just supply this variable. Let me know if you guys have suggestions on how to approach this.


reference.jsonc
```diff
@@ -0,0 +1,84 @@
+{
+    "errors": [
+        // Graph creation
```
JackWilb: Is this reference file just for us in the future or does it have some programmatic use?

multinet/uploaders/csv.py
```diff
@@ -36,29 +43,34 @@ def validate_csv(rows: Sequence[MutableMapping]) -> None:
...
+            fields: List[str] = []
             if not valid_cell.match(row["_from"]):
                 fields.append("_from")
             if not valid_cell.match(row["_to"]):
                 fields.append("_to")
 
             if fields:
                 # i+2 -> +1 for index offset, +1 due to header row
-                detail.append({"fields": fields, "row": i + 2})
+                invalid_syntax_errors.append({"fields": fields, "row": i + 2})
```
JackWilb: Do you need to instantiate this as a new CSVInvalidRow or is it automatically converted from a dict to that class?

reference.jsonc
AlmightyYakob: Yes I forgot to mention that. This is what I originally started with and used as reference as I kept going, thought it'd be good to include during the draft PR

multinet/uploaders/csv.py
AlmightyYakob: It's technically not required because the type matches that of `CSVInvalidRow` (i.e., instantiating that produces the same dictionary), but we should for consistency. I'll change that

AlmightyYakob:
> Jake and I discussed this IRL, and we think that switching from TypedDicts to dataclasses will eliminate some of the difficulties with type tags and whatnot and lead to cleaner, easier to understand code. Jake's take a stab at it now.

@waxlamp I think it came out pretty well. You'll notice I still use a couple of TypedDicts towards the bottom of `types.py`, which I think makes sense, as they're being used to compose sections of the higher order data classes, and they're much easier to cast, specifically in the [newick uploader](https://github.com/multinet-app/multinet/pull/256/files#diff-2871e7a39b3717bdfbcf3bac3e8c66eaR68). Let me know what you think.

waxlamp:
@AlmightyYakob, I added a few commits containing changes we talked about in person yesterday evening. Otherwise, this looks good to me. Please take a look at my changes, and then merge if you believe it's ready to go.

Thanks again for your hard work on this!

AlmightyYakob:
@waxlamp Looks good to me but I need another review because I merged in master.

waxlamp:
> @waxlamp Looks good to me but I need another review because I merged in master.

Did you merge in master manually, or use the "update branch" button on GitHub? I thought the latter wouldn't invalidate current reviews...

AlmightyYakob:
I used the update branch function. I also thought that but :man_shrugging:

Issue #260, opened 2020-01-20
Add ability to create multiple tables when uploading data


AlmightyYakob:
Currently, when we're uploading D3 JSON files we're assuming that all nodes semantically belong in one node table (a reasonable assumption). However, when developing the downloader for this data type I realized that if there are multiple node tables in a graph, then when trying to download the graph as d3 json, node ids would have to be represented as `table/id` (to prevent id collisions along multiple node tables). 

If you were to then try to upload this exported data, it would be added to one node table, possibly leading to lots of empty columns when displaying the table (due to this new table essentially being a union of several types of tables). It would also cause the keys to grow after repeated uploads, as it would treat the new id of `table/id` as the whole key, with an export after that resulting in `newtable/table/id`.

The solution to this is to keep track of the types of data (or different sets of fields) that are seen when processing uploads, to automatically split data into separate tables if this is detected. This also is an issue that would apply to data upload in general, and wouldn't need to be limited to just D3 JSON.

PR #262, opened 2020-01-21
Add first downloaders

AlmightyYakob:
This is a draft PR as it's main purpose is to gauge our reaction to the initial implementation of these downloaders. Downloading tables as a CSV file and graphs as a D3 JSON file are two of the biggest initial use cases, so they've been implemented first. After our discussion here we can determine what we should change in the structure itself, as well as what formats we should support export as. 

Currently the D3 JSON downloader is prone to #260, as we're planning on addressing that separately from this. However it's an edge case and it's unsure exactly how large the scope of that issue is.


multinet/downloaders/d3_json.py
```diff
@@ -0,0 +1,77 @@
...
+
+    space = db.get_workspace_db(workspace)
+    if not space.has_graph(graph):
+        raise GraphNotFound(workspace, graph)
+
+    loaded_graph = space.graph(graph)
+
+    node_tables = loaded_graph.vertex_collections()
+    for node_table in node_tables:
+        table_nodes = list(loaded_graph.vertex_collection(node_table).all())
```
waxlamp: Coercing the full collection to a list up front seems like a bad idea. Some graphs may have a *lot* of nodes.

multinet/downloaders/d3_json.py
```diff
@@ -0,0 +1,77 @@
...
+
+        for edge in edges:
+            source = edge["_from"]
+            target = edge["_to"]
+            source_match = pattern.search(source)
+            target_match = pattern.search(target)
+
+            if source_match and target_match:
+                source = "".join(source_match.groups())
+                target = "".join(target_match.groups())
```
waxlamp: What's the purpose of extracting the stuff before "_nodes" in the table names?

multinet/downloaders/swagger/csv.yaml
```diff
@@ -0,0 +1,21 @@
+Download a csv file to a table
```
waxlamp:
```suggestion
Download a table as a CSV file
```

multinet/swagger/template.yaml
```diff
@@ -203,3 +203,5 @@ tags:
     description: Table retrieval, inspection, creation, and deletion
   - name: uploader
     description: Uploaders for various input data formats
+  - name: downloader
+    description: Downloaders   for various input data formats
```
waxlamp:
```suggestion
    description: Downloaders for various input data formats
```

Also, "downloaders for input formats" sounds weird to me. These are just downloaders for tables and graphs in general, I think.

multinet/util.py
```diff
@@ -4,14 +4,25 @@
...
-from typing import Sequence, Any, Generator
+from typing import Sequence, Any, Generator, Dict
 
 from . import db
 from .errors import DatabaseNotLive, DecodeFailed
 
 TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../test/data"))
 
 
+def filter_unwanted_keys(rows: Sequence[Dict]) -> Sequence[Dict]:
```
waxlamp: This function actually returns a list, which as I mentioned elsewhere in the review, might be a bad idea. There should be a way to write this function to return a generator instead, which will help with unexpected memory blowup etc.

multinet/__init__.py
```diff
@@ -27,6 +27,9 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     app.register_blueprint(uploaders.nested_json.bp, url_prefix="/api/nested_json")
     app.register_blueprint(uploaders.d3_json.bp, url_prefix="/api/d3_json")
 
+    app.register_blueprint(downloaders.csv.bp, url_prefix="/api/csv")
```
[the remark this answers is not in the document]
waxlamp: Agreed that this should probably be

```suggestion
    app.register_blueprint(downloaders.csv.bp, url_prefix="/api/csv/download")
```

but also see the discussion below about where this section of API should ultimately go.

In general, I don't think it's right to set a full-on blueprint at the single route where it wants to be called. The point of setting this prefix is for *multiple* endpoints to be sequestered together under a common prefix, rather than giving the actual route to a single endpoint.

multinet/downloaders/csv.py
```diff
@@ -0,0 +1,46 @@
...
+
+# Import types
+from typing import Any
+
+
+bp = Blueprint("download_csv", __name__)
+bp.before_request(util.require_db)
+
+
+@bp.route("/<workspace>/<table>", methods=["GET"])
```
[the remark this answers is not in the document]
waxlamp: Jake and I did discuss this some (and he decided to publish a PR with this concept since it existed in code already) but we did both feel that using the existing API infrastructure would be good for this functionality. After looking at this I agree that the downloaders for tables should be found at `GET /{workspace}/tables/{table}/download` and those for graphs at `GET /{workspace}/graphs/{graph}/download`. I'm not sure if this is what you meant, @JackWilb, but I think it's the right choice. If/when we support multiple download file types, we can specify those with a `format` parameter.

Jake mentioned that our *uploaders* don't follow this format, and he has a point. In a future effort, we should probably find a way to distribute those `POST` endpoints into something like `POST /{workspace}/tables/{table}` with a `format` parameter specifying which type of file is coming. This might be difficult since some of our uploaders result in multiple tables being created, etc., so we'll have to determine what REST API best practices are for such a situation.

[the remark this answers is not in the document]
waxlamp: Just for completeness: Jake also pointed out that doing these endpoints the way we have done the uploaders has one upside--the documentation is nice and explicit in the swagger page. To do it the more traditional way (i.e. how I'm suggested the downloaders be done) means that the documentation for the different supported file formats will all have to be crammed into the single `/download` endpoint's entry, but I am ok with that tradeoff.

multinet/downloaders/d3_json.py
```diff
@@ -0,0 +1,77 @@
...
+from flask import Blueprint, make_response
+
+# Import types
+from typing import Any
+
+bp = Blueprint("download_d3_json", __name__)
+bp.before_request(util.require_db)
+
+
+@bp.route("/<workspace>/<graph>", methods=["GET"])
```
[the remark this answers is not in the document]
waxlamp: Agreed; see comment above.

multinet/downloaders/d3_json.py
AlmightyYakob: I mainly did that because it's additional information added by multinet when uploaded, and the `_nodes` section is redundant when exporting since these fields can only refer to node tables anyway. So in this case `table_nodes/key` would become `table/key`

multinet/swagger/template.yaml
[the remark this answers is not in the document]
AlmightyYakob: Nope that's a typo, I'll fix it

AlmightyYakob: Agree, I'll change it.

AlmightyYakob:
@waxlamp @JackWilb I've made the suggested changes, let me know what you think. 

I haven't changed the overall structure of the API regarding the discussion above, but if you guys want I can make the changes to the downloaders API in this PR (i.e. downloading tables at `GET /{workspace}/tables/{table}/download` and graphs at `GET /{workspace}/graphs/{graph}/download`), and either adding the format parameter now just to future proof, or I could leave it off until we need it. Let me know what you guys think.

multinet/util.py
```diff
@@ -12,15 +12,17 @@
...
 
-    for row in rows:
-        new_rows.append({k: v for k, v in row.items() if k not in unwanted})
 
-    return new_rows
+def generate_filtered_docs(rows: Sequence[Dict]) -> Generator[Dict, None, None]:
+    """Filter unwanted keys from all documents with a generator."""
+
+    for row in rows:
+        yield filter_unwanted_keys(row)
```
waxlamp: I think you can actually just write this as `(filter_unwanted_keys(row) for row in rows)` (note the parentheses in place of square brackets), in which case this function wouldn't be needed. But this is just an observation; no need to make that change here.

PR #265, opened 2020-01-27
Add an "about" panel that can optionally include a Git SHA

waxlamp:
This PR adds an "about multinet" panel, activated by a button in the sidebar. If, at build time, the top-level contains a file `git-sha`, the contents of that file will be injected into the about panel, giving us an easy way to verify which exact commit on GitHub was used to build a given deployment.

TODO:
- [x] find a better place for the "about" button (need @jtomeck's help for this)
- [x] update the deployment workflow to spit out a git-sha file before the build process

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,6 +39,7 @@ jobs:
       - run: cd client && yarn lint
       - run: cd client && yarn lint:test
       - run: cd client && yarn test
+      - run: "cd client && git rev-parse HEAD >git-sha"
       - run: cd client && yarn build
 
       # Build and test Multinet client library.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,7 +39,6 @@ jobs:
       - run: cd client && yarn lint
       - run: cd client && yarn lint:test
       - run: cd client && yarn test
-      - run: "cd client && git rev-parse HEAD >git-sha"
       - run: cd client && yarn build
 
       # Build and test Multinet client library.
--- a/client/src/components/AboutDialog.vue
+++ b/client/src/components/AboutDialog.vue
@@ -28,6 +28,16 @@
       </v-card-title>
 
       <v-card-text
+        class="px-4 pt-4 pb-1"
+        >
+        Multinet is a system for storing and processing <a
+          href="https://vdl.sci.utah.edu/mvnv/" rel="noopener
+          noreferrer">multivariate networks</a>. Learn more and explore the code
+        at <a href="https://github.com/multinet-app/multinet" rel="noopener
+          noreferrer">GitHub</a>.
+      </v-card-text>
+
+      <v-card-text
         v-if="gitSha"
         class="px-4 pt-4 pb-1"
         >
--- a/client/src/components/AboutDialog.vue
+++ b/client/src/components/AboutDialog.vue
@@ -1,0 +1,65 @@
+<template>
+  <v-dialog
+    class="ws-dialog"
+    v-model="dialog"
+    width="500"
+    >
+    <template v-slot:activator="{ on }">
+      <v-btn
+        class="ws-btn"
+        block
+        color="blue darken-3"
+        dark
+        depressed
+        large
+        v-on="on"
+        >
+        About Multinet
+      </v-btn>
+    </template>
+
+    <v-card>
+
+      <v-card-title
+        class="headline pb-0 pt-3"
+        primary-title
+        >
+        About Multinet
+      </v-card-title>
+
+      <v-card-text class="px-4 pt-4 pb-1">
+        Hello.
+      </v-card-text>
+
+      <v-divider />
+
+      <v-card-actions class="px-4 py-3">
+        <v-spacer />
+
+        <v-btn
+          color="grey darken-3"
+          dark
+          depressed
+          @click="dialog = false"
+          >
+          OK
+        </v-btn>
+      </v-card-actions>
+
+    </v-card>
+
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend({
+  data() {
+    return {
+      dialog: false,
+    };
+  },
+});
+
+</script>
--- a/client/src/components/AboutDialog.vue
+++ b/client/src/components/AboutDialog.vue
@@ -6,16 +6,12 @@
     >
     <template v-slot:activator="{ on }">
       <v-btn
-        class="ws-btn"
-        block
-        color="blue darken-3"
-        dark
-        depressed
-        large
-        tile
+        class="mt-n1 ml-n1"
+        icon
+        small
         v-on="on"
-        >
-        About Multinet
+      >
+        <v-icon size="18">help</v-icon>
       </v-btn>
     </template>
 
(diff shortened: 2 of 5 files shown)


JackWilb:
Using the test server and client locally, it seems like the sha in the UI isn't responsive to changes in the file; you have to trigger a full rebuild of the app to see the changes. I assume that's intended behavior though since the client will rebuild when the repo changes?

I actually quite like where the button is, but it might look better at the bottom of the sidebar and without the border radius.

waxlamp:
> Using the test server and client locally, it seems like the sha in the UI isn't responsive to changes in the file; you have to trigger a full rebuild of the app to see the changes. I assume that's intended behavior though since the client will rebuild when the repo changes?
> 

Since the git-sha mechanism is really only useful for the production instance, my thinking was there's not much reason to figure out how to make it responsive to changes in the git-sha, as opposed to only doing the injection one time on a full rebuild. For local builds, you wouldn't typically open the about panel to see what sha it was built from; you'd likely just rebuild the whole thing from a known checkout, and then you'd know what rev you're on just by using git.

> I actually quite like where the button is, but it might look better at the bottom of the sidebar and without the border radius.

Would you mind adding a commit to remove the border radius?

I kind of don't mind where it is right now either, but I want @jtomeck's advice on this since he'll have a better vision of the various places this could go (and maybe we'll all like some other non-obvious placement better).

JackWilb:
> Would you mind adding a commit to remove the border radius?

Done.

client/vue.config.js
```diff
@@ -2,17 +2,28 @@ const fs = require('fs');
...
 
 // Read in .env file.
 const env = dotenv.parse(fs.readFileSync(path.resolve('..', '.env')));
 process.env.FLASK_SERVE_PORT = process.env.FLASK_SERVE_PORT || env.FLASK_SERVE_PORT || 5000;
 
+// Look for a git-sha file; if found, inject the value found in it into the
+// application.
+let GIT_SHA = null;
+if (fs.existsSync('git-sha')) {
+  GIT_SHA = JSON.stringify(fs.readFileSync('git-sha').toString());
```
AlmightyYakob: Why is `JSON.stringify` required here if the contents are already a string? I tried removing it locally but the whole site failed to load when I did.

AlmightyYakob: There's currently a trailing newline in the string
```suggestion
  GIT_SHA = JSON.stringify(fs.readFileSync('git-sha').toString().trim());
```

client/src/components/AboutDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+    gitSha() {
+      return GIT_SHA;
+    },
+
+    gitShaURL(this: any) {
+      const {
+        gitSha,
+      } = this;
+
+      return `https://github.com/multinet-app/multinet/tree/${gitSha}`;
```
AlmightyYakob: Is providing `this: any` as a parameter a workaround for typescript? It's shouldn't be required for normal operation in Vue.

Also I think we should use `this.gitSha` directly in the return instead of setting a variable for it.

client/src/components/AboutDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+
+    </v-card>
+
+  </v-dialog>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+declare var GIT_SHA: string;
```
AlmightyYakob: Minor suggestion

```suggestion
declare const GIT_SHA: string;
```

client/src/components/AboutDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+
+export default Vue.extend({
+  data() {
+    return {
+      dialog: false,
+    };
+  },
+
+  computed: {
+    gitSha() {
```
AlmightyYakob: Adding a return type would be good
```suggestion
    gitSha(): string {
```

client/src/components/AboutDialog.vue
```diff
@@ -0,0 +1,97 @@
...
+      return GIT_SHA;
+    },
+
+    gitShaURL(this: any) {
```
AlmightyYakob: Another place for a return type
```suggestion
    gitShaURL(this: any): string {
```

client/vue.config.js
waxlamp: The way `DefinePlugin` works is by doing direct text replacement, so if GIT_SHA is meant to be a string, you need to define it here with its quotation marks. The cleanest way to do that is to call `JSON.stringify()` on it, since JSON strings come with quotation marks.

client/src/components/AboutDialog.vue
waxlamp: `this: any` is indeed to get around TypeScript. I am looking into alternative methods of defining Vue components that don't require this sort of workaround.

I tend to use the destructuring approach to pulling out reactive properties to provide a conventional and highly visible way to indicate what the actual dependencies of a particular property. It's something that I think is worth the extra verbosity.

client/src/components/AboutDialog.vue
waxlamp: Good catch.

waxlamp:
> I think it looks good! Just left a few questions and suggestions.
> 
> We also might consider specifying the branch that the SHA belongs to, as just being linked to a specific commit doesn't indicate which branch you're on in Github, unless I'm missing something. Just a suggestion.

I considered adding branch information but I'm not sure how helpful it is. In the vast majority of cases, the branch will just be "master" for our deployment, and even if it were something else, the commit is what can reconstitute the codebase, since the branch name is just an alias for a particular commit.

For now I'll add the workflow update step, and then in the future we can consider adding branch info. I'll file an issue so we can revisit it later.

.github/workflows/main.yml
```diff
@@ -39,6 +39,7 @@ jobs:
       - run: cd client && yarn lint
       - run: cd client && yarn lint:test
       - run: cd client && yarn test
+      - run: "cd client && git rev-parse HEAD >git-sha"
```
JackWilb: This won't generate the git-sha file on the server. For that we'll need to update the deploy tasks for multinet client

waxlamp: Oh yeah of course.

Ok I'll look into making that change. We'll probably want to merge that change before this PR then, right?

JackWilb: The app launches without the file, right? If so, it doesn't matter. If it won't build without the file then, yeah, it would need to be merged first

JackWilb: Actually, we should merge the deploy branch first so that it doesn't get stuck without the file because the commit hash hasn't changed on this branch between the updates.

JackWilb:
The help icon animation is clipping for some reason on firefox
image

.github/workflows/main.yml
waxlamp: Yeah that's what I was thinking. I am going to revert this change, though, because it's not necessary for CI. (The app builds properly whether or not the `git-sha` file is present.)

waxlamp:
I like @jtomeck's edit, though we should figure out the Firefox issue if we can. It would be good to get this merged so we can focus on the deployment aspect and just sort of move on from this relatively simple addition :stuck_out_tongue:

Can someone give a final review? Or an objection if we really want to tackle the FF issue first?

waxlamp:
> It looks good to go. Did you want to roll the CI back first though?

Oops, yeah. I just pushed a revert.

waxlamp:
Thanks @JackWilb. Now just need @AlmightyYakob's approval and I'll merge.

PR #266, opened 2020-01-28
Add flake8 plugins

AlmightyYakob:
This PR adds several flake8 plugins that should improve our codestyle.

I curated these from [this](https://github.com/DmytroLitvinov/awesome-flake8-extensions) list. Let me know if you think there should be any additions from this list. 

There's a couple of quirks included with this change:

https://github.com/multinet-app/multinet/blob/093484eb1d5674c44dee2d34439341acea16d5ce/multinet/api.py#L6-L7

For some reason mypy can't find the import for `use_args`, even though it's definitely in the `flaskparser` module with `use_kwargs`

https://github.com/multinet-app/multinet/blob/093484eb1d5674c44dee2d34439341acea16d5ce/multinet/api.py#L44-L45

Mypy complains about the unknown type of the passed argument. I believe this weakness existed anyway before this change, it just wasn't caught by mypy.

If you guys have any thoughts on the above issues, please leave a comment or suggestion. If not, I can take out the notes and leave the typecheck ingore.

TODO:
- [ ] Address mypy import error
- [ ] Address mypy expression type error


multinet/api.py
```diff
@@ -35,11 +37,13 @@ def get_workspace(workspace: str) -> Any:
...
-@use_kwargs({"type": fields.Str()})
+@use_args({"type": fields.Str()})
 @swag_from("swagger/workspace_tables.yaml")
-def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:
+def get_workspace_tables(args: Dict, workspace: str) -> Any:
     """Retrieve the tables of a single workspace."""
-    tables = db.workspace_tables(workspace, type)
+    # TODO: Fix mypy complaint about below expression.
+    table_type: TableType = args.get("type") or "all"  # type: ignore
+    tables = db.workspace_tables(workspace, table_type)
```
waxlamp: Can you explain why this change is positive? It changes the signature of the function in a way that makes it less legible, IMO.

waxlamp: I will mention, I understand why `type` is suggested to be changed, but the problem is that we're using `type` as a parameter for a public API function, and changing it to a non-keyword just to satisfy flake8 seems like a loss if it impacts the design of the API.

multinet/uploaders/nested_json.py
```diff
@@ -22,15 +22,14 @@ def analyze_nested_json(
     `data` - the text of a nested_json file
     `(nodes, edges)` - a node and edge table describing the tree.
     """
-    id = itertools.count(100)
+    _id = itertools.count(100)
```
waxlamp: I don't think it's good to use a leading underscore like this, since that has a specific conventional meaning in Python code. Let's go with `ident` instead.

multinet/util.py
```diff
@@ -81,10 +85,10 @@ def require_db() -> None:
         raise DatabaseNotLive()
 
 
-def decode_data(input: bytes) -> str:
+def decode_data(input_data: bytes) -> str:
```
waxlamp: Let's just change the name to `data`.

multinet/api.py
```diff
@@ -2,9 +2,11 @@
 from flasgger import swag_from
 from flask import Blueprint, request
 from webargs import fields
-from webargs.flaskparser import use_kwargs
 
-from typing import Any, Optional
+# NOTE: Not sure why mypy flags this
+from webargs.flaskparser import use_args, use_kwargs  # type: ignore
```
waxlamp: The reason is we're using our own type stubs to provide types for this module; see mypy_stubs/webargs/flaskparser.pyi for where to fix this error.

waxlamp: However, before fixing this, let's figure out if we can leave the `get_workspace_tables()` function more or less as it was before, which would obviate the need to use `use_args()` anyway.

multinet/api.py
```diff
@@ -35,11 +37,13 @@ def get_workspace(workspace: str) -> Any:
...
 @bp.route("/workspaces/<workspace>/tables", methods=["GET"])
-@use_kwargs({"type": fields.Str()})
+@use_args({"type": fields.Str()})
 @swag_from("swagger/workspace_tables.yaml")
-def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:
+def get_workspace_tables(args: Dict, workspace: str) -> Any:
     """Retrieve the tables of a single workspace."""
-    tables = db.workspace_tables(workspace, type)
+    # TODO: Fix mypy complaint about below expression.
+    table_type: TableType = args.get("type") or "all"  # type: ignore
```
waxlamp: I'm actually not seeing a mypy error/warning on this line. Do you have more info?

multinet/api.py
AlmightyYakob: The only reason that's it's positive is not having a variable named `type`. I agree it's a nuisance to  change all of this just for that. Ideally, we'd have a way to map the "type" key to a different variable name. I'd be okay adding an ignore to the type checker here though.

multinet/uploaders/nested_json.py
AlmightyYakob: Good point, I'll change it.

multinet/util.py
AlmightyYakob: Sounds good

multinet/api.py
AlmightyYakob: If I remove the `# type: ignore` I get the following error:
```
multinet/api.py:45: error: Incompatible types in assignment (expression has type "Union[Any, str]", variable has type "Union[Literal['all'], Literal['node'], Literal['edge']]")
```

AlmightyYakob: I'm guessing it's because it has no way of knowing what is assigned to `args["type"]`. However if we remove this change this error will go away

multinet/api.py
waxlamp: Ok, let's put this function back to how it was and add an ignore directive (with a comment explaining why).

multinet/api.py
waxlamp: This will become moot because of how we're fixing the problem of a parameter `type` in `get_workspace_tables()`.

multinet/api.py
waxlamp: Just for future reference: you need to tell mypy that `args.get("type")` will be "restricted" to the type it is expecting. Typescript has a specific way to do this, where you write a bool function whose purpose is to test a value and return `true` if it really is of the correct type, and then Typescript knows enough about that to mark the value as having the restricted type. I'm sure Python has a way to express this though I've never done it and I'm not sure what it's called in Pythonworld.

Also: I am somewhat concerned that I wasn't able to get mypy to raise this error locally. I think it could have to do with those stars in Pipfile maybe. But it's setting off my spider-sense.

AlmightyYakob: I had that same thought too but I'm not sure how to do it in python. Might be good to look into.

Did you re-run `pipenv install`? If it's installing from the lockfile it should install the same versions. Either way we should probably pin those.

AlmightyYakob:
For some reason the linting in CI is failing but it passes fine on my local. I'm not sure why that is but it's only happening in the `test` folder, which makes me think it's being excluded somewhere.

multinet/api.py
waxlamp: > Did you re-run `pipenv install`? If it's installing from the lockfile it should install the same versions. Either way we should probably pin those.

Whoops, this was the problem. Good catch.

waxlamp:
> For some reason the linting in CI is failing but it passes fine on my local. I'm not sure why that is but it's only happening in the `test` folder, which makes me think it's being excluded somewhere.

I did some detective work and figured this out.

When CI runs, it's running on the version of things that would be merged to master if we were to click the green button right now--that is, it's essentially what you would get if you merged master into the branch and then ran the tests.

You can try this yourself: create a local branch `test` off of the tip of `add-flake8-plugins`, then hit that branch with a `get merge origin/master`, re-run `pipenv install` (just to make sure you're up-to-date) and then run `pipenv run lint` and you'll see the same error reported by CI.

The solution here is to use the "update branch" button, fix the resulting errors, and then watch for passing CI.

I definitely learned something here!

AlmightyYakob:
Interesting, I wasn't aware of that. That's definitely good to know. Doing what you suggested fixed it.

PR #271, opened 2020-01-29
Fix the scripts for macos

### LINKED ISSUE #268: Can't start the test server on mac
The docker tries to bind to /var/ which throws an error on mac. This isn't mission critical but would be nice to fix going forward

JackWilb:
Closes #268


client/scripts/start.sh
```diff
@@ -15,7 +15,7 @@ realpath() {
 FILE_PATH=$(dirname $(realpath "$0"))
 source $FILE_PATH/../../.env.test
 
-nohup yarn serve --port 58080 >server.out &
+FLASK_SERVE_PORT=$FLASK_SERVE_PORT nohup yarn serve --port 58080 >server.out &
 echo $! >server.pid
```
AlmightyYakob: Why is it required to do this? Since we're already sourcing `.env.test` isn't this essentially a no op?

JackWilb: It shouldn't be, I'm just troubleshooting. Having some issues with sourcing the .env.test on the ubuntu runner. I'm spinning up a local ubuntu instance here to test

waxlamp:
With the new style of local testing, I think this can be closed without further action. @JackWilb, if not, please reopen the issue with some discussion.

JackWilb:
These issues remain. Starting the test server causes issues with the readlink command and the mounting paths for the arango data. I think it might be better to start fresh from a new branch and just fix those 2 issues though since this was quite a large change for a small problem

waxlamp:
> These issues remain. Starting the test server causes issues with the readlink command and the mounting paths for the arango data. I think it might be better to start fresh from a new branch and just fix those 2 issues though since this was quite a large change for a small problem. Let me know what you think, @waxlamp

Ok. Can you update the description of #268 with a screencap or transcript of the exact commands you're running and the exact error messages that result? You shouldn't have any problems with `readlink` if you're following the README instructions, so it's important to know exactly what is leading to your error conditions.

PR #272, opened 2020-01-30
Restructure packages/dev-packages

AlmightyYakob:


AlmightyYakob:
With this change we'll have to update the CI to install with the `-d` flag. Also, I would say we should probably move the `uwsgi` dep to the main packages section. It's not required for local development, but is for deploy, and would prevent unnecessary packages from being installed on the deploy server.

multinet/db.py
```diff
@@ -7,7 +7,7 @@
 from requests.exceptions import ConnectionError
 
 from typing import Callable, Any, Optional, Sequence, List, Set, Generator, Tuple
-from mypy_extensions import TypedDict
+from typing_extensions import TypedDict
```
waxlamp: Was this just straight up wrong before?

AlmightyYakob: Yep

AlmightyYakob:
> Go ahead and move `uwsgi` to the non-dev dependencies.
> 
> Also, let's update the deployment CI before merging this. I can do it myself, but just to make sure: I should change the `pipenv install` command in the multinet-deploy workflow to read `pipenv install -d` instead, is that right?

Actually no, at least I don't think. Once we move the `uwsgi` dep to the core packages section, everything that's required for _deploy_ will be there. However we would still need to update the regular CI with `pipenv install -d`, as that's the process that requires the linting and typechecking addons.

waxlamp:
> > Go ahead and move `uwsgi` to the non-dev dependencies.
> > Also, let's update the deployment CI before merging this. I can do it myself, but just to make sure: I should change the `pipenv install` command in the multinet-deploy workflow to read `pipenv install -d` instead, is that right?
> 
> Actually no, at least I don't think. Once we move the `uwsgi` dep to the core packages section, everything that's required for _deploy_ will be there. However we would still need to update the regular CI with `pipenv install -d`, as that's the process that requires the linting and typechecking addons.

Oh I see. In that case I'd say go ahead and update the CI as part of this PR.

AlmightyYakob:
> Oh I see. In that case I'd say go ahead and update the CI as part of this PR.

It actually looks like we already use the `--dev` flag, so we should be good :+1:

AlmightyYakob:
@waxlamp @JackWilb I need another review after resolving conflicts

PR #284, opened 2020-01-31
Add sentry

AlmightyYakob:
This is a very simple change that adds sentry error reporting. This will only notify of errors through sentry if an Exception bubbles up completely through the application (i.e. if the Exception would have been printed to the console).


multinet/__init__.py
```diff
@@ -1,15 +1,21 @@
...
+from sentry_sdk.integrations.flask import FlaskIntegration
 
 from typing import Optional, MutableMapping, Any, Tuple, Union
 
 from multinet import api
 from multinet import uploaders, downloaders
 from multinet.errors import ServerError
 
+SENTRY_DSN = os.getenv("SENTRY_DSN", default="")
+sentry_sdk.init(dsn=SENTRY_DSN, integrations=[FlaskIntegration()])
```
waxlamp: Can you lowercase `SENTRY_DSN`?

.env.default
```diff
@@ -5,3 +5,5 @@ FLASK_APP=multinet
 FLASK_ENV=development
 
 MYPYPATH=mypy_stubs
+
+SENTRY_DSN=
```
waxlamp: These lines don't seem necessary due to how `os.getenv()` is called during app startup. Is this just to show the general structure of this file?

AlmightyYakob: Yeah that was my main reason. If we don't like this we could just list this in the docs, along with other env vars if we're not currently listing them.

multinet/__init__.py
AlmightyYakob: Sure

.env.default
waxlamp: No worries, we can leave this as is.

AlmightyYakob:
> Approving ahead of the changes I requested, since they're pretty straightforward.
> 
> Do we need to update deployment with SENTRY_DSN secrets prior to merging this? How did you manage to test this in production yesterday?

Yes, however I've already updated the deploy repo with that secret. I didn't test this in production, only on my local to ensure that the sdk initialized properly.

waxlamp:
> > Approving ahead of the changes I requested, since they're pretty straightforward.
> > Do we need to update deployment with SENTRY_DSN secrets prior to merging this? How did you manage to test this in production yesterday?
> 
> Yes, however I've already updated the deploy repo with that secret. I didn't test this in production, only on my local to ensure that the sdk initialized properly.

Cool, so upon merging this, sentry should go live with our production instance right?

AlmightyYakob:
> > > Approving ahead of the changes I requested, since they're pretty straightforward.
> > > Do we need to update deployment with SENTRY_DSN secrets prior to merging this? How did you manage to test this in production yesterday?
> > 
> > 
> > Yes, however I've already updated the deploy repo with that secret. I didn't test this in production, only on my local to ensure that the sdk initialized properly.
> 
> Cool, so upon merging this, sentry should go live with our production instance right?

Actually, I'm not sure anymore. Using Github secrets, the variable would be exposed through `secrets.SENTRY_DSN`. So we would probably want to inject that variable into the `.env` file.

AlmightyYakob:
> Actually, I'm not sure anymore. Using Github secrets, the variable would be exposed through `secrets.SENTRY_DSN`. So we would probably want to inject that variable into the `.env` file.

So I apparently already [did that](https://github.com/multinet-app/multinet-deploy/pull/42). After that is merged, sentry should work upon this PR merging.

Issue #291, opened 2020-02-03
Refactor each instance of the word "graph" to read "network" instead


waxlamp:
This specifically refers to the codebase: to avoid confusion between "graphs" and "networks" (the latter being the desired terminology for this project), we'll be sure to refer internally to "networks".

PR #292, opened 2020-02-03
Rename "graphs" to "networks" in UI, and make network list primary

### LINKED ISSUE #278: Rename "Graphs" to Networks, and update order
image

This should be "Networks" (we don't want to use the work Graph ever); and since Networks are the key thing, we want to put this to the left.

waxlamp:
Closes #278.

This PR doesn't refactor the use of the word "graph" from the codebase itself, just where it's visible within the UI (see #291).

Reviewers, please point out if I missed any of these.


client/src/views/WorkspaceDetail.vue
```diff
@@ -80,19 +80,21 @@
             text
           >
             <item-panel
-              ref="tablePanel"
-              title="Tables"
-              :items="tables"
+              ref="graphPanel"
+              title="Networks"
+              :items="graphs"
```
JackWilb: Why did you change the order of these? Better UX?

waxlamp: Alex requested this change to make "networks" front-and-center. (That's what I meant by "make network list primary" in the issue title, but that was probably unclear.)

JackWilb: Gotcha!

PR #295, opened 2020-02-03
Add a timeout to puppeteer

### LINKED ISSUE #294: Integration tests need a timeout
The integration tests will simply hang if something goes wrong. For instance, if a test is waiting for a specific element to show up, and it never does (due to, say, a change in the code), then the test will simply run forever. Instead, we should have some reasonable time limit (perhaps 60 seconds) after which the test should simply fail so the suite can move on.

JackWilb:
Adds a timeout to puppeteer and refactors the setup into its own function to reduce repetition. Closes #294


client/test/e2e.test.js
```diff
@@ -13,6 +13,16 @@ function browser(width, height) {
   });
 }
 
+// Sets up the browser with some default settings
+async function setup() {
+  const b = await browser(width, height);
+  const p = await b.newPage();
+  await p.setViewport({ width, height });
+  await p.goto('http://127.0.0.1:58080/');
+  await p.setDefaultTimeout(2000); // Default timeout of 2 seconds
```
waxlamp: The timeout should be longer, maybe 30 seconds. That's long enough to cover every reasonable case of delay, etc., but not long enough that hanging tests will cause a serious headache.

JackWilb: Updated to 30 seconds in 7b73cc9

PR #296, opened 2020-02-04
Add Google Analytics infrastructure

### LINKED ISSUE #283: Set up Google Analytics for the site.
Here's the include snippet: 

```
<!-- Global site tag (gtag.js) - Google Analytics -->
<script async src="https://www.googletagmanager.com/gtag/js?id=UA-157426670-1"></script>
<script>
  window.dataLayer = window.dataLayer || [];
  function gtag(){dataLayer.push(arguments);}
  gtag('js', new Date());

  gtag('config', 'UA-157426670-1');
</script>
```

waxlamp:
Closes #283

To make use of this, a file called `ga-tag` needs to be present in the "client/" directory containing the GA id. If present, the build process will read it and inject it into the app, which then uses the vue-gtag plugin to make it all work with the router.

This will require a change to the deployment playbooks before it will actually be used in production, so we need to hold off on merging it until that task is done.

TODO:
- [x] update deployment with ga-tag generation step


JackWilb:
Is the plan to have it come from a file for security purposes? So we'd inject it from secrets?

waxlamp:
> Is the plan to have it come from a file for security purposes? So we'd inject it from secrets?

Yep, it's set up now as a secret in multinet-deploy (following the SENTRY_DSN example).

AlmightyYakob:
Why not just use the environment variable directly instead of writing it to a file and loading it?

waxlamp:
> Why not just use the environment variable directly instead of writing it to a file and loading it?

No real reason I guess. I think it makes the process a bit more explicit. Otherwise, you have to add environment variables to the build line (e.g. `GIT_SHA=... GA_TAG=... yarn build`) which is less standard and therefore might interfere with services like Netlify (or other things relying on that "12 factor app" spec).

waxlamp:
> This all looks great but I don't see how it injects the google analytics code into the html. Is it happening with the VueGtag plugin that you made?

[This block](https://github.com/multinet-app/multinet/blob/c0d2a9cc0c719de8648befafcf4d13fb867b49bf/client/src/vuegtag.ts#L8-L14) is what does it. `GA_TAG` gets injected via vue.config.js and picked up here. The plugin manages the rest.

PR #297, opened 2020-02-04
Automatically create graph on relevant filetype uploaders

### LINKED ISSUE #277: Data Upload Master Issue
There should be two ways to upload a network: 

 * Upload a network file, such as a nested json or a d3-style json. This should create the network, a table for each node type, and a table for each edge type. No further assembly should be required. The initial data formats should be the d3-style json gefx: and https://gephi.org/gexf/format/
 * Upload a table of nodes and links from CSV. In a next step, I want to manually assemble how to link up the nodes based on the keys.

### LINKED ISSUE #280: CSV Upload silently fails
I just tried to upload this file: 
[movies.csv.zip](https://github.com/multinet-app/multinet/files/4136495/movies.csv.zip)

And it silently failed. I didn't get any feedback that it failed or why.

AlmightyYakob:
Fixes #277.
Fixes #280.
This automatically creates a graph in the `csv`, `d3_json` and `nested_json` uploaders.
Depends on #298 to be merged first.

TODO:
- [x] create a multinetjs release
- [x] update client to use new multinetjs version


multinet/uploaders/newick.py
```diff
@@ -82,14 +80,19 @@ def upload(workspace: str, table: str) -> Any:
     validate_newick(tree)
 
     space = db.db(workspace)
-    edgetable_name = "%s_edges" % table
-    nodetable_name = "%s_nodes" % table
+    if space.has_graph(graph):
+        raise AlreadyExists("graph", graph)
+
+    edgetable_name = "%s_edges" % graph
+    nodetable_name = "%s_nodes" % graph
```
waxlamp: Use an f-string here.

AlmightyYakob: There's actually a bunch of places where the old style string formatting is being used, so it might be better to open a separate PR that addresses this throughout the application

AlmightyYakob:
> It's hard to test the functionality of this without the UI, right? And do only csvs work in this branch until the other graph upload option is available in #298?

Following this and the suggestion in the other PR, that branch has been rebased off of this one, so you should be able to test out this functionality in the UI by checking out that branch.

multinet/uploaders/newick.py
waxlamp: I'm a bit nervous to do it that way, because if we introduce bugs while changing them all out wholesale, those will be harder to catch than if we do so on a case-by-case basis as we run across these instances.

But go ahead and open that issue, and I'll recommend we change this instance now since it's in front of us.

AlmightyYakob: I see what you mean. I'll change this.

AlmightyYakob:
This should be good to go @waxlamp

multinet/uploaders/newick.py
JackWilb: I can only see 4 instances, all in the newick.py uploader. I may have missed some though

Issue #274, opened 2020-01-30
Create Sample Views on the Landing Page


alexsb:
We want to show shortcuts with screenshots of demo applications and datasets on the landing page.

Something like this: 

![image](https://user-images.githubusercontent.com/2039375/73493016-1c183200-436f-11ea-8875-73c48671f8b3.png)
https://taggle.caleydoapp.org/

Doesn't have to be a fancy carousel though. 

We could call them "featured project" and use that down the line to feature new dataset.

jtomeck:
We have another application that has a screen full of sample data to use if you don't have your own to upload: https://kitware.github.io/paraview-glance/app/

Does the grid view shown here seem like an acceptable solution? The carousel is visually appealing, but I think it adds more work for the user to find what they're looking for, as most of their options aren't visible.

waxlamp:
@alexsb, can you provide some examples of what you might expect to see in the Multinet app? I assume things like Graffinity and Juniper, perhaps linked with example data so that clicking on it goes either to the network view for that dataset, or directly to the appropriate vis app?

If that's right, then we need a bit of guidance: right now, we just have the node-link diagram and the adjacency matrix views. We can set those up as the options on the landing page, but it's not clear to me what clicking on them should do. Can you give us some ideas there?

alexsb:
I think the grid is a fine solution. 

I would only want to show what is currently available in multinet, so right now that would be a bit boring, but that's OK, I'm sure the ParaView example didn't look much better 1 year in :).

alexsb:
And clicking them should actually load the view with this dataset, just like in paraview.

PR #299, opened 2020-02-05
Add sample datasets

waxlamp:
This PR adds the "boston" and "miserables" sample datasets to the repository, and enables the data population script to work properly with the now three sample datasets.

This is in service of solving #274: the next step will be to create some sample views for these datasets and have those appear on the landing page.


scripts/data.py
```diff
@@ -49,7 +49,9 @@ def get_edge_tables(workspace: str) -> List[str]:
 
 def get_table_rows(workspace: str, table: str) -> List:
     """Return the rows of a table."""
-    resp = requests.get(f"{root_api_endpoint()}/workspaces/{workspace}/tables/{table}")
+    resp = requests.get(
+        f"{root_api_endpoint()}/workspaces/{workspace}/tables/{table}/rows"
```
AlmightyYakob: Was this endpoint just outdated before?

waxlamp: Yeah, I made this change to the API as part of my proposed improvements, but I forgot to update your data uploader script.

PR #303, opened 2020-02-06
Add file format descriptions

### LINKED ISSUE #282: Provide Context for File Formats when uploading
Since we have a few non-standard file formats that we want to read, we might have to give a little context. This is the current state: 

image

 * [x] The drop down should be descriptive. For example, instead of newick, it should be "Newick, Tree format, Ext: .csv, .tsv, .txt" 
 * [x] After I've selected the drop-down, we want to show a bit of text about the file format below. That could be as simple as "gefx, The [Gephi Network File Format](https://gephi.org/gexf/format/)" (with the link); or as verbose as "D3-Style Json Network Format"  followed by an actual example.

AlmightyYakob:
This adds a "hint" to the file type selection field, that describes each file format.
Depends on #297.
Closes #282


JackWilb:
I re-ran the last test since it had failed. It passed this time, probably due to the new multinetjs version

jtomeck:
Nice! This is exactly what I would have done from a UI standpoint, my only "critique" is the multiline message nudges the other form field down. I think it's something we can take care of later, but wanted to at least comment about it. I'm not even sure how to get around it 🤷‍♂

AlmightyYakob:
> Approved, though this includes some big changes I made myself. @AlmightyYakob, please take a look at what I pushed and let me know if it's ok.

@waxlamp Those changes look great! Tested locally with no issues, I'm good to merge if you are :+1:

waxlamp:
> > Approved, though this includes some big changes I made myself. @AlmightyYakob, please take a look at what I pushed and let me know if it's ok.
> 
> @waxlamp Those changes look great! Tested locally with no issues, I'm good to merge if you are

Weapons free!

waxlamp:
> Nice! This is exactly what I would have done from a UI standpoint, my only "critique" is the multiline message nudges the other form field down. I think it's something we can take care of later, but wanted to at least comment about it. I'm not even sure how to get around it

I had the same thought. I opened #306 to address this.

PR #304, opened 2020-02-06
Sample views

### LINKED ISSUE #274: Create Sample Views on the Landing Page

jtomeck:
1. Moved getting started text to a popover next to the new workspace button
2. Created new structure for front page to display samples
3. Added collaborators section to banner (WIP)

Closes #274


waxlamp:
This is looking great.

Here are some ideas/suggestions:
- reduce the "collaborators" card to just the portion below the "fold" (i.e., just the institution logos)
- tighten the blue area so that rather than slotting into the full width of the panel, it just looks like a placard behind the header text

jtomeck:
> This is looking great.
> 
> Here are some ideas/suggestions:
> 
> * reduce the "collaborators" card to just the portion below the "fold" (i.e., just the institution logos)
> * tighten the blue area so that rather than slotting into the full width of the panel, it just looks like a placard behind the header text

@waxlamp all set!

waxlamp:
This is looking awesome! I will take some time to fill in actual text and then we'll see how it looks.

waxlamp:
@jtomeck I added some text and this is now ready for final review. I'm approving it but I'd like you to look at whether I did anything dumb.

In particular, I added `target="_blank"` to the `<v-btn>` component, and I'm assuming Vuetify does the right thing with `rel="noopener noreferrer"` but I couldn't confirm it by examining the DOM (when I ask to inspect the v-btn element, it disappears again when I move to the dev console to explore its markup).

PR #305, opened 2020-02-06
Auto populate name field on file select, misc improvments

### LINKED ISSUE #300: Automatically populate name field after selecting filename during file upload
Stems from #298, applies to the newly abstracted file upload component.

AlmightyYakob:
Fixes #300.
This automatically populates the name fields when a file is selected, and includes some misc formatting/code style improvements.

### DIFF
--- a/client/src/components/FileUploadForm.vue
+++ b/client/src/components/FileUploadForm.vue
@@ -21,7 +21,7 @@
             filled
             label="File type"
             v-if="typeList.length"
-            v-model="selectedType"
+            v-model="selectedFileType"
             :items="typeList"
           />
         </v-flex>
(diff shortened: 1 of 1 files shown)


client/src/components/FileUploadForm.vue
```diff
@@ -107,31 +107,43 @@ export default Vue.extend({
...
+      if (!file) {
+        this.selectedFileName = null;
+        this.selectedFileType = null;
+        this.fileUploadError = null;
+        return;
+      }
+
+      const fileInfo = this.fileInfo(file);
+      if (fileInfo !== null) {
+        [this.selectedFileName, this.selectedFileType] = fileInfo;
```
waxlamp: The autopopulation should only happen if the file name field is currently blank. The driving principle here is to never irreversibly lose any work the user may have done previously.

AlmightyYakob: Good point. I've just changed that.

PR #311, opened 2020-02-11
Simplify client testing

waxlamp:
This looks like a big change, but I really just removed the complexity of running the client tests, and refactored the existing tests into more logically scoped ones.

TODO
- [x] make sure this is working in CI


AlmightyYakob:
Under this approach, if one of the tests fails, will it leave behind created workspaces?

waxlamp:
> Under this approach, if one of the tests fails, will it leave behind created workspaces?

It definitely can. This is still intended to be primarily a CI testing strategy, but it can certainly be helpful to run it locally as well.

JackWilb:
> > Under this approach, if one of the tests fails, will it leave behind created workspaces?
> 
> It definitely can. This is still intended to be primarily a CI testing strategy, but it can certainly be helpful to run it locally as well.

Also, each time the CI runs, it's a fresh system so it wouldn't affect CI, just local testing.

.github/workflows/main.yml
```diff
@@ -38,6 +38,8 @@ jobs:
       - run: cd client && yarn install
       - run: cd client && yarn lint
       - run: cd client && yarn lint:test
+      - run: pipenv run test-server-up
```
JackWilb: You could do this with yarn if you'd like, there is a script called test:server:up

waxlamp: No need for that script anymore, since this way is more direct.

PR #312, opened 2020-02-12
Add two more placards to front page

waxlamp:


waxlamp:
> LGTM. Just a note though, the Paul Revere Adjacency Matrix placard looks a bit zoomed in.

I agree. The adjacency matrix for that dataset has a long, skinny shape due to its bipartiteness. I'm not sure how to capture that without the image itself looking super zoomed out. Any ideas?

AlmightyYakob:
> I agree. The adjacency matrix for that dataset has a long, skinny shape due to its bipartiteness. I'm not sure how to capture that without the image itself looking super zoomed out. Any ideas?

Oh I see. Not sure what we could do then.

PR #313, opened 2020-02-12
Update logo assets

waxlamp:
Steve Jordan recently made some geometric corrections to the Multinet logo mark. This PR uses the updated version in the app.

Depends on #312.


jtomeck:
@waxlamp - The logo in the sidebar looks kind of fuzzy to me. Can we get an SVG from Steve?

waxlamp:
> @waxlamp - The logo in the sidebar looks kind of fuzzy to me. Can we get an SVG from Steve?

There is an SVG in the logo bundle from Steve. Can I just use that in place of the PNG image?

jtomeck:
> There is an SVG in the logo bundle from Steve. Can I just use that in place of the PNG image?

Yeah you should be able to. If it doesn't work, let me take a stab at optimizing the SVG file.

waxlamp:
> > There is an SVG in the logo bundle from Steve. Can I just use that in place of the PNG image?
> 
> Yeah you should be able to. If it doesn't work, let me take a stab at optimizing the SVG file.

Cool. Let me know how that looks to you.

JackWilb:
There's an issue on my local with firefox, but not chrome. See below:

image

waxlamp:
> There's an issue on my local with firefox, but not chrome. See below:
> 
> image
> 
> Just confirmed this on my mac, too, so it's definitely a browser problem.

@jtomeck, can you look into this?

PR #315, opened 2020-02-12
Convert fields with "filled" prop to "outlined"

### LINKED ISSUE #302: Use dense components for dialog form fields
I'd like to try using dense form fields in the table/network creation dialogs. This is a low priority item but I think the style of the app would be served well by this change.

### LINKED ISSUE #306: Multiline hint text in network creation dialog nudges second row of UI down

jtomeck:
@waxlamp - We had talked early on in the project about having the outlined style for the form fields once we move to Vuetify 2.0.

This also fixes #306 
Closes #302


waxlamp:
@jtomeck do you want to try addressing #302 on this branch as well?

waxlamp:
@jtomeck please go ahead and merge this if you're ready.

jtomeck:
@waxlamp Just added the dense props. Want to review it again? I honestly don't think it is any better or worse. The normal sized components seem to have more aesthetically pleasing margins and paddings, but other than that, I don't know whether to call this an improvement or not. Thoughts?

PR #316, opened 2020-02-13
Use generators in file downloaders

### LINKED ISSUE #270: Use streaming download to mitigate latency of downloading large tables/graphs

AlmightyYakob:
This is to prevent large memory usage when downloading large files. I've tested this locally with a large file and confirmed that this change uses less memory when downloading.

Closes #270


multinet/db.py
```diff
@@ -36,6 +36,7 @@
     host=os.environ.get("ARANGO_HOST", "localhost"),
     port=int(os.environ.get("ARANGO_PORT", "8529")),
 )
+RESTRICTED_KEYS = {"_rev", "_id"}
```
waxlamp: Does the linter complain if you use lowercase for this?

multinet/db.py
```diff
@@ -154,20 +155,55 @@ def workspace_table(workspace: str, table: str, offset: int, limit: int) -> dict
...
+    workspace: str, table: str, filter_keys: bool = False
+) -> List[str]:
+    """Get the keys of a table in a workspace."""
+
+    query = f"""
+    FOR d in {table}
+        LIMIT 0, 1
+        RETURN ATTRIBUTES(d)
+    """
+    cursor = next(aql_query(workspace, query))
```
waxlamp: Even though it's not supposed to happen, we need to guard against the case when there are no rows in the table. We should actually test what happens when we upload a CSV file that has only a header row, and no data rows.

multinet/db.py
```diff
@@ -154,20 +155,55 @@ def workspace_table(workspace: str, table: str, offset: int, limit: int) -> dict
...
+
```
waxlamp: I would move this blank line to below the if statement right after this.

multinet/downloaders/d3_json.py
```diff
@@ -25,52 +26,60 @@ def download(workspace: str, graph: str) -> Any:
     `graph` - the target graph
     """
 
-    nodes = []
-    links = []
-
+    table_nodes_pattern = re.compile(r"^([^\d_]\w+)_nodes(/.+)")
```
waxlamp: Can you put in a comment explaining how this regex works? Besides that, it doesn't seem like we should assume the node tables have that `_nodes` suffix right?

multinet/downloaders/d3_json.py
```diff
@@ -25,52 +26,60 @@ def download(workspace: str, graph: str) -> Any:
...
+
+                yield f"{comma}{json.dumps(edge)}"
+                comma = comma or ","
+
+    def d3_json_generator():
+        yield """{"nodes":["""
+        yield from node_generator()
+        yield """],"links":["""
+        yield from link_generator()
+        yield "]}"
```
waxlamp: Whoa this is neat.

We may want a version that doesn't include whitespace (since those bytes will grow linearly with total data size).

multinet/db.py
AlmightyYakob: No, I can change it.

multinet/downloaders/d3_json.py
AlmightyYakob: Yeah I can add a comment. This assumption mirrors the functionality from the `d3_json` uploader, but I agree we should also support the case where that suffix isn't present.

AlmightyYakob: Actually, this does support both cases currently. If the regex matches, it overwrites the `source` and `target` variables, but otherwise leaves the `source` and `target` variables (representing the source and target node tables) unchanged.

multinet/db.py
```diff
@@ -202,11 +202,17 @@ def workspace_table_keys(
...
+        keys = next(cur)
+    except StopIteration:
+        return []
+
     if filter_keys:
-        return [k for k in cursor if k not in restricted_keys]
+        return [k for k in keys if k not in restricted_keys]
 
-    return list(aql_query(workspace, query))
+    return list(cur)
```
waxlamp: Shouldn't this be 

```suggestion
    return keys
```

Since you ran the cursor already in like 208, if `filter_keys` isn't `True`, won't this just return an empty list?

PR #327, opened 2020-02-21
Add pipenv and yarn caching

### LINKED ISSUE #323: Implement yarn and pipenv caching for the CI
Using [this](https://github.com/actions/cache) action we should start caching to speed up our builds. The yarn specific instructions are [here](https://github.com/actions/cache/blob/master/examples.md#node---yarn) and the pipenv instructions (not fully implemented yet) are [here](https://github.com/actions/cache/issues/152).

JackWilb:
Closes #323


JackWilb:
To get this working, I've set the environment variable `PIPENV_VENV_IN_PROJECT=1` and added an exclusion for the folder it makes, `.venv` in the flake8 config. Let me know if that seems reasonable or if you want me to try figure out the default install directory to cache

waxlamp:
> To get this working, I've set the environment variable `PIPENV_VENV_IN_PROJECT=1` and added an exclusion for the folder it makes, `.venv` in the flake8 config. Let me know if that seems reasonable or if you want me to try figure out the default install directory to cache.
> 

I'd prefer not to do this. Instead you can just run `pipenv --venv` in the project's top-level and it will tell you the path to the virtual environment.

> Also, the pipenv cache caches the whole install python 3 install, which might not be ideal. Additionally, it only saves about 20%-40% of the install time so it might be simpler to let it install fresh each time.

What's the problem with caching the Python 3 install? Disk space is not at a premium is it?

JackWilb:
> > To get this working, I've set the environment variable `PIPENV_VENV_IN_PROJECT=1` and added an exclusion for the folder it makes, `.venv` in the flake8 config. Let me know if that seems reasonable or if you want me to try figure out the default install directory to cache.
> 
> I'd prefer not to do this. Instead you can just run `pipenv --venv` in the project's top-level and it will tell you the path to the virtual environment.

Sure, I'll give it a go.

> > Also, the pipenv cache caches the whole install python 3 install, which might not be ideal. Additionally, it only saves about 20%-40% of the install time so it might be simpler to let it install fresh each time.
> 
> What's the problem with caching the Python 3 install? Disk space is not at a premium is it?

It was mentioned in the pipenv issue on the caching action that it might cause unknown errors since it's not just the packages that get cached, it's the whole python env.

waxlamp:
> > > Also, the pipenv cache caches the whole install python 3 install, which might not be ideal. Additionally, it only saves about 20%-40% of the install time so it might be simpler to let it install fresh each time.
> > 
> > 
> > What's the problem with caching the Python 3 install? Disk space is not at a premium is it?
> 
> It was mentioned in the pipenv issue on the caching action that it might cause unknown errors since it's not just the packages that get cached, it's the whole python env.

Need some more detail on this. Can you quote the appropriate discussion or put a link here? "Unknown errors" sounds ominous, and not a place I want to go with out software.

JackWilb:
This is from the issue I linked to in the issue that this PR closes:

>I really dislike the virtualenv cache, since it's caching a lots of extra stuff, not only the dependencies. I am not sure what (terrible?) consequences this may cause, but I haven't had any issues with this settings in the past month

Let me know what you think

AlmightyYakob:
I'd say it mostly looks good. For the python caching, I think it'd be good to follow the pattern used the [issue](https://github.com/actions/cache/issues/152#issuecomment-584657684) you linked:

```
key: ${{ runner.os }}-${{ env.PYTHON_VERSION }}-virtualenvs-${{ hashFiles('Pipfile.lock') }} 
```

Which protects the CI from updates to the minor version of python.

JackWilb:
> I'd say it mostly looks good. For the python caching, I think it'd be good to follow the pattern used the [issue](https://github.com/actions/cache/issues/152#issuecomment-584657684) you linked:
> 
> ```
> key: ${{ runner.os }}-${{ env.PYTHON_VERSION }}-virtualenvs-${{ hashFiles('Pipfile.lock') }} 
> ```
> 
> Which protects the CI from updates to the minor version of python.

I had added this before but using the `${{ env.PYTHON_VERSION }}` variable in the code wasn't producing any output. The cache was just linux--virtualenvs, no python version

AlmightyYakob:
> I had added this before, but using the `${{ env.PYTHON_VERSION }}` variable in the code wasn't producing any output. The cache was just linux--virtualenvs, no python version

Gotcha. In that case I think it looks good.

PR #330, opened 2020-03-02
Update length query and remove duplicate code

AlmightyYakob:
This PR updates the way we count documents in arango, with the more efficient and preferred call. [Relevant documentation](https://www.arangodb.com/docs/stable/aql/examples-counting.html).

### DIFF
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -156,14 +156,8 @@ def workspace_table(workspace: str, table: str, offset: int, limit: int) -> dict
     """Return a specific table named `name` in workspace `workspace`."""
     get_table_collection(workspace, table)
 
-    query = f"""
-    FOR d in {table}
-        LIMIT {offset}, {limit}
-        RETURN d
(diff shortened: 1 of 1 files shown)


multinet/db.py
```diff
@@ -156,14 +156,8 @@ def workspace_table(workspace: str, table: str, offset: int, limit: int) -> dict
...
 
-    query = f"""
-    FOR d in {table}
-        LIMIT {offset}, {limit}
-        RETURN d
-    """
-
     count = workspace_table_row_count(workspace, table)
-    rows = aql_query(workspace, query)
+    rows = workspace_table_rows(workspace, table, offset, limit)
```
waxlamp: Is this change just to take advantage of the existence of `workspace_table_rows()`?

AlmightyYakob: Yeah, since the query here is exactly the same as the one used in `workspace_table_rows`

PR #332, opened 2020-03-03
Remove use of DefinePlugin, switch to VUE_APP env vars

AlmightyYakob:
This uses a more standard way of injecting env vars, and removes most of the complexity from `vue.config.js`. I've also created a `.env.default` file, as a template that describes the new env vars. 

Let me know what you guys think.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -17,6 +17,7 @@
     "test": "tape -r esm test/**/*.test.js | tap-spec"
   },
   "dependencies": {
+    "@types/node": "^13.7.7",
     "axios": "^0.18.1",
     "core-js": "^2.6.5",
     "material-design-icons-iconfont": "^5.0.1",
--- a/client/package.json
+++ b/client/package.json
@@ -17,7 +17,6 @@
     "test": "tape -r esm test/**/*.test.js | tap-spec"
   },
   "dependencies": {
-    "@types/node": "^13.7.7",
     "axios": "^0.18.1",
     "core-js": "^2.6.5",
     "material-design-icons-iconfont": "^5.0.1",
--- a/client/src/components/AboutDialog.vue
+++ b/client/src/components/AboutDialog.vue
@@ -71,7 +71,7 @@ import Vue from 'vue';
 
 import AExt from '@/components/AExt.vue';
 
-declare const GIT_SHA: string;
+const GIT_SHA: string = process.env.VUE_APP_GIT_SHA || '';
 
 export default Vue.extend({
   components: {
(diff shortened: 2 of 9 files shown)


client/package.json
```diff
@@ -17,6 +17,7 @@
     "test": "tape -r esm test/**/*.test.js | tap-spec"
   },
   "dependencies": {
+    "@types/node": "^13.7.7",
```
waxlamp: Why is this needed?

AlmightyYakob: I was getting an error when referencing `process.env` saying that process wasn't defined, so this is the fix.

AlmightyYakob: Actually I can't reproduce that error, it seems to be working if I remove that dependency (and do a fresh install). So I can remove this.

waxlamp: Looking at how Vue accomplishes this injection (and using `vue inspect` to look at the DefinePlugin section), it does seem like you shouldn't need to bring in any external support for this. Essentially, Webpack is injecting a `process.env` variable into the global scope, and so it should "just work" (though, I'm not entirely getting how TypeScript is ok with this, without some kind of `declare` statement :thinking:).

client/src/components/AboutDialog.vue
```diff
@@ -70,8 +70,7 @@
 import Vue from 'vue';
 
 import AExt from '@/components/AExt.vue';
-
-const GIT_SHA: string = process.env.VUE_APP_GIT_SHA || '';
+import { gitSha as GIT_SHA } from '@/environment';
```
waxlamp: No need to shift it to `GIT_SHA` (for consistency with the other uses).

AlmightyYakob: This is done to follow linting guidelines (and I only noticed it from the linter), as without this renaming, the `gitSha` computed prop shadows the original `gitSha` import.

JackWilb:
Looks like the client comes up during testing and on my local. However, running `yarn test:client:up` on my local fails because of one command, `curl -s -I --max-time 0.5 http://localhost:8080/api/workspaces` 404s.

client/src/components/AboutDialog.vue
waxlamp: Ah I see. This seems to be implemented poorly (my fault). I will open an issue to fix this (by passing in the SHA value as a prop).

waxlamp:
> Looks like the client comes up during testing and on my local. However, running `yarn test:client:up` on my local fails because of one command, `curl -s -I --max-time 0.5 http://localhost:8080/api/workspaces` 404s.

Ah I figured it out: the point of this PR is that it detaches the server from the client altogether. So it's no longer viable to use localhost:8080/api to probe whether the client build has succeeded; however, we can do the same thing by just curl'ing localhost:8080 instead. See 8e3d7a5b6b1bf44e1a4e1e786e804c44f738ccbd.

PR #337, opened 2020-03-05
Prepare project for Heroku deployment

waxlamp:


setup.py
```diff
@@ -21,8 +19,21 @@
...
+        "requests",
+        "webargs",
+        "flask",
+        "flask-cors",
+        "flasgger",
+        "typing-extensions",
+        "uwsgi",
+        "sentry-sdk[flask]",
+        "gunicorn",
+    ],
```
AlmightyYakob: Why are these needed here? Since we already have the `requirements.txt`

waxlamp: Long story but read this article: https://caremad.io/posts/2013/07/setup-vs-requirement/.

I chatted with David Manthey about it this morning, and the bottom line is, this is the technically correct way to express the "install requirements", and is indeed redundant with the requirements.txt file, but for completeness I did it this way.

setup.py
```diff
@@ -21,8 +19,21 @@
         "Programming Language :: Python :: 3.6",
```
JackWilb: I know you haven't changed these in this PR, but why are there versions of python here that are incompatible with our app?

waxlamp: Good catch, I will fix that on this branch.

waxlamp: See 883b3c5

PR #338, opened 2020-03-05
Add ansible playbook for installing and running ArangoDB on AWS EC2 instance

waxlamp:


JackWilb:
Do we want to add a github actions yaml along with this?

waxlamp:
> Do we want to add a github actions yaml along with this?

To automatically update a running server? Probably not. This playbook is so simple as to be something that we could just run by hand when needed. (That reminds me that I need to add some documentation about how this is used.)

JackWilb:
> > Do we want to add a github actions yaml along with this?
> 
> To automatically update a running server? Probably not. This playbook is so simple as to be something that we could just run by hand when needed. (That reminds me that I need to add some documentation about how this is used.)

Will that be on this PR? If not, it's ready to go.

waxlamp:
> > > Do we want to add a github actions yaml along with this?
> > 
> > 
> > To automatically update a running server? Probably not. This playbook is so simple as to be something that we could just run by hand when needed. (That reminds me that I need to add some documentation about how this is used.)
> 
> Will that be on this PR? If not, it's ready to go.

I've got a general, self-assigned task to improve documentation. So, probably not on this PR.

PR #341, opened 2020-03-11
Fix workspace names

AlmightyYakob:
Fixes https://github.com/multinet-app/multinet-client/issues/13.
Adds backend support for https://github.com/multinet-app/multinet-client/issues/18.

These issues are addressed by adding a "mapping collection", which is a collection that resides in the `_system` database, and maps the workspace names to a valid generated workspace name on the backend.

If you were to check this out locally, you should be able to create a workspace with spaces/whatever you like in the name. However, note that if you do this and then switch to a different branch, you'll have workspaces with gibberish names (as they're no longer being mapped), so you'll just have to delete them.

Currently there's no UI component for renaming workspaces, but this PR adds this functionality through an API endpoint (`PUT /api/workspaces/{workspace}`), and making a change through the API will result in that change being shown in the UI.

Along with the normal review, please let me know if you think there's any missing features.


JackWilb:
There is UI for renaming workspaces. When you're in a workspace, the top bar has an edit symbol (on hover). Maybe it would be work connecting that up in this PR, too.

multinet/swagger/rename_workspace.yaml
```diff
@@ -0,0 +1,20 @@
...
+    required: true
+    schema:
+      type: string
+      example: Renamed Workspace
+responses:
+  200:
+    description: Workspace successfully renamed
+    schema:
+      type: string
+      example: workspace3
```
JackWilb: Let's add in the possible failures here

multinet/util.py
```diff
@@ -98,3 +101,8 @@ def data_path(file_name: str) -> str:
     file_path = os.path.join(TEST_DATA_DIR, file_name)
     print(file_path)
     return file_path
+
+
+def generate_arango_workspace_name() -> str:
+    """Generate a string that can be used as an ArangoDB workspace name."""
+    return f"{random.choice(ascii_lowercase)}-{uuid1()}"
```
waxlamp: Why pick a random letter to use as a prefix? And there is also a sort of randomness in uuid1, since it encodes the time as well. I think it would be better to simply use md5 or a sha to get a name. Technically, there can be name clashes, but in practice it's just not going to happen (and if it does, we can publish a paper with the colliding inputs :smirk_cat:).

Even better would be to figure out a reversible mapping between arbitrary names and legal Arango names. This would be a guard against database corruption (imagine if, e.g., the workspace_mapping table were lost but the data tables were all still there). To do that would require a closer study of Arango's naming rules, but it could be a good goal to have in mind since data integrity is going to matter long-term.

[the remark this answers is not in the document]
waxlamp: See my longer notes about this.

waxlamp:
> There is UI for renaming workspaces. When you're in a workspace, the top bar has an edit symbol (on hover). Maybe it would be work connecting that up in this PR, too.
> 
> If we want to leave it for later, it will be worth opening an issue for.

Now that the server and client are separated, this will have to be a separate issue + PR :smile_cat:

JackWilb:
> > There is UI for renaming workspaces. When you're in a workspace, the top bar has an edit symbol (on hover). Maybe it would be work connecting that up in this PR, too.
> > If we want to leave it for later, it will be worth opening an issue for.
> 
> Now that the server and client are separated, this will have to be a separate issue + PR 😸

Oh yeah :joy:

multinet/util.py
waxlamp: Here are Arango's rules for database names: https://www.arangodb.com/docs/stable/data-modeling-naming-conventions-database-names.html

AlmightyYakob:
> There is UI for renaming workspaces. When you're in a workspace, the top bar has an edit symbol (on hover). Maybe it would be work connecting that up in this PR, too.

I mean't it isn't connected to this change and does nothing right now

multinet/db.py
```diff
@@ -58,38 +58,114 @@ def check_db() -> bool:
...
+    """Add legacy workspaces to the workspace mapping."""
+    sysdb = db("_system")
+    coll = workspace_mapping_collection()
+
+    databases = {name for name in sysdb.databases() if name != "_system"}
+    registered = {doc["internal"] for doc in workspace_mapping_collection().all()}
+
+    unregistered = databases - registered
+    for workspace in unregistered:
+        coll.insert({"name": workspace, "internal": workspace})
```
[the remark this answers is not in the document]
AlmightyYakob: That should be possible, however I'm not sure how exactly we'd do it. We should probably have a separate discussion on this.

multinet/db.py
```diff
@@ -58,38 +58,114 @@ def check_db() -> bool:
...
+    """Return the collection used for mapping external to internal workspace names."""
+    sysdb = db("_system")
+
+    if not sysdb.has_collection("workspace_mapping"):
+        sysdb.create_collection("workspace_mapping")
+
+    return sysdb.collection("workspace_mapping")
+
+
+def workspace_mapping(name: str) -> Union[Dict, None]:
```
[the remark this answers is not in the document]
AlmightyYakob: Good point

multinet/db.py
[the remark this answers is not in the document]
waxlamp: See #342.

multinet/util.py
AlmightyYakob: The reason to pick a random letter is that it's a requirement of arango database names (they can't start with a number), so this is a way to get around that.

The problem with using a hash function is that renaming workspaces becomes much harder. I don't know how exactly we would actually implement it, but when changing the "external" workspace name, the hash function would produce a different "internal" name, which means we'd have to copy the arango database, which is something we should avoid. I'm not sure if that addresses your concern, but please lmk if I'm missing something.

multinet/api.py
```diff
@@ -137,6 +137,15 @@ def delete_workspace(workspace: str) -> Any:
     return workspace
 
 
+@bp.route("/workspaces/<workspace>", methods=["PUT"])
```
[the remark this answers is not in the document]
AlmightyYakob: Gotcha, I'll fix it

multinet/util.py
waxlamp: Ah I think I understand. So the idea is to select a legal ArangoDB name once, at the time the workspace is created, and then never again change that for the lifetime of the workspace? That is, even if the workspace name changes, only the "display name" changes, while the "internal name" would stay the same as it always was?

In that case, instead of picking a random letter, let's just use `w` (for workspace).

BTW, in order to rename the internal name as well, it would be sufficient to just rename the corresponding Arango database--I don't think you'd have to actually copy anything.

AlmightyYakob:
> One question: what happened to the `@lru_cache` I saw in one of the commits? My major concern with this approach is database performance. The caching mechanism might be one way to mitigate that, but I also feel like we'll need to run some tests to see if there's any difference in performance. If there is, we may need to converge to the alternative approach of programmatically computing the internal name from the external name, and shunt those back and forth via the type system. But this looks really great and solves our immediate problem.


I agree with your concerns, that's the biggest concern I had as well. The reason I got rid of the `@lru_cache` was because when I was using it, it seemed to be causing false cache hits, and breaking the behavior. I can experiment with this more to see if we can use this properly. That being said, when using it locally at least, the loading time isn't slower than is usual for a web app. Testing it against `multinet.app` it seems to run just as fast.

multinet/util.py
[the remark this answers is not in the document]
AlmightyYakob: Addressed below

waxlamp:
> > One question: what happened to the `@lru_cache` I saw in one of the commits? My major concern with this approach is database performance. The caching mechanism might be one way to mitigate that, but I also feel like we'll need to run some tests to see if there's any difference in performance. If there is, we may need to converge to the alternative approach of programmatically computing the internal name from the external name, and shunt those back and forth via the type system. But this looks really great and solves our immediate problem.
> 
> I agree with your concerns, that's the biggest concern I had as well. The reason I got rid of the `@lru_cache` was because when I was using it, it seemed to be causing false cache hits, and breaking the behavior. I can experiment with this more to see if we can use this properly. That being said, when using it locally at least, the loading time isn't slower than is usual for a web app. Testing it against `multinet.app` it seems to run just as fast.

Ok, let's keep away from this sort of lightweight caching then.

multinet/util.py
AlmightyYakob: Agree with your point on using just `w` as a prefix.

Regarding renaming the internal database name, I'm pretty sure that functionality doesn't exist. You'd have to dump and restore. See [this issue](https://github.com/arangodb/arangodb/issues/9143).

waxlamp: > Regarding renaming the internal database name, I'm pretty sure that functionality doesn't exist. You'd have to dump and restore. See [this issue](https://github.com/arangodb/arangodb/issues/9143).

Good lord.

multinet/swagger/rename_workspace.yaml
AlmightyYakob: Good point

multinet/db.py
```diff
@@ -58,38 +58,114 @@ def check_db() -> bool:
...
-        except DatabaseCreateError:
-            raise InvalidName(name)
-    else:
+
+    if workspace_exists(name):
         raise AlreadyExists("Workspace", name)
 
+    coll = workspace_mapping_collection()
+    new_doc = {"name": name, "internal": generate_arango_workspace_name()}
+    coll.insert(new_doc)
```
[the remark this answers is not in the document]
AlmightyYakob: I tried to reduce how much I repeated code, but I agree we can discuss in person

AlmightyYakob:
@waxlamp @JackWilb Lmk if you still have concerns I haven't addressed

waxlamp:
Before you merge--I just realized this doesn't actually fix https://github.com/multinet-app/multinet-client/issues/18. We'll need to add UI to the client to full solve that. Can you update the description before merging?

AlmightyYakob:
> Before you merge--I just realized this doesn't actually fix multinet-app/multinet-client#18. We'll need to add UI to the client hook up the existing UI to this functionality to full solve that. Can you update the description before merging?

Good catch, sure thing.

PR #351, opened 2020-03-19
Specify key csv

AlmightyYakob:
Backend support for https://github.com/multinet-app/multinet-client/issues/15.

---

One thing to note, after adding this support, I think we should split up the `validate_csv` function, as I'm no longer satisfied with the way it operates:

https://github.com/multinet-app/multinet-server/blob/f3bf95a1da2c257958837e00844f0ad20f769097/multinet/uploaders/csv.py#L48-L91

The current way it works is messy, and does the testing for both node and edge tables, but really they're mutually exclusive test cases, and have no real reason to be in the same function. Splitting them up to do their respective checking would eliminate the need for confusing conditionals against the `key` argument. I would have done this here, but I thought it'd be too big of a change, and thought it'd be better to do that in a separate PR.

---

Anyway, please lmk what you guys think.


AlmightyYakob:
Also I reworked the `test_csv_uploader` file a bit, as before all the tests were in one function, and it would have grown uncomfortably large with these additions. Hopefully that wasn't too off topic for this PR.

multinet/uploaders/swagger/csv.yaml
```diff
@@ -16,6 +17,13 @@ parameters:
         0,picard,captain
         1,riker,commander
         2,data,lieutenant commander
+  -
+    name: key
+    in: query
```
JackWilb: Is this shorthand for query string? Maybe it would be useful to new users to specify the whole thing?

AlmightyYakob: This just specifies to the api docs that it's a query param, not a body or path param.

AlmightyYakob:
> I'm not sure we should force users to use `_key` if it exists though. I think we should suggest it, if it exists, but I don't think it should throw a validation error if another key variable is specified

I see your point, but if we allow that then we'll be overwriting the `_key` field. This isn't a problem necessarily, if the user is fully aware of what they're doing, but I'm uneasy about it. I'd like to hear @waxlamp weigh in.

JackWilb:
> > I'm not sure we should force users to use `_key` if it exists though. I think we should suggest it, if it exists, but I don't think it should throw a validation error if another key variable is specified
> 
> I see your point, but if we allow that then we'll be overwriting the `_key` field. This isn't a problem necessarily, if the user is fully aware of what they're doing, but I'm uneasy about it. I'd like to hear @waxlamp weigh in.

Makes sense. Maybe a suitable solution would be to just rename the field the original `_key` column to something else and write the new column? Then nothing is overwritten and the users intentions are executed. I think Roni will have useful insight, though.

multinet/uploaders/csv.py
```diff
@@ -111,10 +122,13 @@ def upload(workspace: str, table: str) -> Any:
     # Read the request body into CSV format
     body = decode_data(request.data)
 
-    rows = list(csv.DictReader(StringIO(body)))
+    # Type to a Dict rather than an OrderedDict
+    rows: List[Dict[str, str]] = list(csv.DictReader(StringIO(body)))
 
     # Perform validation.
-    validate_csv(rows)
+    key = key or "_key"
```
waxlamp: Instead of this, why not just let the default arg value be `"_key"`? Then you could also type the `key` arg as `str` instead of `Optional[str]`.

multinet/uploaders/swagger/csv.yaml
```diff
@@ -6,7 +6,8 @@ consumes:
 parameters:
   - $ref: "#/parameters/workspace"
   - $ref: "#/parameters/table"
-  - name: data
+  -
+    name: data
```
waxlamp: Why break the line? YAML lets you create an object without this extra newline.

multinet/uploaders/swagger/csv.yaml
waxlamp: Same style comment as above.

waxlamp:
> > > I'm not sure we should force users to use `_key` if it exists though. I think we should suggest it, if it exists, but I don't think it should throw a validation error if another key variable is specified
> > 
> > 
> > I see your point, but if we allow that then we'll be overwriting the `_key` field. This isn't a problem necessarily, if the user is fully aware of what they're doing, but I'm uneasy about it. I'd like to hear @waxlamp weigh in.
> 
> Makes sense. Maybe a suitable solution would be to just rename the field the original `_key` column to something else and write the new column? Then nothing is overwritten and the users intentions are executed. I think Roni will have useful insight, though.

The best thing to do is add one more toggle in the UI, a checkbox that says something like "Overwrite _key field if it exists". If that checkbox is unchecked, and the uploaded file contains a _key field, then it should fail validation with something like "Existing _key field would be overwritten". On the serverside, this ought to be fairly simple to add--it requires one more `bool` argument to the upload and validation functions, and the validation function would wrap the new checks for existing _key field in an if-block conditioned on that argument.

What do you think, @AlmightyYakob?

AlmightyYakob:
> The best thing to do is add one more toggle in the UI, a checkbox that says something like "Overwrite _key field if it exists". If that checkbox is unchecked, and the uploaded file contains a _key field, then it should fail validation with something like "Existing _key field would be overwritten". On the serverside, this ought to be fairly simple to add--it requires one more `bool` argument to the upload and validation functions, and the validation function would wrap the new checks for existing _key field in an if-block conditioned on that argument.
> 
> What do you think, @AlmightyYakob?

That sounds like a good idea. Specifically, I think the default should be the unchecked state, to be safe.

multinet/uploaders/swagger/csv.yaml
AlmightyYakob: I just thought it was easier to read, and I'm not sure if there's a standard for this, but every yaml linter/validator that I've used has formatted it to be this way, so I thought it was preferred.

multinet/uploaders/swagger/csv.yaml
AlmightyYakob: Likewise

multinet/uploaders/csv.py
AlmightyYakob: Good point. I can change that.

AlmightyYakob:
> Do you want to add this feature in on the current PR, or in a followup?

I can add it here, it's pretty relevant to this change so I think it's warranted.

multinet/uploaders/swagger/csv.yaml
waxlamp: Huh if the linters like it better then I suppose we should keep it.

Maybe we should have a linter/validator for these files...

AlmightyYakob: There is [YamlLint](https://github.com/adrienverge/yamllint), which could be useful. If we're serious about this we should open an issue for it.

waxlamp: It's pretty low-priority but yeah we should do it. This is much more useful if there's also a validator for swagger YAML files.

test/test_csv_uploader.py
```diff
@@ -51,20 +51,34 @@ def test_invalid_key_field():
...
     assert len(validation_resp) == 1
     assert validation_resp[0] == correct
 
 
+def test_key_field_already_exists_b():
+    """
+    Test that specifying a key when one already exists results in an error.
+
+    (overwrite = True).
+    """
```
waxlamp: Is this description correct? Looks like this shouldn't/doesn't result in an error?

AlmightyYakob: You're correct, a product of copy and paste

multinet/uploaders/swagger/csv.yaml
waxlamp: @AlmightyYakob, did you open an issue for this?

AlmightyYakob: I did not, I'll do it now

waxlamp: #353

PR #355, opened 2020-03-25
Raise an exception when CSV reading fails

waxlamp:
Addresses https://github.com/multinet-app/multinet-client/issues/20


multinet/uploaders/csv.py
```diff
@@ -91,7 +99,10 @@ def upload(workspace: str, table: str) -> Any:
     # Read the request body into CSV format
     body = decode_data(request.data)
 
-    rows = list(csv.DictReader(StringIO(body)))
+    try:
+        rows = list(csv.DictReader(StringIO(body)))
+    except csv.Error:
+        raise CSVReadError()
```
AlmightyYakob: Do you have a file what would produce this error? Every time I try it raises other errors instead.

AlmightyYakob:
It seems like this one overlaps with #354? All of the changes there are included in this PR as well.

multinet/uploaders/csv.py
waxlamp: I used a gpg-encrypted file, but I think you can use any binary file (like some random PDF file you have lying around) and it's likely to trigger this.

waxlamp:
> It seems like this one overlaps with #354? All of the changes there are included in this PR as well.

Yeah it looks like this PR and that one are orthogonal. I thought this one depended on that one for some reason, but it's moot now since I've merged #354.

multinet/uploaders/csv.py
AlmightyYakob: Yep that did it, thanks :+1:

Issue #352, opened 2020-03-19
Split up validate_csv function to validate edge and node tables separately


AlmightyYakob:
Follow up from #351

waxlamp:
This is a great idea. One of the heuristics for code complexity is definitely function length. When you refactor this, please be sure to prioritize code legibility (i.e. so the rest of us + future you understand what the heck is going on in this function).

PR #356, opened 2020-03-26
Fix key specifier bug in csv uploader

AlmightyYakob:
I've left code comments in the affected areas, this is pretty much directly related to #352. Next steps after this is merged will be to address that issue.

### DIFF
--- a/multinet/uploaders/csv.py
+++ b/multinet/uploaders/csv.py
@@ -60,7 +60,7 @@ def validate_csv(
 
     fieldnames = rows[0].keys()
 
-    if key_field != "_key" and key_field not in fieldnames:
+    if key_field not in fieldnames:
         data_errors.append(KeyFieldDoesNotExist(key=key_field))
         raise ValidationFailed(data_errors)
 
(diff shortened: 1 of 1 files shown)


multinet/uploaders/csv.py
```diff
@@ -142,7 +142,12 @@ def upload(
     # Perform validation.
     validate_csv(rows, key, overwrite)
 
-    if key != "_key" and overwrite:
+    # Once we reach here, we know that the specified key field must be present,
+    # and either:
+    #   key == _key     # noqa: E800
+    #   or key != _key and the _key field is not present
+    #   or key != _key and _key is present, but overwrite = True
```
waxlamp: Could you surround the instances of the word `_key` in these comments by quotation marks? Unless you mean a variable by the name of `_key` (which I don't believe is the case).

AlmightyYakob: Good point

AlmightyYakob:
@waxlamp or @JackWilb I need another review. I had modified something which I shouldn't have due to the complex logic in that code. It's a good thing we have tests for that though :+1: otherwise I wouldn't have caught it.

PR #358, opened 2020-04-15
Add Oauth2/OpenID-Connect

AlmightyYakob:
This is a first pass at authenticating users via Google's OpenID-Connect provider.  At this point, it'll mainly be a discussion about the approach I took, and if/how we want that to change. There's definitely bugs, as I don't think I've covered all edge cases entirely, but I thought I'd open this up sooner rather than later, so that we can all get a handle on what's going on, and how we want to proceed.

This also doesn't actually touch the implementation of workspaces/tables/etc having users attached to them. I feel that those changes are different enough to warrant their own PR.


AlmightyYakob:
Also, I'll be sure to do some rebasing before the eventual merge, as right now the commit history isn't that clean.

AlmightyYakob:
@waxlamp @JackWilb This is ready for review

multinet/auth/google.py
```diff
@@ -0,0 +1,159 @@
...
+    oauth.init_app(app)
+    info = google_oauth2_info()
+    oauth.register(
+        name="google",
+        client_id=CLIENT_ID,
+        client_secret=CLIENT_SECRET,
+        access_token_url=info["token_endpoint"],
+        authorize_url=info["authorization_endpoint"],
+        api_base_url=GOOGLE_BASE_API,
+        client_kwargs={"scope": "profile email openid"},
```
JackWilb, line 86: >The scope parameter must begin with the openid value and then include the profile value, the email value, or both.
This is from the [docs](https://developers.google.com/identity/protocols/oauth2/openid-connect#scope-param)

JackWilb: Also, do we need profile and email? I assume email just gives email and profile gives name, picture, etc.?

multinet/auth/google.py
```diff
@@ -0,0 +1,159 @@
...
+    token = google.authorize_access_token()
+    rawinfo: GoogleUserInfo = parse_id_token(token["id_token"])
+    userinfo = filter_user_info(rawinfo)
+
+    loaded_user = load_user(userinfo)
+    if loaded_user is None:
+        user = register_user(userinfo)
+    else:
+        new_user: User = {**loaded_user, **userinfo}
+        user = save_user(new_user)
```
JackWilb, line 130: Is this the section that updates a user's info? If so, would `updated_user` be a better variable name here?

multinet/auth/types.py
```diff
@@ -0,0 +1,54 @@
...
+    # Time of token issue
+    iat: int
+
+    # Issuer
+    iss: str
+
+    aud: str
+    azp: str
+    at_hash: str
+    hd: str
```
JackWilb, line 30: I'd love a definition for these pieces, please. Also if they're not absolutely necessary, should we track them?

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return next(coll.find(inserted_info, limit=1))
+
+
+def set_user_cookie(user: User) -> User:
+    """Update the user cookie."""
+    new_user: User = dict(user)
+
+    new_cookie = uuid4().hex
+    new_user["multinet"]["session"] = new_cookie
```
JackWilb, line 66: Is there any chance of a collision here?

multinet/auth/types.py
```diff
@@ -0,0 +1,54 @@
...
+    sub: str  # Unique identifier
+
+
+class MultinetInfo(TypedDict):
+    """Data specific to multinet."""
+
+    session: str
+
+
+# TODO: Should we include ArangoDB document metadata in this?
```
JackWilb, line 50: I'd vote no if it can easily be retrieved from the database

multinet/auth/google.py
AlmightyYakob: > This is from the [docs](https://developers.google.com/identity/protocols/oauth2/openid-connect#scope-param). We might want to just change the order here to have parity with the spec.

Good catch, I'll do that and make sure it behaves like we expect.

> Also, do we need profile and email? I assume yes, and that email just gives email and profile gives name, picture, etc.?

Yes, I believe we'd need both of these.

multinet/auth/google.py
AlmightyYakob: Good point, I'll change it.

multinet/auth/types.py
AlmightyYakob: They're not necessary for the data that we store, but it does represent the response that we receive from Google, so I think it's worth keeping. The main reason I didn't explain most of those is because we don't actually use them, so I didn't think it necessary, but if we think it'll provide clarity I can add comments.

multinet/user.py
AlmightyYakob: Collision as in two users having the same session cookie? Not really, according to the [wikipedia page](https://en.wikipedia.org/wiki/Universally_unique_identifier), there are 5.3 undecillion possible UUID4 values, with the following anecdote:

> the probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion.

So we're probably fine.

JackWilb: Works for me, figured I'd check

multinet/auth/google.py
```diff
@@ -0,0 +1,159 @@
+"""Handling of Google Authorization."""
+import requests
+import base64
+import json
+
+from os import getenv
```
waxlamp, line 6: Is there a specific reason to pluck `getenv` out like this? In general I prefer for stdlib functions to be fully qualified (i.e., change this to just `import os`, and change `getenv()` throughout to `os.getenv()`).

Also, what's the reason for the linebreaks between the blocks of imports?

multinet/auth/google.py
```diff
@@ -0,0 +1,159 @@
...
+
+def parse_id_token(token: str) -> GoogleUserInfo:
+    """Parse the base64 encoded id token."""
+    parts = token.split(".")
+    if len(parts) != 3:
+        # Should probably throw something here
+        pass
+
+    payload = parts[1]
+    padded = payload + ("=" * (4 - len(payload) % 4))
```
waxlamp, line 53: Just curious, why do they make you pad this value before decoding it?

multinet/auth/google.py
```diff
@@ -0,0 +1,159 @@
...
+    resp = make_response(redirect(ensure_external_url(return_url)))
+    resp.set_cookie(MULTINET_COOKIE, cookie)
+
+    return resp
+
+
+@bp.route("/info")
+def user_info() -> ResponseWrapper:
+    """Return the filtered user object."""
+    # TODO: Filter out unwanted keys from the user object
```
waxlamp, line 146: We should either resolve all TODOs in code, or by linking to an issue in GitHub.

multinet/auth/types.py
waxlamp: Why would we include the metadata?

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+
+from multinet.db import db
+from multinet.auth.types import GoogleUserInfo, UserInfo, User
+
+from typing import Optional, Dict
+
+
+# This essentially duplicated the UserInfo TypedDict.
+# Is this a good reason to switch to dataclasses?
+USER_FIELDS = {"email", "name", "family_name", "given_name", "picture", "sub"}
```
waxlamp, line 14: Isn't there a way to extract the fields from `UserInfo`?

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+    coll = user_collection()
+    inserted_info: Dict = coll.update(user)  # type: ignore
+    return next(coll.find({"_id": inserted_info["_id"]}, limit=1))
+
+
+def register_user(userinfo: UserInfo) -> User:
+    """Register a user with the given user info."""
+    coll = user_collection()
+
+    # TODO: Fix mypy errors
```
waxlamp, line 54: Has this been resolved?

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+    new_cookie = uuid4().hex
+    new_user["multinet"]["session"] = new_cookie
+
+    return updated_user(new_user)
+
+
+def load_user_from_cookie(cookie: str) -> Optional[User]:
+    """Use provided cookie to load a user, return None if they dont exist."""
+    coll = user_collection()
+    document = list(coll.find({"multinet.session": cookie}, limit=1))
```
waxlamp, line 74: `.find_one()` again

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+    return user["multinet"]["session"]
+
+
+def filter_user_info(info: GoogleUserInfo) -> UserInfo:
+    """Return a subset of the User Object."""
+    return {k: v for k, v in info.items() if k in USER_FIELDS}
+
+
+def filter_document_meta(doc: Dict) -> User:
+    """Remove meta information from a document."""
```
waxlamp, line 97:
```suggestion
    """Remove ArangoDB metadata from a document."""
```

multinet/user.py
```diff
@@ -0,0 +1,98 @@
...
+
+    return user["multinet"]["session"]
+
+
+def filter_user_info(info: GoogleUserInfo) -> UserInfo:
+    """Return a subset of the User Object."""
+    return {k: v for k, v in info.items() if k in USER_FIELDS}
+
+
+def filter_document_meta(doc: Dict) -> User:
```
waxlamp, line 96: This function should probably be called `filter_document_metadata`, since (at least to my ears) `filter_document_meta` sounds like a meta operation that occurs above regular document filtering.

multinet/util.py
```diff
@@ -104,3 +106,19 @@ def data_path(file_name: str) -> str:
 def generate_arango_workspace_name() -> str:
     """Generate a string that can be used as an ArangoDB workspace name."""
     return f"w-{uuid1()}"
+
+
+def get_or_init_flask_secret_key() -> str:
```
waxlamp, line 111: You can just call this `flask_secret_key()`.

multinet/util.py
```diff
@@ -104,3 +106,19 @@ def data_path(file_name: str) -> str:
...
+    env_file_path = Path(".env")
+    dotenv.load_dotenv(env_file_path)
+
+    secret_key = dotenv.get_key(env_file_path, "FLASK_SECRET_KEY")
+
+    if not secret_key:
+        secret_key = uuid4().hex
+        ok, _, secret_key = dotenv.set_key(
+            env_file_path, "FLASK_SECRET_KEY", secret_key
+        )
```
waxlamp, line 122: We should probably have a live discussion about this. I'm still not convinced this should affect the .env file on disk. But beyond that, we are definitely breaking the dotenv protocols here (I think). I don't fully understand it, but I think there's a cascade of filenames that are checked, things like .env.local, etc., beyond which, the dotenv system supports different files in development and production. We're not making use of that sort of thing now, but whenever we depart from a standard, pain usually results in the future.

multinet/__init__.py
```diff
@@ -37,6 +41,9 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     app.register_blueprint(downloaders.csv.bp, url_prefix="/api")
     app.register_blueprint(downloaders.d3_json.bp, url_prefix="/api")
 
+    app.register_blueprint(google.bp, url_prefix="/google")
```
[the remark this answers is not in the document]
JackWilb: As an extension, we could even do `/user/oauth/google` since we may have more than 1 oauth provider – we could use github for example.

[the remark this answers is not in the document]
waxlamp: Right, that was I meant in case it wasn't clear: `/user/oauth/` could be the namespace for these blueprints; `/user/oauth/google` would be the path for google logins. This would allow things like `/user/info` (what Girder calls `/user/me`) and other things like that to slot in cleanly.

multinet/auth/types.py
AlmightyYakob: Sorry, this is a bit confusing. There's parts of the code where we use the `User` type, but there's actually extra arango document data in there, and the spots where that data is/isn't there is unclear right now. I'll have to address this.

multinet/user.py
```diff
@@ -0,0 +1,60 @@
...
+    return user[0]
+
+
+def register_user(userinfo, token):
+    """Register a user with the given user info, and token."""
+    coll = user_collection()
+
+    document = dict(userinfo)
+    document.update(
+        {"multinet_id": uuid4().hex, "token": token, "session": uuid4().hex}
```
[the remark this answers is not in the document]
AlmightyYakob: The revision that you commented on here is actually out of date. The `User` type documents this.

multinet/user.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Probably handle errors here
+    coll.insert(document)
+    return document
+
+
+def load_user_from_cookie(cookie):
+    """Use provided cookie to load a user, return None if they dont exist."""
+    coll = user_collection()
+    document = list(coll.find({"session": cookie}, limit=1))
```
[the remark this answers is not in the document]
AlmightyYakob: There actually doesn't seem to be one, so we'll have to just work around it.

multinet/user.py
```diff
@@ -0,0 +1,60 @@
...
+
+def user_exists(userinfo):
+    """Return the existance of a user."""
+    return load_user(userinfo) is not None
+
+
+def load_user(userinfo):
+    """Return a user doc if it exists, else None."""
+    coll = user_collection()
+    user = list(coll.find({"sub": userinfo["sub"]}, limit=1))
```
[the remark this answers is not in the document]
AlmightyYakob: I like that, I've fix it.

multinet/auth/types.py
AlmightyYakob: Added via d8b455f

.env.default
```diff
@@ -7,3 +7,12 @@ FLASK_ENV=development
 MYPYPATH=mypy_stubs
 
 SENTRY_DSN=
+
+# So urls can be routed consistently internally
+SERVER_NAME=http://localhost:5000
```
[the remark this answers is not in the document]
AlmightyYakob: On the backend, if we ever want to generate a URL for a specific route in our application (to be used externally), we need this to be set. This is used during the oauth process.

.env.default
```diff
@@ -7,3 +7,12 @@ FLASK_ENV=development
...
+
+GOOGLE_CLIENT_ID=
+GOOGLE_CLIENT_SECRET=
+
+# Will be set automatically by the server
+FLASK_SECRET_KEY=
```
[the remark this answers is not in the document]
AlmightyYakob: The main reason is that it persists between server start/stops, so the session of a user would be preserved. This wouldn't really matter in production, as it doesn't really restart, but for local dev it makes things easier. The production instance should still have access to the basic filesystem, so it should still work.

multinet/__init__.py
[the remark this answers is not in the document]
AlmightyYakob: I like this, I can reorganize it.

multinet/auth/google.py
AlmightyYakob: I can't find a clear reason why, it's just defined in the spec :man_shrugging:

multinet/auth/__init__.py
```diff
@@ -0,0 +1 @@
+"""Authorization types."""
```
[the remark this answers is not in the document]
AlmightyYakob: Looks like it's authorization according to the OAuth [website](https://oauth.net/). This makes sense, because what's really going on is that the account holder (whoever's logging in) is giving the requesting app (multinet) authorization to access their account info.

multinet/user.py
AlmightyYakob: See 2789632

multinet/util.py
AlmightyYakob: I've updated the behavior with ab3b7d7, lmk what you think

multinet/user.py
AlmightyYakob: There doesn't seem to be, as at runtime, any `TypedDict` is just a regular `dict`.

multinet/__init__.py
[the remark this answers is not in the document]
AlmightyYakob: See da5729c, lmk what you guys think.

.env.default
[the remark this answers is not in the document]
waxlamp: It turns out you can get this information from the Flask request, using something like [`url_root()`](https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/#werkzeug.wrappers.BaseRequest.url_root) (or possibly `host_url()`). That would let us not have to store this bit of info in the .env file, where it doesn't feel like it really belongs.

multinet/auth/__init__.py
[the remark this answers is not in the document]
waxlamp: :+1:

multinet/auth/types.py
waxlamp: Is this a problem for typechecking, or some other concern?

multinet/user.py
waxlamp: If you make `UserInfo` a dataclass, then I wonder if you could call [`.fields()`](https://docs.python.org/3/library/dataclasses.html) on the class itself and get this information.

multinet/auth/google.py
AlmightyYakob: Valid, I'll change it. As for the spacing, I tend to separate straight imports (`import os`) from "plucked" imports (`from os import getenv`). No real reason besides that, just seems stylistically  right.

multinet/util.py
```diff
@@ -104,3 +104,8 @@ def data_path(file_name: str) -> str:
 def generate_arango_workspace_name() -> str:
     """Generate a string that can be used as an ArangoDB workspace name."""
     return f"w-{uuid1()}"
+
+
+def flask_secret_key() -> str:
+    """Load or create a flask secret key."""
+    return os.getenv("FLASK_SECRET_KEY") or uuid4().hex
```
waxlamp, line 111: This function will return a different value each time it is called. Even if the function is only called once at startup, it would be better to secure it so it's idempotent. This is just to spare us future pain in case we ever do decide to call this function multiple times, not expecting it to return a random value each time.

I think I would do that by either calling `os.setenv()` (if such a thing exists), or by keeping a private variable here that you check for being `None` before setting it the first time.

.env.default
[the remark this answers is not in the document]
waxlamp: I believe we discussed/resolved this: we'll leave this value here so that local builds can make use of it, but in production we will generate our own random token and use Heroku to push it to the app.

In general, can we follow that procedure with all of the remaining values in this file? In particular, the `GOOGLE_CLIENT_*` values should all be pushed via Heroku, and we would possibly copy those to a local .env setup if we needed the login functionality?

multinet/user.py
AlmightyYakob: That's the main reason I proposed this. I'm currently giving this a shot to see what it looks like.

multinet/auth/types.py
AlmightyYakob: This is addressed in 15d5819. Now `User` includes arango metadata (`_id`, `_key`, `_rev`), and `FilteredUser` (which is what's returned by the API) strips them out.

.env.default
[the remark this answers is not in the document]
AlmightyYakob: Yes, you can just put those into your local `.env` file if you need them (which is what I've done throughout this entire PR).

multinet/user.py
AlmightyYakob: Addressed by 15d5819.

multinet/user.py
AlmightyYakob: Yes, by 15d5819.

JackWilb:
Also, is the CI not running because of the conflicts or for some other reason?

.env.default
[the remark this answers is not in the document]
waxlamp: We need to track the task of pushing these variables to Heroku, and also we need to document them somewhere. Could you open an issue listing the variables that must be set by us in production? We can make sure the deploy has them, and that we fix up our docs in the same stroke.

multinet/auth/types.py
```diff
@@ -0,0 +1,73 @@
...
+    Reference: https://developers.google.com/identity/protocols/oauth2/openid-connect#an-id-tokens-payload  # noqa
+    """
+
+    family_name: str
+    given_name: str
+    name: str
+    picture: str
+    locale: str
+    sub: str  # Unique identifier
+    email: str
```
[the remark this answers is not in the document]
waxlamp: We might not be able to do this. It will muddy up the types in a way that we wouldn't want. Basically, we could return or send in the wrong type from/to a function that expects the other, and the type system will allow it.

That's my speculation at least. It needs to be tested to see if that is actually as I think it is.

multinet/user.py
```diff
@@ -0,0 +1,117 @@
...
+
+    return from_dict(UserInfo, {k: v for k, v in info_dict.items() if k in fields})
+
+
+def filtered_user(user: User) -> FilteredUser:
+    """Remove ArangoDB metadata from a document."""
+    doc = dataclasses.asdict(user)
+
+    fields = {field.name for field in dataclasses.fields(FilteredUser)}
+    filtered = {k: v for k, v in doc.items() if k in fields}
```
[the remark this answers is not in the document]
waxlamp: The nice thing about dataclasses is they allow exactly this type of introspection. What's wrong with a comprehension?

waxlamp:
The changes all look pretty good to me. Let's get all the open threads resolved before merging.

multinet/__init__.py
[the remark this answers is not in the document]
waxlamp: Looks good. I think we will need swagger documentation for these endpoints right?

multinet/util.py
waxlamp: Thanks for this change.

The function still looks like it returns multiple values on consecutive calls (if the environment variable is unset). I mentioned this elsewhere, so if I can find that thread, I will resolve this one.

waxlamp: Ah found it: https://github.com/multinet-app/multinet-server/pull/358#discussion_r416002910.

waxlamp:
> Also, is the CI not running because of the conflicts or for some other reason?

I don't think it's because of merge conflicts. The Actions tab doesn't show this job at all so I assume something crashed and we'll have to find a way to kick off this job again (and/or contact GitHub support).

multinet/user.py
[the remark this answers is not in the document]
AlmightyYakob: > Do you know of any ways to just return the base dataclass fields without a comprehension?

The `dataclasses.fields` function returns the fields, it just also returns other info we don't care about, so we just grab the names. What do you mean by utilizing the inheritance?

multinet/__init__.py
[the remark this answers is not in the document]
AlmightyYakob: Correct

multinet/user.py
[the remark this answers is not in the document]
JackWilb: I feel like you should be able to do something like `filtered = parent_fields(doc)`. If something like that isn't possible, you can resolve this, Jake.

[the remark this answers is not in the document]
AlmightyYakob: I don't really see the benefit of what you're proposing, but I also don't know of a method that does that, so I'll resolve this.

multinet/__init__.py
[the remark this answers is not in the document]
AlmightyYakob: 892b70e

multinet/util.py
AlmightyYakob: b047f7f

AlmightyYakob:
@waxlamp @JackWilb In 949b98e I added a mypy ignore to the `authlib.integrations.flask_client` import, as I tried adding stubs under the directory:
```
mypy_stubs/authlib/integrations/flask_client.pyi
```

But mypy didn't seem to recognize them, and still threw the error:
```
multinet/auth/google.py:13: error: Skipping analyzing 'authlib.integrations.flask_client': found module but no type hints or library stubs
```

If either of you guys know how to, feel free to address this.

waxlamp:
> @waxlamp @JackWilb In [949b98e](https://github.com/multinet-app/multinet-server/commit/949b98eeb5f8c77608120b0c97b588c27931f98c) I added a mypy ignore to the `authlib.integrations.flask_client` import, as I tried adding stubs under the directory:
> 
> ```
> mypy_stubs/authlib/integrations/flask_client.pyi
> ```
> 
> But mypy didn't seem to recognize them, and still threw the error:
> 
> ```
> multinet/auth/google.py:13: error: Skipping analyzing 'authlib.integrations.flask_client': found module but no type hints or library stubs
> ```
> 
> If either of you guys know how to, feel free to address this.

Maybe you need an `__init__.pyi` file in the `authlib/` directory? I got the idea from the [mypy docs](https://mypy.readthedocs.io/en/stable/stubs.html#stub-files) on writing stub files.

AlmightyYakob:
> Maybe you need an `__init__.pyi` file in the `authlib/` directory? I got the idea from the [mypy docs](https://mypy.readthedocs.io/en/stable/stubs.html#stub-files) on writing stub files.

Yep that seems to have fixed it, I'll push a commit with the new stubs.

.env.default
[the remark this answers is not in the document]
waxlamp: #371

PR #363, opened 2020-04-22
Optimize database access

AlmightyYakob:
Some things I forgot to mention on our call:

* I added a profiling script, invoked through `pipenv run profile`. This basically runs `cProfile` on every request, allowing you to see what functions are taking long to complete, etc. This is run instead of `pipenv run serve`, when running locally.
* Some of the performance gains were due to commenting out the `util.require_db` call on every request. If we were to go with the idea of a "Health Indicator", this would be good to refactor into it's own api endpoint, instead of adding ~180ms per request.

Another thing we should discuss: Regarding the front-end querying, I think we all agreed to make the change to only send 1 request, for all of the tables, and do the filtering client side. Should that change be made here, and coordinated with the client repo? Or should we hold off on that? That would greatly reduce the time to load a workspace page.


JackWilb:
I'm about to look through the code, but I figured I'd answer your iniital questions first so I don't forget.

>     * I added a profiling script, invoked through `pipenv run profile`. This basically runs `cProfile` on every request, allowing you to see what functions are taking long to complete, etc. This is run instead of `pipenv run serve`, when running locally.

This is awesome functionality! Thanks for this change. I just ran it here and confirmed that it takes about ~400ms (after the first request) to query `api/workspaces/boston/tables` from the aws arango instance.

>     * Some of the performance gains were due to commenting out the `util.require_db` call on every request. If we were to go with the idea of a "Health Indicator", this would be good to refactor into it's own api endpoint, instead of adding ~180ms per request.

Would caching the response for a small amount of time fix this? We only really want to check it on the first request right? Another option would be to have the api poll the db on a certain cadence and just keep track of if it's responding as expected.

> Another thing we should discuss: Regarding the front-end querying, I think we all agreed to make the change to only send 1 request, for all of the tables, and do the filtering client side. Should that change be made here, and coordinated with the client repo? Or should we hold off on that? That would greatly reduce the time to load a workspace page.

I'm in favor of making this another PR. This PR will make the client much more usable and then reducing the number of queries would be an additional bump. I think they're distinct enough to split up.

multinet/db.py
```diff
@@ -82,6 +84,7 @@ def workspace_mapping_collection() -> StandardCollection:
     return sysdb.collection("workspace_mapping")
 
 
+@lru_cache()
 def workspace_mapping(name: str) -> Union[Dict[str, str], None]:
```
JackWilb, line 88: Is there a huge benefit to adding a cache here since it just calls out to another function you have cached and does a quick look up?

multinet/db.py
```diff
@@ -138,6 +141,10 @@ def rename_workspace(old_name: str, new_name: str) -> None:
     coll = workspace_mapping_collection()
     coll.update(doc)
 
+    # Invalidate the cache for things changed by this function
+    get_workspace_db.cache_clear()
+    workspace_mapping.cache_clear()
```
JackWilb, line 146: Does invalidating just the one workspace mapping cache cause the server to update the mapping_collection and its cache? It looks to me like we should invalidate the whole mapping_collection here instead, since it will now be out of sync with the server

multinet/db.py
```diff
@@ -202,20 +210,21 @@ def edge_table(fields: Sequence[str]) -> bool:
         return "_from" in fields and "_to" in fields
```
JackWilb, line 210: Looks like this and line above are no longer being used now that you use `properties()`

multinet/db.py
AlmightyYakob: Running the `find` command in this function makes a call to the database server, so caching the documents themselves here allows us to skip that call entirely.

multinet/db.py
AlmightyYakob: `workspace_mapping_collection` just returns the Collection itself, of type `StandardCollection`. Changes to any workspace mappings won't actually affect this, as you still have to call `find` on this collection to retrieve any data.

multinet/db.py
AlmightyYakob: Good catch

AlmightyYakob:
> Would caching the response for a small amount of time fix this? We only really want to check it on the first request right? Another option would be to have the api poll the db on a certain cadence and just keep track of if it's responding as expected.

Caching would fix it (I think), but I don't know when we'd invalidate it. We could do it on some sort of timer, but that feels messy. I think the main benefit we get out of that function currently is the graceful errors returned if the database goes down. This could be approached another way though, as really that error is only useful to us, and rarely (if ever) actually pops up in production. If we incorporated an endpoint to check this status from the client, that would also provide the functionality we're looking for.


> I'm in favor of making this another PR. This PR will make the client much more usable and then reducing the number of queries would be an additional bump. I think they're distinct enough to split up.

I agree, but we should keep in mind that it'll require synchronized changes regardless, as we'll be changing both the endpoint, and the access to it from the front-end.

multinet/db.py
JackWilb: Ah, thanks for the clarification.

multinet/db.py
JackWilb: Roger that

multinet/db.py
```diff
@@ -202,20 +210,21 @@ def edge_table(fields: Sequence[str]) -> bool:
...
 
     space = get_workspace_db(workspace)
+
     tables = (
-        (table["name"], edge_table(table_fields(workspace, table["name"])))
+        space.collection(table["name"]).properties()  # type: ignore
```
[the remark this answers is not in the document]
AlmightyYakob: I still have to address this, it's a function that we don't currently have in the type shed.

AlmightyYakob:
> 1. Is there a good way to stress test this to make sure our function caching strategies are working correctly?

I'm not quite sure, I feel like trying to measure this in CI would be kind of odd, and perhaps somewhat stochastic.

> 2. Some comments on the caching decorations would be valuable for the future (explaining why they're there, how much performance boost we got, etc.).

:heavy_check_mark: 

> 3. For documentary value, could you whip up a quick Google doc with a few performance numbers comparing current master and this branch? We've learned an interesting lesson here and I want to capture the essence of it.

Sure thing.

waxlamp:
> Another thing we should discuss: Regarding the front-end querying, I think we all agreed to make the change to only send 1 request, for all of the tables, and do the filtering client side. Should that change be made here, and coordinated with the client repo? Or should we hold off on that? That would greatly reduce the time to load a workspace page.

Why would this require a backend change? We would just replace the two calls in the frontend with one call right? Furthermore, it occurs to me that we could probably retain the ease of use and just make sure we fire both calls at once, rather than `await` them.

AlmightyYakob:
> Why would this require a backend change? We would just replace the two calls in the frontend with one call right? Furthermore, it occurs to me that we could probably retain the ease of use and just make sure we fire both calls at once, rather than `await` them.

If we were to change to using just 1 call for all tables, we'd have to change the response schema to differentiate between edge and node tables. Something like:
```
{
  "edge": [...],
  "node": [...]
}
```

instead of the flat list we return now. 

If we weren't to do that, we could do 2 parallel calls to these tables, but that would still be slower than making the 1 call, due to the server processing two requests at once, python's GIL, etc.

I'm in favor of changing the backend like in the first option I mentioned, as it reduces complexity of the API.

waxlamp:
> > Why would this require a backend change? We would just replace the two calls in the frontend with one call right? Furthermore, it occurs to me that we could probably retain the ease of use and just make sure we fire both calls at once, rather than `await` them.
> 
> If we were to change to using just 1 call for all tables, we'd have to change the response schema to differentiate between edge and node tables. Something like:
> 
> ```
> {
>   "edge": [...],
>   "node": [...]
> }
> ```
> 
> instead of the flat list we return now.
> 
> If we weren't to do that, we could do 2 parallel calls to these tables, but that would still be slower than making the 1 call, due to the server processing two requests at once, python's GIL, etc.
> 
> I'm in favor of changing the backend like in the first option I mentioned, as it reduces complexity of the API.

Good point. Can you open an issue for this, and maybe copy in these comments?

PR #367, opened 2020-04-24
Fix Caching

AlmightyYakob:
Addresses issues in the caching introduced by #363. Currently, the CI should fail, as it has no database to connect to, which is required for the tests introduced by this PR.


AlmightyYakob:
> Is there a distinction between using `workspace_mapping()` and `cached_workspace_mapping()` now? What I mean is, are there places where you must use one and not the other?

Example:
https://github.com/multinet-app/multinet-server/blob/7c309821094f4a9a04a53b97bdbc1805f0261c4b/multinet/db.py#L110-L112

In general, places where you don't want the use of `workspace_mapping` to have any effect on the cache. I could see wanting to rename `workspace_mapping` to `uncached_workspace_mapping` (or similar), and renaming `cached_workspace_mapping` to `workspace_mapping` (or similar), as we use the cached version more often.

waxlamp:
> > Is there a distinction between using `workspace_mapping()` and `cached_workspace_mapping()` now? What I mean is, are there places where you must use one and not the other?
> 
> Example:
> https://github.com/multinet-app/multinet-server/blob/7c309821094f4a9a04a53b97bdbc1805f0261c4b/multinet/db.py#L110-L112
> 
> In general, places where you don't want the use of `workspace_mapping` to have any effect on the cache. I could see wanting to rename `workspace_mapping` to `uncached_workspace_mapping` (or similar), and renaming `cached_workspace_mapping` to `workspace_mapping` (or similar), as we use the cached version more often.

So the problem here is the technical debt we're about to introduce: developers now have to know the *effect* of consulting the workspace mapping, which means our abstraction is leaking in a bad way. Renaming `workspace_mapping` to `uncached_workspace_mapping` would be good in that it would signal to devs that there's some choice to be made here, but it doesn't protect them from making the wrong call and breaking correctness.

It feels like this issue of asking for a workspace by name vs. creating a new entry or changing an existing entry is starting to become larger than can just be handled by the naive use of the lru cache decorator. I see two ways out of this:

1. The simpler way is to use the fix I suggested on Slack yesterday: to simply invalidate the cache whenever a renaming, deletion, or creation occurs, and let the cache handle the common case of lookups when the state isn't changing. This would be a bad option if `workspace_exists()` is used more often than I think it is.

2. The more complex, but probably *correct* way is to abstract these operations into a module, make the workspace mapping collection into a private variable, and export public functions to do the actual operations (create, rename, delete, and check for existence). The caching mechanism would probably become more explicit, like a dict we maintain in that same module between invocations of the public functions. Perhaps there's an actual LRU cache implementing object in the standard library we could use, but access at a finer granularity than the `@lru_cache` decorator does.

The other issue here is testing--we need to figure out how to run those tests in CI, because a feature like this really will require us to make sure it's working long term.

To summarize though: the divergence of our once easy-to-use function into explicitly cached and uncached variants represents a broken abstraction and an associated large amount of technical debt that I'm not willing to take on in the current form. This is a really important crossroads, so let's discuss what to do in good detail before we move on.

AlmightyYakob:
> So the problem here is the technical debt we're about to introduce: developers now have to know the _effect_ of consulting the workspace mapping, which means our abstraction is leaking in a bad way. Renaming `workspace_mapping` to `uncached_workspace_mapping` would be good in that it would signal to devs that there's some choice to be made here, but it doesn't protect them from making the wrong call and breaking correctness.
> 
> It feels like this issue of asking for a workspace by name vs. creating a new entry or changing an existing entry is starting to become larger than can just be handled by the naive use of the lru cache decorator. I see two ways out of this:
> 
> 1. The simpler way is to use the fix I suggested on Slack yesterday: to simply invalidate the cache whenever a renaming, deletion, or creation occurs, and let the cache handle the common case of lookups when the state isn't changing. This would be a bad option if `workspace_exists()` is used more often than I think it is.

Currently, this is what's being done. On calls to creating, renaming or deleting a workspace, the cache is invalidated. I'm not sure what you mean referring to `workspace_exists` being possibly called often? That function won't affect the cache, because it's not using the cached version of `workspace_mapping`.

> 2. The more complex, but probably _correct_ way is to abstract these operations into a module, make the workspace mapping collection into a private variable, and export public functions to do the actual operations (create, rename, delete, and check for existence). The caching mechanism would probably become more explicit, like a dict we maintain in that same module between invocations of the public functions. Perhaps there's an actual LRU cache implementing object in the standard library we could use, but access at a finer granularity than the `@lru_cache` decorator does.

I'm kind of weary to try to implement our own caching, as now we'd have to worry about general caching bugs, as well as bugs in the caching mechanism itself. Can you explain why you think we should stray away from `lru_cache`?

I do think we should be explicit about what's being cached, and where. For example, `cached_workspace_mapping` vs `workspace_mapping`. I think separating the two, and having `cached_` functions that just cache calls to the original function is a good approach, because it makes it clear that we're using caching, and allows us to use the original functions if we need to. If we were to do this, I'd have to (and should) apply the same change to `get_workspace_db`. Let me know what you think about this approach.

> The other issue here is testing--we need to figure out how to run those tests in CI, because a feature like this really will require us to make sure it's working long term.

I agree. If we were to just specify `ARANGO_HOST`, `ARANGO_PROTOCOL` and  `ARANGO_PASSWORD` in the environment where the tests run, they'd succeed (assuming we actually point to a valid arango instance). As far as what that should be, I'm not sure, but I think we can agree it should be separate from our production db.

JackWilb:
> I agree. If we were to just specify `ARANGO_HOST`, `ARANGO_PROTOCOL` and `ARANGO_PASSWORD` in the environment where the tests run, they'd succeed (assuming we actually point to a valid arango instance). As far as what that should be, I'm not sure, but I think we can agree it should be separate from our production db.

If we go for a full staging set up at some point, could this be the staging db?

AlmightyYakob:
> If we go for a full staging set up at some point, could this be the staging db?

Depends on what we define as staging. The database specified here will be running tests on it whenever there's a PR, so we'd have to specify if we want to mix the two or not.

waxlamp:
> > So the problem here is the technical debt we're about to introduce: developers now have to know the _effect_ of consulting the workspace mapping, which means our abstraction is leaking in a bad way. Renaming `workspace_mapping` to `uncached_workspace_mapping` would be good in that it would signal to devs that there's some choice to be made here, but it doesn't protect them from making the wrong call and breaking correctness.
> > It feels like this issue of asking for a workspace by name vs. creating a new entry or changing an existing entry is starting to become larger than can just be handled by the naive use of the lru cache decorator. I see two ways out of this:
> > 
> > 1. The simpler way is to use the fix I suggested on Slack yesterday: to simply invalidate the cache whenever a renaming, deletion, or creation occurs, and let the cache handle the common case of lookups when the state isn't changing. This would be a bad option if `workspace_exists()` is used more often than I think it is.
> 
> Currently, this is what's being done. On calls to creating, renaming or deleting a workspace, the cache is invalidated. I'm not sure what you mean referring to `workspace_exists` being possibly called often? That function won't affect the cache, because it's not using the cached version of `workspace_mapping`.
> 

I thought the bug we had was caused by checking for workspace W's existence leading to `None` being cached as the (wrong) mapping value for W? What I meant about the frequency of `workspace_exists` was that, under this scheme, if we are calling that a lot we'd be invalidating the cache a lot, and losing out on the performance benefits.

> > 1. The more complex, but probably _correct_ way is to abstract these operations into a module, make the workspace mapping collection into a private variable, and export public functions to do the actual operations (create, rename, delete, and check for existence). The caching mechanism would probably become more explicit, like a dict we maintain in that same module between invocations of the public functions. Perhaps there's an actual LRU cache implementing object in the standard library we could use, but access at a finer granularity than the `@lru_cache` decorator does.
> 
> I'm kind of weary to try to implement our own caching, as now we'd have to worry about general caching bugs, as well as bugs in the caching mechanism itself. Can you explain why you think we should stray away from `lru_cache`?
> 

The reason is that the simple caching strategy isn't working, and is leading us to explicitly create cached and uncached paths (see more about this below). The alternative is to refine the caching strategy in some manner. I agree that I don't want to implement function memoization myself, but we're stuck between a rock and a hard place as things stand.

> I do think we should be explicit about what's being cached, and where. For example, `cached_workspace_mapping` vs `workspace_mapping`. I think separating the two, and having `cached_` functions that just cache calls to the original function is a good approach, because it makes it clear that we're using caching, and allows us to use the original functions if we need to. If we were to do this, I'd have to (and should) apply the same change to `get_workspace_db`. Let me know what you think about this approach.
> 

But this is what I'm trying to avoid. What I want is a solution where no one ever needs to know which workspace mapping to call.

If what you're saying is that once we fix this, future developers (including future versions of ourselves) won't have to worry about this anymore, then that's one thing, but I can't imagine that's what the situation will be. For instance, at some point we might want to create "symlinks", at which point we'll want to add a mapping entry that maps a name to an existing name (or something like that; just making up the example off the top of my head). We'll be back to thinking hard about whether to use the cached or uncached variant, and we'll likely get it wrong like we did in this instance.

Ultimately what I'm saying is, we should be able to design the workspace mapping functionality in such a way that all this cached/uncached complexity is hidden away inside the module, and we devs can happily just call functions like `add_mapping()`, `remove_mapping()`, `workspace_exists()`, `get_workspace()`, etc., without every thinking about caching unless we're implementing new functionality within that module.

I would be fine with deferring such a thing to a new PR, but I wouldn't be fine with delaying that work for very long.

It is also possible that I'm misunderstanding the problem here, but this is just setting off my spider sense.

> > The other issue here is testing--we need to figure out how to run those tests in CI, because a feature like this really will require us to make sure it's working long term.
> 
> I agree. If we were to just specify `ARANGO_HOST`, `ARANGO_PROTOCOL` and `ARANGO_PASSWORD` in the environment where the tests run, they'd succeed (assuming we actually point to a valid arango instance). As far as what that should be, I'm not sure, but I think we can agree it should be separate from our production db.

Don't we currently have some tests that are using the live db for testing? I agree that that's not ideal but if it's what we're doing now we should continue, and we should also file an issue to figure out a better staging/testing scenario.

waxlamp:
Had a chat with @AlmightyYakob. He convinced me that this approach is largely correct. We identified two TODOs:
- [ ] instead of a cached and uncached variant, we can simply have the cached function, and refer to its [`__wrapped__` attribute](https://docs.python.org/3/library/functools.html#functools.lru_cache) to access the underlying (uncached) function. This will resolve my fears about confusing future developers.
- [ ] since we don't have a good solution for CI testing with a live database, we'll mark those tests as skipped for now so that CI passes and the automated deployments succeed. @AlmightyYakob has run those tests locally, so we have some confidence that the code is correct. At a future time, we will figure out a robust live testing strategy and unskip those tests.

multinet/db.py
```diff
@@ -129,6 +130,9 @@ def create_workspace(name: str) -> None:
         # Could only happen if there's a name collisison
         raise InternalServerError()
 
+    workspace_mapping.cache_clear()
+    get_workspace_db.cache_clear()
+
```
JackWilb, line 135: This is the only instance of these lines without a comment

test/test_workspace.py
```diff
@@ -0,0 +1,131 @@
...
+def handled_workspace(generated_workspace):
+    """
+    Create a workspace, and yield the name of the workspace.
+
+    On teardown, deletes the workspace.
+    """
+    yield generated_workspace
+
+    delete_workspace(generated_workspace)
+    workspace_mapping.cache_clear()
```
JackWilb, line 34: The `workspace_mapping` cache is invalidated by the delete_workspace function so you might not need to clear the cache manually

AlmightyYakob: Good point

multinet/db.py
AlmightyYakob: Good catch, I'll add it

PR #374, opened 2020-04-30
OAuth Fixes

AlmightyYakob:
* Adds a default `return_url` for the login endpoint
* Replaces the use of `states_to_return_url` (which is __not__ safe across multiple workers) with the use of the `session` object (which is). 
* Make the `hd` field of `GoogleUserInfo` optional, as only some accounts have that field.

### DIFF
--- a/multinet/auth/google.py
+++ b/multinet/auth/google.py
@@ -26,7 +26,7 @@ from multinet.user import (
 from multinet.auth import MULTINET_COOKIE
 from multinet.auth.types import GoogleUserInfo, User
 
-from typing import Dict
+from typing import Dict, Optional
 
 
 CLIENT_ID = os.getenv("GOOGLE_CLIENT_ID")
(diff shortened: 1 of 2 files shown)


multinet/auth/google.py
```diff
@@ -91,10 +98,13 @@ def init_oauth(app: Flask) -> None:
 @bp.route("/login")
 @use_kwargs({"return_url": fields.Str(location="query")})
 @swag_from("swagger/google/login.yaml")
-def login(return_url: str) -> ResponseWrapper:
+def login(return_url: str = None) -> ResponseWrapper:
```
waxlamp, line 101: I think the type of `return_url` needs to be updated to `Optional[str]`.

AlmightyYakob: Good catch, not really sure why mypy doesn't catch that :thinking:. Might be due to the decorators.

AlmightyYakob:
@JackWilb @waxlamp despite you both approving, neither of you actually "approved".

waxlamp:
> @JackWilb @waxlamp despite you both approving, neither of you actually "approved".

Pretty sure I had approved it, but the commit push dismissed my review.

PR #375, opened 2020-05-01
Make Arango root password configurable at provisioning time

### LINKED ISSUE #348: Make the ArangoDB playbook password configurable
The default password [here](https://github.com/multinet-app/multinet-server/blob/master/devops/arangodb.yml#L7) should be configurable via environment variable, etc.

waxlamp:
Closes #348

I am not planning to actually use this playbook to update the ArangoDB password now; I'll do that manually. But this includes the appropriate fix in our automated deployment script.

### DIFF
--- a/devops/arangodb.yml
+++ b/devops/arangodb.yml
@@ -4,7 +4,10 @@
 
   vars:
     ansible_python_interpreter: /usr/bin/python3
-    arangodb_password: "letmein"
+
+  vars_prompt:
+    - name: arangodb_root_password
+      prompt: "Password for ArangoDB root account"
 
   tasks:
     # https://www.arangodb.com/download-major/ubuntu/
@@ -36,7 +39,7 @@
       become: true
 
     - name: Overwrite the default arangodb password
-      shell: ARANGODB_DEFAULT_ROOT_PASSWORD={{ arangodb_password }} arango-secure-installation
+      shell: ARANGODB_DEFAULT_ROOT_PASSWORD={{ arangodb_root_password }} arango-secure-installation
       become: true
 
     - name: Enable arangodb to listen on all interfaces
--- a/devops/README.md
+++ b/devops/README.md
@@ -3,13 +3,15 @@
 The `arangodb.yml` Ansible playbook in this directory installs and configures
 ArangoDB on an Ubuntu server. It does the following:
 - installs ArangoDB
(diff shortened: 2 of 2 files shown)


devops/arangodb.yml
```diff
@@ -4,7 +4,10 @@
 
   vars:
     ansible_python_interpreter: /usr/bin/python3
-    arangodb_password: "letmein"
+
+  vars_prompt:
+    - name: arangodb_root_password
+      prompt: "Password for ArangoDB root account"
```
JackWilb, line 11: We could also supply an override-able default if you think that might be useful?
 
```suggestion
      prompt: "Password for ArangoDB root account"
      default: "letmein"
```

AlmightyYakob: Also, having this as a prompt means we'd never be running this in an automated way, right?

waxlamp: @JackWilb: No default--this forces the devops engineer to select and keep track of a password.

@AlmightyYakob: that's true, but you can avoid the prompt by using the `--extra-vars` option on the command line. If the README updates are not clear enough about that, let me know and I'll fix it.

waxlamp:
> Also, have you managed to test this anywhere?

I have not. I'm not super worried about it because we can solve problems on the fly if and when we do use this to provision a fresh instance of our database. It's not something I expect to happen soon, and in my experience trying to fix Ansible issues proactively is usually pretty inefficient for whatever reason.

>Looks like all our vagrant related files were removed and it's not easy to test without those. Should I look to add them back?

If you want to add a Vagrant setup that can replicate our AWS target, that could be helpful, but don't spend more than an hour or two on it (and do it in a separate PR). I can see that it would be useful for capturing our SSL setup (and also for testing the addition in this PR).

waxlamp:
> > Also, have you managed to test this anywhere?
> 
> I have not. I'm not super worried about it because we can solve problems on the fly if and when we do use this to provision a fresh instance of our database. It's not something I expect to happen soon, and in my experience trying to fix Ansible issues proactively is usually pretty inefficient for whatever reason.
> 

Actually... I did sort of test it. I set up the changes you see here, then added a `fail` task at the head of the task list that displayed the password value, then I (perhaps somewhat recklessly) ran the playbook command against multinet.app. That showed me that both the prompted password input, and the `--extra-vars` switch were working to push a password into the environment.

PR #377, opened 2020-05-02
All ansible steps for ssl and add vagrant back for testing

### LINKED ISSUE #376: Update Ansible playbook to install/run ArangoDB with SSL

JackWilb:
Closes #376.

TODO:
- [x] Fix the chmod command by using the file command? (currently gives a warning)
The `file` and `setfacl` command combination seems to set the permissions to 640 but after adding the `vagrant` user to the group, `root`, server.pem is not readable. Must be an artifact from the `setfacl` command
```
-rw-------  1 root   root   1927 May  2 20:08 cert.pem
-rw-------  1 root   root   1647 May  2 20:08 chain.pem
-rw-------  1 root   root   3574 May  2 20:08 fullchain.pem
-rw-------  1 root   root   1704 May  2 20:08 privkey.pem
-rw-r-----+ 1 root   root   5278 May  2 20:08 server.pem
```
- [x] Figure out how to use arangodb_root_password without a default
- [ ] Determine if there is a way for us to test the server without regular tcp (since the ssl cert doesn't match localhost)
- [x] Set up vagrant with equivalent settings to our prod server


devops/Vagrantfile
```diff
@@ -0,0 +1,21 @@
...
+    v.name = "multinet-ubuntu-arango-test"
+    v.memory = 4096
+    v.cpus = 2
+  end
+
+  config.vm.box = "ubuntu/bionic64"
+  config.ssh.insert_key = false
+
+  config.vm.provision "ansible" do |ansible|
+    ansible.verbose = "v"
```
waxlamp, line 15: We should probably omit this line (and let the user decide on their own level of verbosity).

```suggestion
```

devops/Vagrantfile
```diff
@@ -0,0 +1,21 @@
...
+    ansible.playbook = "arangodb.yml"
+    ansible.raw_arguments  = [
+      "-e arangodb_root_password=letmeinvfile"
```
waxlamp, line 18: How does the Vagrant user supply a cert file?

devops/arangodb.yml
```diff
@@ -1,5 +1,6 @@
 ---
 - hosts: all
+  become: yes
```
waxlamp, line 3: This should not go here. You can add `become: yes` to any individual task that requires root access instead for a safer provisioning setup.

devops/arangodb.yml
```diff
@@ -9,6 +10,10 @@
     - name: arangodb_root_password
       prompt: "Password for ArangoDB root account"
 
+    - name: ssl_cert_path
+      prompt: "Path to parent ssl cert directory"
```
waxlamp, line 14: Maybe this answers the question above about how the user gets the SSL cert into the Vagrant setup.

I don't think this needs to be a prompt though. The password is prompted for in order to provide an option for not echoing the password itself to the screen. This option can just be passed in via command line argument.

devops/arangodb.yml
```diff
@@ -9,6 +10,10 @@
...
+      default: "/etc/letsencrypt/live/multinet.app/"
```
waxlamp, line 15: Why is this path the default?

devops/arangodb.yml
waxlamp, line 14:
```suggestion
      prompt: "Path to parent SSL cert directory"
```

devops/arangodb.yml
```diff
@@ -42,11 +47,39 @@
       shell: ARANGODB_DEFAULT_ROOT_PASSWORD={{ arangodb_root_password }} arango-secure-installation
       become: true
 
+    - name: Copy the compressed ssl certificate files to the server
```
waxlamp, line 50:
```suggestion
    - name: Copy the compressed SSL certificate files to the server
```

devops/arangodb.yml
```diff
@@ -42,11 +47,39 @@
...
+        src: '{{ ssl_cert_path }}'
+        dest: /home/ubuntu/
+        follow: yes
+        mode: u=rw,g=,o=
+
+    - name: Combine the chain and the private key into one file (required by arango)
+      shell:
+        chdir: /home/ubuntu/
+        cmd: cat fullchain.pem privkey.pem > server.pem
+
```
waxlamp, line 61: The command here _is_ idempotent, which is good, but in general when using the `shell` directive, you should put in logic to ensure this task doesn't run if it has already run before. In this case, you can add the property that says this task creates a file (maybe it's `creates`; I can't remember off the top of my head).

devops/arangodb.yml
```diff
@@ -42,11 +47,39 @@
...
+    - name: Set the combined file as 600 permission
+      file:
+        path: /home/ubuntu/server.pem
+        mode: u=rw,g=,o=
+
+    - name: Add an exception to the permissions for the arangodb user
+      shell:
+        chdir: /home/ubuntu/
+        cmd: setfacl -m u:arangodb:r /home/ubuntu/server.pem
```
waxlamp, line 70: Why do we need an exception for this user? Where does the `setfacl` command come from?

devops/arangodb.yml
```diff
@@ -42,11 +47,39 @@
...
+
+    - name: Add the ssl filepath to the config
```
waxlamp, line 72:
```suggestion
    - name: Add the SSL filepath to the config
```

devops/arangodb.yml
```diff
@@ -42,11 +47,39 @@
...
+      lineinfile:
+        path: /etc/arangodb3/arangod.conf
+        regexp: '^\[ssl\]'
+        line: "\n[ssl]\nkeyfile = /home/ubuntu/server.pem"
```
waxlamp, line 76: I'm pretty sure this won't be idempotent--I imagine that every time the playbook runs, this replacement will occur, resulting in multiple "[ssl]" blocks with repeated "keyfile" lines.

Can you confirm this?

devops/arangodb.yml
JackWilb: Just for speed when we're reprovisioning. Since this filepath will always be the same for us, I figured we could save a little time

devops/arangodb.yml
JackWilb: We need an exception for this user so that arango can read the key file. The other option would be to set arangodb as the owner of the file but I'm not sure that's best. The attack vector is that someone has access to the arangodb user and could modify the SSL cert files and setting arangodb as the owner would allow an attacker to change the permissions on the file to `+w`.

The `setfacl` command is just a linux utility for setting file permissions in a more granular way using ACLs (access control lists). I've seen CHPC used it in production here at the U in the protected environment, so I'm fairly certain that this is decent way to do it. I guess the other option is to add a new group, add arangodb to that group, and then make that group the owner of the file. I think ACLs are a faster and cleaner way of managing the permission

devops/README.md
```diff
@@ -4,14 +4,35 @@ The `arangodb.yml` Ansible playbook in this directory installs and configures
...
 
 To set the root account password, you can either supply it on the command line
 (see below) or just let Ansible prompt you for it.
 
-To run it via ssh, run a command such as the following:
+To use a wildcard ssl certificate for your domain you'll need to supply one, we use Let's Encrypt, and pass in the file path to the command.
+If you're using a provider other than Let's Encrypt, you'll have to modify the playbook to work with your provider and file structure. 
+If you're using Let's Encrypt, follow along with [this](https://medium.com/@saurabh6790/generate-wildcard-ssl-certificate-using-lets-encrypt-certbot-273e432794d7) 
+guide to generate a wildcard certificate. This playbook assumes you have the certificate files and that they're in the default location.
+If you don't want a ssl certificate, you'll have to modify the playbook and remove the `ssl_cert_path` env var from the command below.
```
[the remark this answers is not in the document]
JackWilb: Sure, but that will take a more time (that I'd rather devote to building the vis apps) and our use case will generally be deploying a server with a cert. If we ever don't want that, it's easy to comment out the lines that handle the certificates and change `ssl://` to `tcp://`

devops/README.md
```diff
@@ -4,14 +4,35 @@ The `arangodb.yml` Ansible playbook in this directory installs and configures
...
+
+To run the ansible playbook via ssh, run a command such as the following:
+
+```
+sudo ansible-playbook arangodb.yml -i <target-hostname>, --ssh-extra-args="-i <identity.pem>" -e arangodb_root_password=<password> ssl_cert_path=<path-to-files>
+```
+
+*NOTE*: You have to use sudo so that you may copy the local, write protected certificate files.
```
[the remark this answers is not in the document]
JackWilb: That's true. The problem is that the private key file is only readable to root after running certbot. I was concerned that telling people to change the permissions of the files on their machines would be less secure than just telling them to run ansible as root. I could also add some documentation that changes the private key file permission to readable and back to unreadable if that makes more sense?

devops/arangodb.yml
JackWilb: After running it twice, we're left with this:

```
[ssl]
keyfile = /home/ubuntu/server.pem
keyfile = /home/ubuntu/server.pem
```

I'll update the regex

devops/.gitignore
```diff
@@ -0,0 +1,2 @@
+.vagrant/
+ubuntu-bionic-18.04-cloudimg-console.log
```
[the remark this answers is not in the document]
JackWilb: Fixed in 736e3f3

devops/arangodb.yml
JackWilb: Vagrant doesn't prompt for these vars, which is why I had to specify the password in the Vagrantfile. It does take the defaults if they exist though. If a user wanted to not use the default path, they would have to add another argument to `raw_arguments` to the Vagrantfile

devops/arangodb.yml
JackWilb: It also helps us with the vagrant set up, since vagrant doesn't prompt for these. It will take the default as the argument

devops/arangodb.yml
JackWilb: I had to switch to `blockinfile`. Fixed in cfed808

devops/arangodb.yml
JackWilb: Fixed in cfed808

devops/README.md
```diff
@@ -4,14 +4,35 @@ The `arangodb.yml` Ansible playbook in this directory installs and configures
...
+
+# Vagrant
+
+To test locally, you can use vagrant. Vagrant is a python package that is most easily installed with pip by running, 
+`pip install vagrant`. Once you have Vagrant installed, you can run the vagrant file using:
```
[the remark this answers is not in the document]
JackWilb: Fixed in cfed808

devops/arangodb.yml
JackWilb: Added creates in c3f8ef2

devops/README.md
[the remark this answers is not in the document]
JackWilb: If you think there will be use cases where we'll want to use this without a SSL cert, I can add the conditionals. For people installing it on their own machines for testing they should use the `docker-compose` method, and if they're deploying to a webserver they should probably be using SSL, right?

devops/arangodb.yml
JackWilb: Do you think it makes sense to have it use creates? What if we reprovision on a server to update the keyfile? Then the server.pem file wouldn't be updated. I guess the question is, do you think we'll use this file to update the certificate?

devops/README.md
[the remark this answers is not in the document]
waxlamp: Sounds good.

devops/arangodb.yml
waxlamp: Ideally we'd want to make the task depend on those two files as dependencies somehow. Probably not worth figuring this out right now for the reasons you mentioned in the other thread. In this case, like I said originally, you don't have to have the `creates` (since the concatentation is idempotent).

devops/arangodb.yml
waxlamp: Yeah this should be ok. If you have any reading you've done on ACLs, it would be interesting to see those. I don't know much about them, so thanks for bringing them in here.

devops/README.md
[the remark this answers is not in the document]
waxlamp: > That's true. The problem is that the private key file is only readable to root after running certbot. I was concerned that telling people to change the permissions of the files on their machines would be less secure than just telling them to run ansible as root. I could also add some documentation that changes the private key file permission to readable and back to unreadable if that makes more sense?

People should only run as root when it's required, and because you can use sudo to change permissions momentarily on a single file, you should do that instead of running the entire provisioning tool as root. I think you can just explain that changing permissions on the file may be needed to make it readable for `ansible-playbook` to copy it, and it's not even necessary to explain step-by-step how to do it. We need to assume some level of familiarity with Linux, or else our docs will grow exponentially.

But we definitely should *not* instruct people to run `ansible-playbook` as root. I don't know everything it does, and for that reason running it as root is dangerous.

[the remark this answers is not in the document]
waxlamp: > If you think there will be use cases where we'll want to use this without a SSL cert, I can add the conditionals. For people installing it on their own machines for testing they should use the `docker-compose` method, and if they're deploying to a webserver they should probably be using SSL, right?

Well, the main use case right now is for us to test the playbook... with an SSL cert. So you're right, there's no need to do more work now to make it work for non-SSL cert use cases.

devops/arangodb.yml
waxlamp: Ok, makes sense if that's where the LE utility places the file when run locally.

devops/arangodb.yml
JackWilb: I removed the `creates` because it will complicate things when we run this playbook multiple times on the same provider

devops/README.md
```diff
@@ -4,14 +4,35 @@ The `arangodb.yml` Ansible playbook in this directory installs and configures
...
 
 ```
-ansible-playbook arangodb.yml -i <target-hostname>, --ssh-extra-args="-i <identity.pem>" -e arangodb_root_password=<password>
+sudo vagrant up -i
 ```
+
+*NOTE*: Since vagrant is started with sudo, any subsequent command to interact with the instance must be ran with sudo,
+e.g. `sudo vagrant ssh`.
```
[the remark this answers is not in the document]
JackWilb: Deleted in b594996

devops/README.md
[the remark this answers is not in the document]
JackWilb: Added a note about local file permissions in the README. Removed `sudo` from the commands in b594996

devops/Vagrantfile
JackWilb: The current way is just through the defaults in the ansible playbook

devops/arangodb.yml
```diff
@@ -50,15 +50,14 @@
       copy:
         src: '{{ ssl_cert_path }}'
         dest: /home/ubuntu/
-        follow: yes
+        local_follow: yes
```
waxlamp, line 53: What does this do?

JackWilb: Follows local symlinks. The files in `/etc/letsencrypt/live/multinet.app` are symlinked to `/etc/letsencrypt/archive/multinet.app`. The link names in the live folder never change, but I think the link targets do in the archive folder. This just lets us get the live version of the files in the archive

AlmightyYakob:
This seems to look good, I haven't tested it locally though.

> * [ ]  Determine if there is a way for us to test the server without regular tcp (since the ssl cert doesn't match localhost)

Does this currently prevent testing locally?

waxlamp:
> This seems to look good, I haven't tested it locally though.
> 
> > * [ ]   Determine if there is a way for us to test the server without regular tcp (since the ssl cert doesn't match localhost)
> 
> Does this currently prevent testing locally?

Good question. I have a task on my list to upload a GPG-encrypted tarball containing the cert files so that people with the password could indeed test with SSL, and I think the current proposal for testing *without* SSL is to modify the playbook to omit the SSL steps.

I'd prefer to make this possible without modifying the file, but as Jack points out, it may not be worth much of our time to make that possible when the major users are us, and the major use case is with SSL.

JackWilb:
> Does this currently prevent testing locally?

It's possible to test it locally, insofar as you can spin up the server and verify it's running on ssl and using the cert key file. However, is not possible to test locally as in querying from the database, since any local connection will not be using the correct domain name (i.e. it will be localhost, not multinet.app). 

Maybe we should wait to merge this until we reprovision the old server and test that there are no bugs the script?

AlmightyYakob:
> Maybe we should wait to merge this until we reprovision the old server and test that there are no bugs the script?

That sounds like a good idea to me.

PR #381, opened 2020-05-11
Add /api prefix to user endpoints

### LINKED ISSUE #378: Add api prefix to /user and /user/oauth/google blueprint prefixes

AlmightyYakob:
Fixes #378.

Currently, the Google Developer Console has both the new and old endpoints as valid redirect URIs. Once this is merged, we can remove the old ones.

### DIFF
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -42,8 +42,8 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     app.register_blueprint(downloaders.csv.bp, url_prefix="/api")
     app.register_blueprint(downloaders.d3_json.bp, url_prefix="/api")
 
-    app.register_blueprint(auth.bp, url_prefix="/user")
-    app.register_blueprint(google.bp, url_prefix="/user/oauth/google")
+    app.register_blueprint(auth.bp, url_prefix="/api/user")
+    app.register_blueprint(google.bp, url_prefix="/api/user/oauth/google")
 
     google.init_oauth(app)
     register_legacy_workspaces()


waxlamp:
> Currently, the Google Developer Console has both the new and old endpoints as valid redirect URIs. Once this is merged, we can remove the old ones.

Can you elaborate on this? Do you mean it will accept auth calls from both, i.e. that production and the deploy preview for this branch both work?

AlmightyYakob:
> > Currently, the Google Developer Console has both the new and old endpoints as valid redirect URIs. Once this is merged, we can remove the old ones.
> 
> Can you elaborate on this? Do you mean it will accept auth calls from both, i.e. that production and the deploy preview for this branch both work?

Since the production instance still uses the old non-prefixed paths, the google console allows both, but once this is merged, only the prefixed paths will exist, so I can remove the old ones.

AlmightyYakob:
> This LGTM, but we'll need to update the client before we merge this right? (Otherwise, production will break?)

I don't think the production client will actually be affected, because https://github.com/multinet-app/multinet-client/pull/80 isn't merged yet? I don't think these endpoints are used anywhere else yet.

waxlamp:
> > This LGTM, but we'll need to update the client before we merge this right? (Otherwise, production will break?)
> 
> I don't think the production client will actually be affected, because [multinet-app/multinet-client#80](https://github.com/multinet-app/multinet-client/pull/80) isn't merged yet? I don't think these endpoints are used anywhere else yet.

Ah right. Ok, then this should be good to merge, and I'll just make sure to update the code in multinet-app/multinet-client#80. Thanks!

PR #386, opened 2020-05-26
Remove session cookie on logout

waxlamp:
Depends on #380.

### DIFF
--- a/multinet/auth/__init__.py
+++ b/multinet/auth/__init__.py
@@ -36,14 +36,12 @@ def user_info() -> ResponseWrapper:
 def logout() -> ResponseWrapper:
     """Return the filtered user object."""
 
-    # Load the user model and invalidate its session.
-    cookie = session.get(MULTINET_COOKIE)
+    # Instruct the browser to delete its session cookie, if it exists.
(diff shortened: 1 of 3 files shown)


multinet/auth/__init__.py
```diff
@@ -36,5 +36,14 @@ def user_info() -> ResponseWrapper:
...
 
+    # Load the user model and invalidate its session.
+    cookie = session.get(MULTINET_COOKIE)
+    if cookie is not None:
+        user = load_user_from_cookie(cookie)
+        if user is not None:
+            delete_user_cookie(user)
+
+    # Instruct the browser to delete its session cookie.
     session.pop(MULTINET_COOKIE, None)
```
AlmightyYakob, line 47: We could use just one call to `session` here:

```suggestion
    # Instruct the browser to delete its session cookie, if it exists.
    cookie = session.pop(MULTINET_COOKIE, None)
    if cookie is not None:
        # Load the user model and invalidate its session.
        user = load_user_from_cookie(cookie)
        if user is not None:
            delete_user_cookie(user)
```

PR #395, opened 2020-06-03
Fix Swagger UI API failures

### LINKED ISSUE #382: substitute {graph} for actual name in nested-json and newick uploads in swagger.
When testing the uploader API endpoints (for uploading newick and nested-json) through the swagger interface, the graph name is not substituted. For example this is the curl that is attempted:

curl -X POST "https://api.multinet.app/api/nested_json/Curt/{graph}" -H "accept: application/json" -H "Content-Type: text/plain" -d "{\"edge_fields\": [\"weight\"], \"node_fields\": [\"node name\", \"node weight\"], \"node_data\": {...***content deleted for brevity***...}"

NOTE:  Please see how the graph name wasn't correctly substituted for {graph}.   

This happens on the nested-json upload and the newick upload.  I didn't test other upload methods yet.    What I mean is that I entered a graph name as the required parameter and pasted in the file content but received a run time error when executing this from the swagger interface.  When I copied the text to my terminal, and substituted the graph name in the curl call, the call was successful.

waxlamp:
We had the wrong datatype listed in the UI for the Newick, nested JSON, and D3 JSON uploaders. 

I also discovered the example data for the nested JSON uploader was invalid JSON; that is fixed on this branch as well.

Thanks to @curtislisle for bringing this to our attention!

Closes #382

### DIFF
--- a/multinet/uploaders/swagger/d3_json.yaml
+++ b/multinet/uploaders/swagger/d3_json.yaml
@@ -5,7 +5,7 @@ consumes:
 
 parameters:
   - $ref: "#/parameters/workspace"
-  - $ref: "#/parameters/table"
+  - $ref: "#/parameters/graph"
   - name: data
     in: body
     description: Raw JSON text
--- a/multinet/uploaders/swagger/nested_json.yaml
+++ b/multinet/uploaders/swagger/nested_json.yaml
@@ -14,35 +14,36 @@ parameters:
       example: |-
         {
           "children": [
-            "children": [
-              {
-                "edge_data": {
-                  "weight": 12.28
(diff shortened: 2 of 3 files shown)


waxlamp:
@AlmightyYakob, can you do whatever magic is necessary to make this show up as a Heroku review app? Then, if you ping @curtislisle when it's ready, he can carry out the same tests from before to ensure that the problem is fixed.

waxlamp:
@curtislisle, you should be able to test your procedure here https://multinet-pr-395.herokuapp.com/ to see if this branch fixes it. Let me know what happens.

Issue #347, opened 2020-03-18
Possible race condition when retrieving tables after deletion


AlmightyYakob:
I discovered this when adding things to the client side, where one of the operations is to delete a table, and then immediately query for node tables. I've gotten two distinct errors from this, but they seem to stem from the same general premise.

Couple of stack traces for reference (I've annotated where the stack trace goes from our code to arango) :
```
Debugging middleware caught exception in streamed response at a point where response headers were already sent.
Traceback (most recent call last):
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/sentry_sdk/integrations/wsgi.py", line 251, in __iter__
    reraise(*_capture_exception(self._hub))
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/sentry_sdk/_compat.py", line 57, in reraise
    raise value
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/sentry_sdk/integrations/wsgi.py", line 247, in __iter__
    chunk = next(iterator)
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/werkzeug/wsgi.py", line 506, in __next__
    return self._next()
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/werkzeug/wrappers/base_response.py", line 45, in _iter_encoded
    for item in iterable:
  File "/home/jake/projects/multinet-server/multinet/util.py", line 69, in generate
    for row in iterator:
  File "/home/jake/projects/multinet-server/multinet/db.py", line 228, in <genexpr>
    return (table[0] for table in tables if desired_type(table))
  File "/home/jake/projects/multinet-server/multinet/db.py", line 207, in <genexpr>
    if not table["name"].startswith("_")
  File "/home/jake/projects/multinet-server/multinet/db.py", line 362, in table_fields
    if space.has_collection(table) and space.collection(table).count() > 0:
--------------------------------------------------------------------------------------------
  File "/home/jake/.local/share/virtualenvs/multinet-server-he5vy4Rf/lib/python3.7/site-packages/arango/collection.py", line 554, in count
    return self._execute(request, response_handler)
[...]

PR #396, opened 2020-06-15
Filter workspaces based on user permissions

waxlamp:
This PR implements one permissions-based filter for a part of the API. The `GET /workspaces` endpoint now filters the returned workspaces based on the logged in user and properties of the workspaces themselves.

This PR is also a prototype of sorts: what I've done here works but might require some refactoring after we start to fill in more permission-based changes for the other endpoints.

The workspaces endpoint is special in that it's not all or nothing: the full list of workspaces has to come from the database, and then we evaluate what to send as the response. I suspect other endpoints will have a simpler gating mechanism which results in a 200 or a 40x error depending on the user.

This PR adds a new semantics of workspace creation depending on who (if anyone) is logged in: a logged-in user creates a non-public workspace owned by them; a logged-out user creates a public workspace owned by nobody. Only an admin would be able to delete such workspaces, so this is likely the wrong thing to do. What would be better? I suspect it would simply be to disallow logged-out users from creating workspaces.

TODO:
- [x] write a script to upgrade the workspace mapping collection to have the right schema (with proper filled in data)
- [x] update the workspace creation endpoint to fill in correct metadata for the new workspace
- [x] ~use `Optional[User]` as the type for the `owner` field of `Workspace`~
- [x] prevent non-logged-in user from creating workspaces
- [x] ~add tests~
- [x] perform DB migration (before merge)


multinet/api.py
```diff
@@ -112,7 +138,29 @@ def get_node_edges(
...
+            "public": False,
+        },
+    }
+
+    # Assign owner and public flags depending on who's logged in.
+    user = session_user()
+    if user is None:
+        new_doc["permissions"]["public"] = True
+    else:
+        new_doc["permissions"]["owner"] = user.sub
```
JackWilb, line 160: This means that users that are not logged in can make workspaces. Is this what we resolved to do? Also, @waxlamp, you mentioned that you could make the owner an optional string, instead of just blank, which sounds like a good idea

waxlamp: > This means that users that are not logged in can make workspaces. Is this what we resolved to do?

This is exactly what we need to discuss; see the PR description.

> Also, @waxlamp, you mentioned that you could make the owner an optional string, instead of just blank, which sounds like a good idea

That does make more sense from a typing standpoint, but from an execution standpoint, there's a nice advantage to using a blank string for "no owner": no real user has a blank string as their `sub` value, so it leaves all the code for processing users nice and simple. OTOH, the type safety is pretty attractive. I will probably make this change but if there are opposing viewpoints, now would be a good time to hear them.

JackWilb: > > This means that users that are not logged in can make workspaces. Is this what we resolved to do?
> 
> This is exactly what we need to discuss; see the PR description.

Understood. I think we shouldn't allow users that are not logged in to create workspaces. This could lead to abuse, eventually, and it makes sense that every workspace should have an owner. A workspace not having an owner could lead to confusion, where users make workspaces when they're not logged in, then they become public and they're unable to edit them.

I welcome any counter arguments though

> > Also, @waxlamp, you mentioned that you could make the owner an optional string, instead of just blank, which sounds like a good idea
> 
> That does make more sense from a typing standpoint, but from an execution standpoint, there's a nice advantage to using a blank string for "no owner": no real user has a blank string as their `sub` value, so it leaves all the code for processing users nice and simple. OTOH, the type safety is pretty attractive. I will probably make this change but if there are opposing viewpoints, now would be a good time to hear them.

I think we should make the change for better typing. I think making it an optional string will mean that we're less like to forget about the edge case where a workspace has no owner.

waxlamp: > > > This means that users that are not logged in can make workspaces. Is this what we resolved to do?
> > 
> > 
> > This is exactly what we need to discuss; see the PR description.
> 
> Understood. I think we shouldn't allow users that are not logged in to create workspaces. This could lead to abuse, eventually, and it makes sense that every workspace should have an owner. A workspace not having an owner could lead to confusion, where users make workspaces when they're not logged in, then they become public and they're unable to edit them.
> 
> I welcome any counter arguments though
> 

I agree that we should just not allow it for unlogged-in users. @AlmightyYakob, any objection?

I will add a commit to handle this case; it'll be a good prototype use case for a `401 Unauthorized` response.

> > > Also, @waxlamp, you mentioned that you could make the owner an optional string, instead of just blank, which sounds like a good idea
> > 
> > 
> > That does make more sense from a typing standpoint, but from an execution standpoint, there's a nice advantage to using a blank string for "no owner": no real user has a blank string as their `sub` value, so it leaves all the code for processing users nice and simple. OTOH, the type safety is pretty attractive. I will probably make this change but if there are opposing viewpoints, now would be a good time to hear them.
> 
> I think we should make the change for better typing. I think making it an optional string will mean that we're less like to forget about the edge case where a workspace has no owner.

If we're not allowing ownerless workspaces, then the `owner` field *should* be `str` rather than `Optional[str]` no?

AlmightyYakob: > I agree that we should just not allow it for unlogged-in users. @AlmightyYakob, any objection?

@waxlamp I agree, only logged in user should be able to create a workspace. 

> If we're not allowing ownerless workspaces, then the `owner` field _should_ be `str` rather than `Optional[str]` no?

I think so, I don't see any particular benefit to making it optional. However, this means that we'll have to do some migration on existing workspaces. Maybe we could set them to some default user that we create (and have access to), and then we can assign them to whoever they belong to.

JackWilb: I agree with Jake on that last point

waxlamp: > Maybe we could set them to some default user that we create (and have access to), and then we can assign them to whoever they belong to.

Since we don't have an in-app admin console yet, I don't see the value in assigning these to an existing user, as opposed to just leaving the string blank, and setting the workspaces to be public. I wrote a migration script (2d1491c143d10d10b0b6a07bb26209f6590fc36d in this PR) to do just that. Once the actual owners of these workspaces have accounts, we can re-assign them manually.

AlmightyYakob:
It would also be good to have tests for this new behavior, as it adds a decent amount of complexity and it'll be good to keep these things straight.

multinet/api.py
waxlamp: Added a 401 error for logged-out users trying to create a workspace: see dbabdbc.

waxlamp:
This is now ready for final review.

JackWilb:
I'm having an issue. When I'm not logged in, the `/api/workspaces` endpoint is returning me invalid JSON, instead of an empty list. It returns:

```
[
```

I'm also struggling to sign in using the api endpoints on the preview app. It's giving me an error about the redirect URI

AlmightyYakob:
> I'm having an issue. When I'm not logged in, the `/api/workspaces` endpoint is returning me invalid JSON, instead of an empty list. It returns:
> 
> ```
> [
> ```


This sounds like another instance of #347. Did you encounter this error when testing the app locally? Or were you using the preview app?

> 
> I'm also struggling to sign in using the api endpoints on the preview app. It's giving me an error about the redirect URI

This is probably due to us having to explicitly register the redirect URLs on Google's side for OAuth. I'm guessing @waxlamp didn't add the preview app to the allowed list in the dev console. We don't have a great process for this yet.

JackWilb:
> This sounds like another instance of #347. Did you encounter this error when testing the app locally? Or were you using the preview app?

This was on the preview app. It definitely seems related to #347.

> This is probably due to us having to explicitly register the redirect URLs on Google's side for OAuth. I'm guessing @waxlamp didn't add the preview app to the allowed list in the dev console. We don't have a great process for this yet.

I figured. Is there a way to test the login without adding the redirct URI to google? Even locally, right now?

multinet/api.py
```diff
@@ -24,7 +27,31 @@
...
+        # Otherwise, check to see if the workspace is public, or the user is at
+        # least a Reader of the workspace.
+        sub = user.sub
+        return (
+            perms["public"]
+            or perms["owner"] == sub
+            or sub in perms["readers"]
+            or sub in perms["writers"]
+            or sub in perms["maintainers"]
+        )
```
AlmightyYakob, line 49: I think this would be better placed outside of `get_workspaces` (and probably renamed to something like `workspace_readable`), as nothing in this function relies on anything besides the input parameters. 

Alternatively, we could make a general `workspace_permissions` function (name not withstanding), which would take a workspace and a user, but would instead return all of the permissions the user has on that workspace, instead of just read, just write, etc. This would have a new signature that describes a user's permissions on a workspace, something like:

```
{
    "owner": bool,
    "reader": bool,
    "writer": bool,
    "maintainer": bool
}
```

Then below, instead of `can_read(user, w)`, you'd replace it with `workspace_permissions(user, w)["reader"]`.

I'd also think that if we were to create this function, we'd put it in either `db.py` or `util.py`.

multinet/api.py
```diff
@@ -24,7 +27,31 @@
 @swag_from("swagger/workspaces.yaml")
 def get_workspaces() -> Any:
     """Retrieve list of workspaces."""
-    return util.stream(db.get_workspaces())
+    user = session_user()
+    current_app.logger.info(user)
```
AlmightyYakob, line 31: Is this left over from debugging?

multinet/db.py
```diff
@@ -114,18 +113,17 @@ def workspace_exists_internal(name: str) -> bool:
...
-    new_doc = {"name": name, "internal": generate_arango_workspace_name()}
-    coll.insert(new_doc)
+    coll.insert(ws_doc)
 
     try:
-        db("_system").create_database(new_doc["internal"])
+        db("_system").create_database(ws_doc["internal"])
     except DatabaseCreateError:
         # Could only happen if there's a name collisison
         raise InternalServerError()
```
AlmightyYakob, line 129: This may not be strictly related to this PR, but I think here we should wait to insert the workspace document into the mapping collection until after the database creation has succeeded. The way that things are currently implemented, we could have dangling entries in the workspace mapping collection. I'm okay with addressing this in a separate issue if that's what we want to do.

multinet/types.py
```diff
@@ -1,10 +1,25 @@
...
+from typing import Dict, Set, List
 from typing_extensions import Literal, TypedDict
 
 EdgeDirection = Literal["all", "incoming", "outgoing"]
 TableType = Literal["all", "node", "edge"]
 
+WorkspacePermissions = TypedDict(
+    "WorkspacePermissions",
+    {
+        "owner": str,
```
AlmightyYakob, line 11: To clarify, this maps to the `sub` field of a user? If so, it might be good to add a comment here to explain that.

jjnesbitt (changes requested):

AlmightyYakob:
> > This is probably due to us having to explicitly register the redirect URLs on Google's side for OAuth. I'm guessing @waxlamp didn't add the preview app to the allowed list in the dev console. We don't have a great process for this yet.
> 
> I figured. Is there a way to test the login without adding the redirct URI to google? Even locally, right now?

I don't think so, as the login always calls out to Google. However, we can temporarily add the preview app to the allowed origins in the Google dev console, for testing this. Any objections @waxlamp?

AlmightyYakob:
> > This sounds like another instance of #347. Did you encounter this error when testing the app locally? Or were you using the preview app?
> 
> I don't think so--this happens because of how we turn an iterable into a JSON response, [here](https://github.com/multinet-app/multinet-server/blob/master/multinet/util.py#L65-L74). If an exception is raised during the "meat" of this function, you'll get a malformed JSON response. It is not a great way to do this, so I welcome any ideas for improving it.

My bad, what I meant was that this same issue with the `generate` function that is documented in #347. We should probably address the bug with the `generate` function separately from addressing #347, as it's a more general bug.

devops/scripts/ensure_workspace_metadata.py
```diff
@@ -0,0 +1,77 @@
+from arango import ArangoClient
+import getpass
+import sys
+
+from mypy_extensions import TypedDict
```
JackWilb, line 5: mypy_extensions are not in the Pipfile which causes errors on fresh installs. It may be worth adding it in, even as a dev dependency

JackWilb (commented):
Just a little note here in this review that I noticed on a fresh install.

Also, I think it would definitely be worth testing the login and ownership before we push, I'm not sure how best to do that. We could set the preview to point at the test arango db instance and add in the preview for the google logins. That means the main database wouldn't be affected until we're ready to go.

We'll also need to think about using sensitive/not public datasets on our applications and how to handle those logins. Would users just need to log into multinet first and be (at least) a reader of the workspace?

multinet/user.py
```diff
@@ -92,6 +95,15 @@ def load_user_from_cookie(cookie: str) -> Optional[User]:
         return None
 
 
+def session_user() -> Optional[User]:
```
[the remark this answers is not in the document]
waxlamp: See 8754b2a73aaab43a5135b9a4eb291e83dc117ee9.

multinet/user.py
```diff
@@ -82,7 +85,7 @@ def delete_user_cookie(user: User) -> User:
     return updated_user(user_copy)
 
 
-def load_user_from_cookie(cookie: str) -> Optional[User]:
+def cookie_user(cookie: str) -> Optional[User]:
```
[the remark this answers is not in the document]
waxlamp: In discussions we settled on `user_from_cookie()` for the name of this function (eliminating the imperative-sounding verb from the function name).

See c58546db94ec3da1883669e01e37ad4652715b00.

multinet/api.py
```diff
@@ -112,7 +139,29 @@ def get_node_edges(
...
+        "name": workspace,
+        "internal": util.generate_arango_workspace_name(),
+        "permissions": {
+            "owner": user.sub,
+            "maintainers": [],
+            "writers": [],
+            "readers": [],
+            "public": False,
+        },
+    }
```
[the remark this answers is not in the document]
waxlamp: All excellent suggestions. See 8cd1fb5052075ff635d704726ae8a44997c7d60b (please make sure I addressed all points).

multinet/api.py
```diff
@@ -112,7 +139,29 @@ def get_node_edges(
...
+
+    # Perform the actual backend update.
+    db.create_workspace(workspace, new_doc)
     return workspace
```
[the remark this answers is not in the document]
waxlamp: See 8cd1fb5052075ff635d704726ae8a44997c7d60b.

multinet/db.py
waxlamp: See 255bfb6d356cc60a89f78d13e8d808ab82d49189.

In general, we may want to write a sort of fsck script to validate our server data state. See #398.

multinet/types.py
waxlamp: See ca96a922d26c192cce8fbcada77024cca62196e6.

devops/scripts/ensure_workspace_metadata.py
waxlamp: Good catch. For that matter, `mypy` isn't listed in `Pipfile` either. I will send this update on a separate branch, so as not to conflict with the content changes in this branch.

waxlamp:
> Also, I think it would definitely be worth testing the login and ownership before we push, I'm not sure how best to do that. We could set the preview to point at the test arango db instance and add in the preview for the google logins. That means the main database wouldn't be affected until we're ready to go.

I agree with you and @AlmightyYakob both that testing is really needed for this type of addition. However, since our testing infrastructure requires a bit of love, I'm going to defer that to after this PR is merged. In particular, there's a little bit of cleanup we can do, and some skipped tests that need to be unskipped, and we need to institute our policy of testing with the testing DB. That all seems like too much for this branch, so we'll live on the edge a bit until that's done.

> We'll also need to think about using sensitive/not public datasets on our applications and how to handle those logins. Would users just need to log into multinet first and be (at least) a reader of the workspace?

I'm not sure what you mean exactly--if there's sensitive data in the system, it would be owned by whoever uploaded it, and that user can grant permissions to view the workspace. We do need to make sure that people don't assume there's more security here than there actually is; I've been getting some advice from our legal folks on how this can work (most likely by posting a notice that explains that you need to remain responsible for your data, etc.).

devops/scripts/ensure_workspace_metadata.py
waxlamp: See #399.

multinet/api.py
waxlamp: In our permissions model, each role includes implicitly the roles that are "weaker"; that is, an owner of a workspace is also, by definition, a maintainer, writer, and reader. Since that's the case, and our needs are for functions that can filter by these roles, I'm going to create a collection of `-> bool` functions for each role, and create associated decorators to guard the use of API endpoints (in this case, we wouldn't use the decorator, since we're filtering all workspaces by that hypothetical `is_reader()` function).

I will push a commit shortly to do a bit of refactoring here.

JackWilb:
> > Also, I think it would definitely be worth testing the login and ownership before we push, I'm not sure how best to do that. We could set the preview to point at the test arango db instance and add in the preview for the google logins. That means the main database wouldn't be affected until we're ready to go.
> 
> I agree with you and @AlmightyYakob both that testing is really needed for this type of addition. However, since our testing infrastructure requires a bit of love, I'm going to defer that to after this PR is merged. In particular, there's a little bit of cleanup we can do, and some skipped tests that need to be unskipped, and we need to institute our policy of testing with the testing DB. That all seems like too much for this branch, so we'll live on the edge a bit until that's done.

I meant even some manual testing that things are working since it's difficult to verify that this PR is working as intended without updating the database to the new permissions model and testing it with logins. I agree that the automated tests can be a follow up PR.

> > We'll also need to think about using sensitive/not public datasets on our applications and how to handle those logins. Would users just need to log into multinet first and be (at least) a reader of the workspace?
> 
> I'm not sure what you mean exactly--if there's sensitive data in the system, it would be owned by whoever uploaded it, and that user can grant permissions to view the workspace. We do need to make sure that people don't assume there's more security here than there actually is; I've been getting some advice from our legal folks on how this can work (most likely by posting a notice that explains that you need to remain responsible for your data, etc.).

I mean for an app like Jen's evo-bio. If she is sharing data that is not public on there, how does she grant access to people to use data from multinet that she uploaded? Is the flow: she uploads, adds the person she wants to show it to as a viewer, and then they have to be logged into multinet in the background while viewing her app?

multinet/api.py
waxlamp: See 282cbf4101ae3b4e0a221453ce3f82a6bc16f997.

multinet/api.py
waxlamp: Yup :)

See a9e328c4e2404460df452d0107acd500db50acac.

multinet/api.py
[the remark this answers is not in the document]
waxlamp: (I think I did address everything, so I'll resolve this conversation.)

waxlamp:
> > > Also, I think it would definitely be worth testing the login and ownership before we push, I'm not sure how best to do that. We could set the preview to point at the test arango db instance and add in the preview for the google logins. That means the main database wouldn't be affected until we're ready to go.
> > 
> > 
> > I agree with you and @AlmightyYakob both that testing is really needed for this type of addition. However, since our testing infrastructure requires a bit of love, I'm going to defer that to after this PR is merged. In particular, there's a little bit of cleanup we can do, and some skipped tests that need to be unskipped, and we need to institute our policy of testing with the testing DB. That all seems like too much for this branch, so we'll live on the edge a bit until that's done.
> 
> I meant even some manual testing that things are working since it's difficult to verify that this PR is working as intended without updating the database to the new permissions model and testing it with logins. I agree that the automated tests can be a follow up PR. It's possible to test that workspaces are either public or hidden, but not that logging in reveals them to the owner of the workspace.
> 

Ah, gotcha.

I have been doing this type of manual testing locally. If you log in, create a few workspaces, and make some of them public, then log out (and refresh) you should see that the non-public workspaces are hidden from view.

(There is a bug where logging out doesn't refresh the workspace list, and that's something I should address either on this branch or in an immediate followup.)

Let me know if you're not able to build locally in order to test. I am able to do it, so there must be a way to make it work on your end too--let me know if you want to work it out together.

> > > We'll also need to think about using sensitive/not public datasets on our applications and how to handle those logins. Would users just need to log into multinet first and be (at least) a reader of the workspace?
> > 
> > 
> > I'm not sure what you mean exactly--if there's sensitive data in the system, it would be owned by whoever uploaded it, and that user can grant permissions to view the workspace. We do need to make sure that people don't assume there's more security here than there actually is; I've been getting some advice from our legal folks on how this can work (most likely by posting a notice that explains that you need to remain responsible for your data, etc.).
> 
> I mean for an app like Jen's evo-bio. If she is sharing data that is not public on there, how does she grant access to people to use data from multinet that she uploaded? Is the flow: she uploads, adds the person she wants to show it to as a viewer, and then they have to be logged into multinet in the background while viewing her app?

That's exactly the workflow as far as I know. In Jen's particular case, the data was always public, so anyone can view it in Trevo without being logged in. But if it were non-public data, then it would go exactly as you said.

We'll need to test this out in practice to see if it works properly.

waxlamp:
> (There is a bug where logging out doesn't refresh the workspace list, and that's something I should address either on this branch or in an immediate followup.)

I just realized this is a *client* bug so of course I won't fix it on this branch.  See https://github.com/multinet-app/multinet-client/issues/103.

multinet/api.py
```diff
@@ -24,7 +27,30 @@
...
+        # Otherwise, check to see if the workspace is public, or the user is at
+        # least a Reader of the workspace.
+        sub = user.sub
+        return (
+            perms["public"]
+            or perms["owner"] == sub
+            or sub in perms["readers"]
+            or sub in perms["writers"]
+            or sub in perms["maintainers"]
+        )
```
JackWilb, line 48: You refactored this, but left the old function here. Looks to me like this can be removed

multinet/auth/util.py
```diff
@@ -0,0 +1,44 @@
...
+    """Decorate an API endpoint to check for a logged in user."""
+
+    def wrapper(workspace: str, *args: Any, **kwargs: Any) -> Any:
+        user = current_user()
+        if user is None:
+            raise Unauthorized("You must be logged in to create new workspaces")
+
+        return f(workspace, *args, **kwargs)
+
+    return wrapper
```
JackWilb, line 24: Thinking ahead to when we implement different actions that will require logins (making tables and networks, for example), we'll probably want this to be a bit more general. Is it worth doing the work to make it general now or deferring it until we modify the other endpoints to require a log in?

JackWilb (changes requested):
Just a couple notes, but this looks really close

multinet/auth/util.py
waxlamp: Do you mean the error message or the whole function?

The error message should be fixed now; I'll take care of that shortly.

multinet/api.py
waxlamp: Good catch. See 4f2d28d18263d845646c3cd976d467943cf7d88e.

multinet/auth/util.py
waxlamp: See 1387677bfceabc80985d05d29cea2d084ed45d74.

JackWilb: Works for me

JackWilb (dismissed):

JackWilb (dismissed):

multinet/auth/util.py
```diff
@@ -0,0 +1,44 @@
...
+    # Otherwise, check to see if the workspace is public, or the user is at
+    # least a Reader of the workspace.
+    sub = user.sub
+    return (
+        perms["public"]
+        or perms["owner"] == sub
+        or sub in perms["readers"]
+        or sub in perms["writers"]
+        or sub in perms["maintainers"]
+    )
```
AlmightyYakob, line 44: Small thing, but I think as a matter of principle, we should check permissions from most common to least common. Since the role of `owner` is the least common, I'd suggest the following:
```suggestion
    return (
        perms["public"]
        or sub in perms["readers"]
        or sub in perms["writers"]
        or sub in perms["maintainers"]
        or perms["owner"] == sub
    )
```

jjnesbitt (changes requested):

multinet/auth/util.py
```diff
@@ -0,0 +1,44 @@
...
+    """Decorate an API endpoint to check for a logged in user."""
+
+    def wrapper(workspace: str, *args: Any, **kwargs: Any) -> Any:
+        user = current_user()
+        if user is None:
+            raise Unauthorized("You must be logged in to perform this action")
+
+        return f(workspace, *args, **kwargs)
+
+    return wrapper
```
AlmightyYakob, line 24: Fixing my other suggestion, as it was out of date. Sorry it's so large, I wanted to include the import:

```suggestion
from typing import Any, Optional, Callable

from multinet.errors import Unauthorized
from multinet.types import Workspace
from multinet.auth.types import UserInfo
from multinet.user import current_user


# NOTE: unfortunately, it is difficult to write a type signature for this
# decorator. I've opened an issue to ask about this here:
# https://github.com/python/mypy/issues/9032.
def require_login(f: Callable) -> Callable:
    """Decorate an API endpoint to check for a logged in user."""

    def wrapper(*args: Any, **kwargs: Any) -> Any:
        user = current_user()
        if user is None:
            raise Unauthorized("You must be logged in to perform this action")

        return f(*args, **kwargs)

    return wrapper
```

multinet/api.py
```diff
@@ -137,32 +136,18 @@ def get_node_edges(
...
-        },
-    }
-
-    # Perform the actual backend update.
-    db.create_workspace(workspace, new_doc)
-    return workspace
+    # The `require_login()` decorator ensures that a user is logged in by this
+    # point.
+    user = current_user()
+    assert user is not None
```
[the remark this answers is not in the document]
waxlamp: This is actually for the typechecker. `current_user()` returns an `Optional[User]` so this assert statement is necessary to narrow the type down to `User`. The use of `@require_login` ensures that in this instance the function never returns `None`.

jjnesbitt (approved):
:shipit:

JackWilb (approved):

Issue #368, opened 2020-04-24
Un-skip Workspace tests


AlmightyYakob:
Once we've settled on a strategy for using a database in CI, we need to remove the `pytest.mark.skip` decorators introduced by 1791a45.

AlmightyYakob:
Closed via #400

PR #400, opened 2020-06-22
Add pytest fixtures that allow for api testing

AlmightyYakob:
This change allows for making requests to the api, exposed through the `server` fixture in `test/conftest.py`. Making a request in a testing function would look like so:

```
def test_something(server):
    resp = server.get("api/workspaces")
    assert "expected_workspace" in resp.json
```

The `server` fixture is of the type `FlaskClient`, which is an extension of the regular `Client` class, documented [here](https://werkzeug.palletsprojects.com/en/1.0.x/test/#werkzeug.test.Client).


AlmightyYakob:
> This is awesome. Could you add a really simple test that makes use of this so we have something to model as we write more integration tests?

Sure, I can try to find something.

waxlamp:
> > This is awesome. Could you add a really simple test that makes use of this so we have something to model as we write more integration tests?
> 
> Sure, I can try to find something.

I'm talking as simple as, use the handled workspace fixture to verify that the workspace exists in the workspace mapping collection and in the DB itself.

AlmightyYakob:
This PR now sets up CI to use the testing database, which will allow us to address #368.

AlmightyYakob:
The reason I added the `.env.base` file, was that the `.env.default` file is more of an assumption of the local dev environment, as well as a template for possible env vars, and some things in that would be overwritten by the needs of the CI environment. Rather then place duplicate env vars at the end of the copied `.env` file, I opted to create this new file, which specifies the bare minimum things that most likely won't change. This allows the specification of environment variables in the CI, without having to append to the `.env` file.

.env.base
```diff
@@ -0,0 +1,7 @@
+FLASK_SERVE_PORT=5000
+DOC_SERVE_PORT=8000
+
+FLASK_APP=multinet
+FLASK_ENV=development
+
+MYPYPATH=mypy_stubs
```
waxlamp, line 7: Can you explain further why `.env.default` isn't good enough?

multinet/__init__.py
```diff
@@ -34,6 +34,9 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     """Create a Multinet app instance."""
     app = Flask(__name__)
 
+    if config is not None:
+        app.config.update(config)
```
waxlamp, line 38: What does this do? Why wasn't it needed before now?

multinet/db.py
```diff
@@ -153,6 +153,7 @@ def rename_workspace(old_name: str, new_name: str) -> None:
     workspace_mapping.cache_clear()
 
 
+# TODO: Add internal function for deleting workspace
```
waxlamp, line 156: Why?

test/conftest.py
```diff
@@ -0,0 +1,43 @@
+"""Pytest configurations for multinet tests."""
```
waxlamp, line 1: Is `conftest.py` a standard name for a module that defines PyTest fixtures?

test/test_meta.py
```diff
@@ -0,0 +1,7 @@
+"""Tests that ensure fixtures act properly."""
+
+
+def test_generated_workspace(handled_workspace, server):
```
waxlamp, line 4: Just having an argument named `handled_workspace` is enough to trigger the fixture?

test/conftest.py
AlmightyYakob: Yes https://stackoverflow.com/questions/34466027/in-pytest-what-is-the-use-of-conftest-py-files

test/test_meta.py
AlmightyYakob: Yep, that's the beauty of pytest.

multinet/db.py
AlmightyYakob: Should've made this more clear, what I meant was that there should be a function that deletes a workspace given it's `internal` name. It seems like a generally useful function, only really to be used in edge cases or worst case scenarios. I thought of it when looking at the [`test_workspace_rename`](https://github.com/multinet-app/multinet-server/blob/acc8864e8f7f114b15b5f380ce29540955957088/test/test_workspace.py#L84) test. If we had this internal function, we could use this in a way similar to the `handled_workspace` fixture.

multinet/__init__.py
AlmightyYakob: Before, we had the `config` parameter in this function, but we were never using it. We need to use it to allow setting the `TESTING` flag, for test servers. https://github.com/multinet-app/multinet-server/blob/acc8864e8f7f114b15b5f380ce29540955957088/test/conftest.py#L13

.env.base
AlmightyYakob: My [above comment](https://github.com/multinet-app/multinet-server/pull/400#issuecomment-648321332) summarizes this pretty well I think. TLDR: `.env.default` specifies more environment variables than we need/want for testing.

multinet/db.py
waxlamp: Gotcha. Can you file an issue to do this and remove the comment?

.env.base
waxlamp: It seems this file is needed to make testing in CI easier, so I'd prefer to call it `.env.testing`. We may want to think about renaming `.env.default` to `.env.development` or something like that in the same vein.

I'm opposed to naming this `.env.base`, even if that's descriptive of its basiness, because it implies that this is something we expect people to build upon, when really it isn't. It seems to come down to: this .env file is useful for testing, so we should keep it, but it should be named explicitly after its purpose.

AlmightyYakob: I'm also in favor of changing `.env.default` to `.env.development`, not sure if that change should be done here. I'm okay with changing `.env.base` to `.env.testing` though, so I'll do that.

AlmightyYakob: f3fb91d

multinet/db.py
AlmightyYakob: 435b35c

.env.base
waxlamp: > I'm also in favor of changing `.env.default` to `.env.development`, not sure if that change should be done here. I'm okay with changing `.env.base` to `.env.testing` though, so I'll do that.

It will require some changes to documentation, so I've opened #410 to track it.

AlmightyYakob:
@waxlamp I needed to update some tests to reflect new things from master, and I also unskipped the workspace tests, since we have the testing db now.

test/conftest.py
```diff
@@ -22,12 +24,38 @@ def server(app):
...
+                email="test@test.test",
+                sub=uuid4().hex,
+            )
+        )
+    )
+
+    yield user
+
+    # TODO: Once the function exists, use `delete_user` here instead
+    user_collection().delete(asdict(user))
```
waxlamp, line 49: Will the `delete_user()` function really be this simple? If yes, then I'd say go ahead and refactor it now.

test/test_meta.py
```diff
@@ -1,7 +1,13 @@
 """Tests that ensure fixtures act properly."""
 
+from multinet.user import MULTINET_COOKIE
 
-def test_generated_workspace(handled_workspace, server):
+
+def test_generated_workspace(handled_workspace, handled_user, server):
     """Test that a generated workspace exists when querying the API."""
+
+    with server.session_transaction() as session:
```
waxlamp, line 9: keanu_saying_whoa.gif

AlmightyYakob: :sunglasses:

test/conftest.py
AlmightyYakob: I'm not sure, there may be other things to consider when creating that function. Since this anti-pattern only exists here in the tests, I think it's okay to leave it until it's addressed on its own.

PR #402, opened 2020-06-22
Add @require_reader decorator where appropriate

waxlamp:
This PR demonstrates how to create a route decorator for enforcing the workspace "reader" role.

If this passes muster, I will plan to create the other three decorators and place them on the appropriate route functions as well.

Depends on #396


multinet/auth/util.py
```diff
@@ -14,6 +17,7 @@
 def require_login(f: Any) -> Any:
     """Decorate an API endpoint to check for a logged in user."""
 
+    @functools.wraps(f)
```
AlmightyYakob, line 20: Is this just done to allow access to the underlying function?

multinet/swagger/template.yaml
```diff
@@ -14,35 +14,46 @@ definitions:
...
       owner: laforge
       readers:
         - worf
         - data
         - troi
-      writers:
+      maintainers:
         - picard
         - riker
+      writers: []
```
AlmightyYakob, line 56: This example is missing the `public` flag

multinet/types.py
```diff
@@ -16,9 +16,19 @@
...
+
+    name: str
+    internal: str
+    permissions: WorkspacePermissions
+
+
+class WorkspaceDocument(Workspace):
+    """Workspace metadata as it appears in ArangoDB."""
+
+    _id: str
```
AlmightyYakob, line 31: There's also the `_key` and `_rev` fields on an arango document:
```suggestion
    _id: str
    _key: str
    _rev: str
```

multinet/api.py
```diff
@@ -35,13 +35,25 @@ def get_workspaces() -> Any:
...
+    perms = metadata["permissions"]
+
+    return {
+        "name": metadata["name"],
+        "owner": perms["owner"],
+        "maintainers": perms["maintainers"],
+        "writers": perms["writers"],
+        "readers": perms["readers"],
+        "public": perms["public"],
+    }
```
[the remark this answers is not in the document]
waxlamp: Mainly because that includes ArangoDB metadata. So while I was plucking the useful data out of that, I thought I would flatten the structure of the return value as well. This is in the interest of keeping API return values as simple as possible.

multinet/auth/util.py
waxlamp: I think I explained it in a commit message, but without this decorator, the name of the returned function becomes the local name of the decorator function, which causes trouble for Flask--it thinks you're defining multiple routes named "wrapped" (or whatever the local decorator function name is).

multinet/swagger/template.yaml
waxlamp: Good catch. See d08ef44d9251f21c6e9e865a43c0323415d7fda2.

Of course, the example is already really wrong, because those entries should be `sub` values, not usernames.

multinet/api.py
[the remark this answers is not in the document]
AlmightyYakob: I recognize your concern, however `metadata["permissions"]` shouldn't contain any arango metadata, since that's included on the root level of `metadata`, and not the subfield of `permissions`.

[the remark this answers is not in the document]
waxlamp: That is a good point. I am also storing the name at top level of the return value. But thinking about it some more, I can see how we might wish to extend this in the future by adding other sections, like a list of tables, list of networks, etc., which could be optionally included in the response. I will do something closer to what you've suggested.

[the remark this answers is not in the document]
waxlamp: See b8185a72d1e30db359f6f2baa4d9cc69ffd3a41c.

AlmightyYakob:
@waxlamp it seems you've addressed all of my concerns. Is your plan still create the other role decorators in this PR, or a follow-up one?

waxlamp:
> @waxlamp it seems you've addressed all of my concerns. Is your plan still create the other role decorators in this PR, or a follow-up one?

Follow-up.

PR #403, opened 2020-06-23
Cast _key field to str on graph/table upload

AlmightyYakob:
This fixes a bug @gotdairyya and I found, where if you tried to upload a graph in D3 JSON format, if the value of the `id` field for each node was a number, it would silently fail (as numeric `_key` fields aren't allowed in arango), and skip these documents, leaving the node table empty. This is partly because we weren't casting the value of the `_key` field to a string (which this PR fixes). However, another reason this was hard to spot was that for some reason (as I mentioned above), the `insert_many` call would silently fail, instead of throwing an error, and it wasn't until I tried a singular insert that I found the source of the bug. I'm not sure what we can do about this, but at the very least it's something to be aware of.

If we are okay with casting the `_key` field to a string (I don't see a downside to it), then I can apply this approach to other uploaders where it may apply (for example, it doesn't apply to csv, as it treats everything as strings anyway).


waxlamp:
> This fixes a bug @gotdairyya and I found, where if you tried to upload a graph in D3 JSON format, if the value of the `id` field for each node was a number, it would silently fail (as numeric `_key` fields aren't allowed in arango), and skip these documents, leaving the node table empty. This is partly because we weren't casting the value of the `_key` field to a string (which this PR fixes). However, another reason this was hard to spot was that for some reason (as I mentioned above), the `insert_many` call would silently fail, instead of throwing an error, and it wasn't until I tried a singular insert that I found the source of the bug. I'm not sure what we can do about this, but at the very least it's something to be aware of.
> 

It's not _really_ silently failing--just very quietly failing. According to [the `insert_many()` docs](https://python-driver-for-arangodb.readthedocs.io/en/4.4.0/specs.html#arango.collection.StandardCollection.insert_many), instead of raising an exception, it places the exception object into a list which is the return value. That way, you can see exactly which inserts failed.

But, we don't truly care about that. There's another function, [`import_bulk()`](https://python-driver-for-arangodb.readthedocs.io/en/4.4.0/specs.html#arango.collection.StandardCollection.import_bulk), which _does_ raise an exception on failure, and therefore can operate faster than `insert_many()`. I wonder if we should be using this function here since we don't expect this failure mode very often and could thus have faster bulk import of data.

> If we are okay with casting the `_key` field to a string (I don't see a downside to it), then I can apply this approach to other uploaders where it may apply (for example, it doesn't apply to csv, as it treats everything as strings anyway).

The other alternative is to fail validation for non-string keys. What do you think about that? I would argue it's safer, but I can't tell if the extra safety is worth the extra trouble to the client. It probably is, but I'd like to hear your thoughts.

AlmightyYakob:
> The only downside I can think of to casting the `_key` values to `str` this way is that it could cause confusion for the end user in atypical situations. Supposing the user has some weird value type in their `id` field, and they expect it to work; when this cast happens, maybe Python generates something akin to JavaScript's `[object Object]`, which then goes on to fail validation, and the resulting error message is not very helpful. Part of me thinks it would be better to simply fail validation early if the `id` field is not already a string, but a bigger part of me thinks this solution is fine for the typical case.

I see what you mean. However, specifically in this uploader, it's using JSON, so the set of allowed value types is well defined. I'll investigate if any other uploaders need to have this behavior added as well.

AlmightyYakob:
> There's another function, [`import_bulk()`](https://python-driver-for-arangodb.readthedocs.io/en/4.4.0/specs.html#arango.collection.StandardCollection.import_bulk), which _does_ raise an exception on failure, and therefore can operate faster than `insert_many()`. I wonder if we should be using this function here since we don't expect this failure mode very often and could thus have faster bulk import of data.

Ahh good find. I think using that function instead would be better.

> The other alternative is to fail validation for non-string keys. What do you think about that? I would argue it's safer, but I can't tell if the extra safety is worth the extra trouble to the client. It probably is, but I'd like to hear your thoughts.

I'm not sure. If just interacting with the data in multinet, the user wouldn't notice any difference, as when referenced elsewhere, these `_key` fields are referenced as strings anyway (because they need to be), and just casting it to a `str` would allow for that easily. I'm not sure how many use cases there are where casting it to a string would be a bad idea. Maybe input from @gotdairyya is helpful here.

waxlamp:
> > The other alternative is to fail validation for non-string keys. What do you think about that? I would argue it's safer, but I can't tell if the extra safety is worth the extra trouble to the client. It probably is, but I'd like to hear your thoughts.
> 
> I'm not sure. If just interacting with the data in multinet, the user wouldn't notice any difference, as when referenced elsewhere, these `_key` fields are referenced as strings anyway (because they need to be), and just casting it to a `str` would allow for that easily. I'm not sure how many use cases there are where casting it to a string would be a bad idea. Maybe input from @gotdairyya is helpful here.

I think casting to str is probably fine right now (i.e., merge this PR whenever you are ready).

Just for discussion's sake, here's a pretty crazy edge case where someone could run into trouble possibly: imagine two items with numeric keys of `1e2` and `100.0` respectively. In Python:

```python
>>> str(1e2)
'100.0'
>>> str(100.0)
'100.0'
```

So if the user meant these to be strings, then the validation failure will be confusing.

But if you're willfully using IEEE floating point number representations as your database item keys, then you probably have bigger problems than how Multinet tries to coerce your values to valid Arango keys.

PR #404, opened 2020-06-23
Create tables from AQL queries

AlmightyYakob:
Depends on #400.

~~With a readonly user provisioned, I've tested it working locally no problem. However, I was having some trouble provisioning this automatically with `arangosh` and ansible. If I can't figure this out, I do know how to provision the server manually, although I hope it doesn't come to that.~~

Todo:

- [x] Create read-only credentials for production db
- [x] Create read-only credentials for testing db
- [x] Add production read-only credentials into heroku
- [x] Add testing read-only credentials into github secrets


devops/arangodb.yml
```diff
@@ -46,6 +46,14 @@
       shell: ARANGODB_DEFAULT_ROOT_PASSWORD={{ arangodb_root_password }} arango-secure-installation
       become: true
 
+    # - name: Ensure there is a read-only user
+    #   executable: /usr/bin/arangosh --server.password {{ arangodb_root_password }}
```
JackWilb, line 50: Looks like you probably just need more flags when you run this. See [here](https://www.arangodb.com/docs/stable/programs-arangosh-examples.html). The main ones are the server protocol and probably passing in the user. Here's a quote from that doc that seems relevant:

> If the server endpoint is configured for SSL then clients such as arangosh need to connect to it using an SSL socket as well. For example, use http+ssl:// as schema in --server.endpoint for an SSL-secured HTTP connection.

You might also need to append [javascript.execute-string](https://www.arangodb.com/docs/stable/programs-arangosh-options.html#javascript) so it doesn't drop you into a shell at this step.

AlmightyYakob: I've done more work locally to try to make this work, which does include extra parameters, but ansible doesn't seem to allow appending arguments to the executable specifier. I'm not sure how to get around this in a sane way.

JackWilb: Can you not run is as a shell command like the `arango-secure-installer` above?

AlmightyYakob: That command is running a command with the `shell` command, using the default shell (probably `sh` or `bash`). The problem for me isn't the commands themselves, it's that the executable I want to use (`arangosh`) needs a password specified, and ansible recognizes the whole command (with arguments) as the executable, sayings it's not found.

AlmightyYakob: With help from @waxlamp, this has been resolved in 45f96ac

AlmightyYakob:
I've tested this new step for creating a read-only user against the testing database, and it was successful, which is what allows all of the tests to pass in CI.

devops/arangodb.yml
```diff
@@ -46,6 +52,10 @@
       shell: ARANGODB_DEFAULT_ROOT_PASSWORD={{ arangodb_root_password }} arango-secure-installation
       become: true
 
+    - name: Ensure there is a read-only user
+      command: arangosh --server.password {{ arangodb_root_password }} --server.endpoint ssl://127.0.0.1:8529 --javascript.execute-string "const users = require('@arangodb/users'); if (!users.exists('{{ arango_readonly_username }}')) { users.save('{{ arango_readonly_username }}', '{{ arango_readonly_password }}'); } users.grantDatabase('{{ arango_readonly_username }}', '*', 'ro');"
```
waxlamp, line 56: For readability, can you pull the script into a file, and use the appropriate `arangosh` flag to read from that instead? If that simple transformation doesn't work because of variable interpolation, there is also the [template](https://docs.ansible.com/ansible/latest/modules/template_module.html) module.

multinet/api.py
```diff
@@ -43,6 +43,17 @@ def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:  # noq
     return util.stream(tables)
 
 
+@bp.route("/workspaces/<workspace>/tables", methods=["POST"])
+@use_kwargs({"table": fields.Str()})
+@swag_from("swagger/workspace_aql_tables.yaml")
+def create_table_from_aql(workspace: str, table: str) -> Any:
```
waxlamp, line 49: I prefer shorter function names when possible. This should be `create_aql_table`.

multinet/api.py
```diff
@@ -43,6 +43,17 @@ def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:  # noq
...
+    """Retrieve the tables of a single workspace."""
+    aql = request.data.decode()
+    table = db.create_workspace_table_from_aql(workspace, table, aql)
```
waxlamp, line 52: Similarly, this should just be `db.create_aql_table`.

multinet/db.py
```diff
@@ -47,7 +53,20 @@
...
+        verify=True,
+    )
+
+
+def read_only_db(name: str) -> StandardDatabase:
+    """Return a read-only handle for the Arango database `name`."""
+    return arango.db(
+        name,
+        username=str(os.environ.get("ARANGO_READONLY_USERNAME", "")),
+        password=str(os.environ.get("ARANGO_READONLY_PASSWORD", "")),
```
waxlamp, line 68: This doesn't seem right. The fallback case essentially guarantees runtime failure.

Can we perhaps think about setting the username statically (since we do this for the root user already)--something like `readonly` or `reader`? And it seems to me the default password should also be `letmein` so that simple, local setups are easier to experiment with.

multinet/db.py
```diff
@@ -181,12 +200,15 @@ def get_workspace(name: str) -> WorkspaceSpec:
...
+def get_workspace_db(name: str, readonly: bool = False) -> StandardDatabase:
     """Return the Arango database associated with a workspace, if it exists."""
     doc = workspace_mapping(name)
     if not doc:
         raise WorkspaceNotFound(name)
 
+    if readonly:
+        return read_only_db(doc["internal"])
+
     return db(doc["internal"])
```
waxlamp, line 212: The following is more pythonic:

```suggestion
    name = doc["internal"]
    return read_only_db(name) if readonly else db(name)
```

multinet/db.py
```diff
@@ -301,6 +323,31 @@ def workspace_table_keys(
     return keys
 
 
+# TODO: Ensure current user has writer role
+def create_workspace_table_from_aql(workspace: str, name: str, aql: str) -> str:
+    """Create a new workspace from an aql query."""
```
waxlamp, line 328:
```suggestion
    """Create a new table from an AQL query."""
```

multinet/db.py
```diff
@@ -301,6 +323,31 @@ def workspace_table_keys(
...
+    if db.has_collection(name):
+        raise AlreadyExists("table", name)
+
+    # In the future, the result of this validation can be
+    # used to determine dependencies in virtual tables
+    try:
+        db.aql.validate(aql)
+        rows = list(db.aql.execute(aql))
+    except (AQLQueryValidateError, AQLQueryExecuteError) as e:
+        raise MalformedRequestBody(str(e))
```
waxlamp, line 340: This doesn't seem right--it's not that the request body is bad, it's that the AQL query was bad. We probably need a specific 400-series error for this.

multinet/db.py
```diff
@@ -301,6 +323,31 @@ def workspace_table_keys(
...
+    db = get_workspace_db(workspace, readonly=True)
+
+    if db.has_collection(name):
+        raise AlreadyExists("table", name)
+
+    # In the future, the result of this validation can be
+    # used to determine dependencies in virtual tables
+    try:
+        db.aql.validate(aql)
+        rows = list(db.aql.execute(aql))
```
waxlamp, line 338: Can we get away with not coercing this to a list here? Database queries are, in general, supposed to be able to handle lots of data and return lots of results.

multinet/db.py
```diff
@@ -387,7 +434,12 @@ def aql_query(workspace: str, query: str) -> Generator[Any, None, None]:
     """Perform an AQL query in the given workspace."""
     aql = get_workspace_db(workspace).aql
 
-    cursor = aql.execute(query)
+    try:
+        aql.validate(query)
+        cursor = aql.execute(query)
+    except (AQLQueryValidateError, AQLQueryExecuteError) as e:
+        raise MalformedRequestBody(str(e))
```
waxlamp, line 441: This chunk should probably become a function (it's largely identical to a code hunk above).

multinet/swagger/workspace_aql_tables.yaml
```diff
@@ -0,0 +1,35 @@
+Create a table from an AQL query
+---
+parameters:
+  - $ref: "#/parameters/workspace"
+  -
+    name: table
```
waxlamp, line 6: I prefer to avoid the linebreak here:

```suggestion
  - name: table
```

test/conftest.py
```diff
@@ -41,3 +50,26 @@ def handled_workspace(generated_workspace):
...
+
+@pytest.fixture
+def populated_workspace(
+    handled_workspace, data_directory, server
+) -> Tuple[str, Tuple[List[str], List[str]]]:
+    """
+    Populate a workspace with some data.
+
+    Returns a Nested Tuple of the form:
+    `(workspace_name: str, (graphs: List[str], tables: List[str]))`
```
waxlamp, line 63: Why a Tuple? A TypedDict would generally be preferable.

multinet/db.py
AlmightyYakob: The guaranteed runtime failure was by design, as if we're missing credentials we'll want to know, and I didn't want to create a multinet error for it, as it's not useful for the user to know this, just us. However, it would also necessitate that we've provisioned the db with this user, so it would also fail in that case with defaults. I think your idea is good, I can add it.

multinet/db.py
AlmightyYakob: The AQL query is sent in the request body, so it seems fitting to me :man_shrugging:. But if you feel strongly about creating a new error, I'm not opposed to it.

multinet/db.py
AlmightyYakob: We use it in both `validate_csv` and `coll.insert_many` below. If we were not to coerce it, I believe the cursor would be exhausted by the time we tried to insert the documents into the collection.

test/conftest.py
AlmightyYakob: I thought a tuple felt like a more natural way to return this, plus it allows for argument destructuring, like done here: https://github.com/multinet-app/multinet-server/blob/45f96accf551dd743af71233d9dc8c0f9412b2d6/test/test_aql_table_creation.py#L54-L56

But I don't feel too strongly about it.

devops/arangodb.yml
AlmightyYakob: Ahh, I wasn't aware of the template module. I'll try this out.

multinet/db.py
AlmightyYakob: 8d4355d

multinet/db.py
waxlamp: This is a case where a user would get "malformed body" as an error, think oh I must have sent something in the wrong shape, then dig around, then get annoyed when they find out it's actually a syntax error in their AQL query. Since we already know it's a more specific error, it makes sense to send back a more specific error condition.

Does the validation step reveal syntax errors? and what kind of errors occur at execution time? I think having a couple of AQL-based error classes (in the 400-series) makes sense here.

multinet/db.py
waxlamp: Ok let's leave it as is for now. I have opened #408 to address it at some point.

test/conftest.py
waxlamp: The destructuring assignment is definitely nice for writing tests, and the type annotations save it from being the disaster that tuples used to be. But the nesting adds complexity without much of an upside. This would read and write easier as

```python
workspace, graphs, node_table, edge_table = populated_workspace
```

If you make that change then I'm happy with this.

multinet/db.py
AlmightyYakob: Yeah I agree, like having separate aql validation and execution errors. The validation step does reveal syntax, which is what is passed into the raised error as `str(e)`

multinet/db.py
```diff
@@ -64,8 +66,8 @@ def read_only_db(name: str) -> StandardDatabase:
     """Return a read-only handle for the Arango database `name`."""
     return arango.db(
         name,
-        username=str(os.environ.get("ARANGO_READONLY_USERNAME", "")),
-        password=str(os.environ.get("ARANGO_READONLY_PASSWORD", "")),
+        username=str(os.environ.get("ARANGO_READONLY_USERNAME", "readonly")),
```
waxlamp, line 69: My thinking was to remove this choice altogether. It's one extra knob that's not really needed. We just call the user `readonly`, let the admin set the password, and leave it at that.

The justification is to leave the Ansible configuration stuff as dead simple as we can, since it's kind of the wart in all of this.

multinet/db.py
```diff
@@ -382,12 +425,20 @@ def delete_table(workspace: str, table: str) -> str:
...
+        cursor = aql.execute(query)
+    except (AQLQueryValidateError, AQLQueryExecuteError) as e:
+        raise MalformedRequestBody(str(e))
+
+    return cursor
+
+
 def aql_query(workspace: str, query: str) -> Generator[Any, None, None]:
     """Perform an AQL query in the given workspace."""
     aql = get_workspace_db(workspace).aql
```
waxlamp, line 440: Shouldn't this be called with `readonly = True`?

It makes me wonder if we should flip the polarity of the default value--make `get_workspace_db()` readonly by default, and force us to call it with a flag if we want a writable DB. I know that sounds obnoxious, but it's definitely more fail-safe.

I opened #409 to discuss this.

multinet/api.py
```diff
@@ -43,6 +43,17 @@ def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:  # noq
     return util.stream(tables)
 
 
+@bp.route("/workspaces/<workspace>/tables", methods=["POST"])
+@use_kwargs({"table": fields.Str()})
+@swag_from("swagger/workspace_aql_tables.yaml")
+def create_aql_table(workspace: str, table: str) -> Any:
+    """Retrieve the tables of a single workspace."""
```
waxlamp, line 50: This docstring is wrong.

devops/arangodb.yml
```diff
@@ -9,6 +9,12 @@
     - name: arangodb_root_password
       prompt: "Password for ArangoDB root account"
 
+    - name: arango_readonly_username
+      prompt: "Username for ArangoDB readonly user"
+
```
waxlamp, line 14: Mentioned elsewhere but we should be able to cut this bit out.

devops/arangodb.yml
```diff
@@ -46,6 +52,22 @@
...
+    - name: Ensure there is a read-only user
+      command: arangosh --server.password {{ arangodb_root_password }} --server.endpoint ssl://127.0.0.1:8529 --javascript.execute /tmp/create-readonly-user.js
+      become: true
+
+    - name: Delete templated file
+      file:
+        path: /tmp/create-readonly-user.js
+        state: absent
+
```
waxlamp, line 70: Woo hoo!

multinet/db.py
AlmightyYakob: > Shouldn't this be called with readonly = True?

Probably, I wasn't sure if we were going to enforce readonly on general aql queries. If that's the case, I can add it.


> It makes me wonder if we should flip the polarity of the default value--make `get_workspace_db()` readonly by default, and force us to call it with a flag if we want a writable DB. I know that sounds obnoxious, but it's definitely more fail-safe.
> 
> I opened #409 to discuss this.

I'm generally in favor :+1:

waxlamp: > > Shouldn't this be called with readonly = True?
> 
> Probably, I wasn't sure if we were going to enforce readonly on general aql queries. If that's the case, I can add it.
> 

Unless I'm missing something, I think we have to: otherwise, anyone with read access could write mutating AQL queries and completely screw up the workspace. The whole purpose of the readonly Arango user was to safely enable AQL queries, right?

devops/arangodb.yml
AlmightyYakob: :tada: I've also tested this against the testing database, and it seemed to work no problem.

multinet/db.py
AlmightyYakob: 16ac1d5

devops/arangodb.yml
AlmightyYakob: 16ac1d5

multinet/db.py
AlmightyYakob: True, I'll make the change

multinet/db.py
AlmightyYakob: Lmk what you think of fd8f5be

mypy_stubs/arango/aql.pyi
```diff
@@ -0,0 +1,40 @@
...
+        profile: Optional[bool] = None,
+        max_transaction_size: Optional[int] = None,
+        max_warning_count: Optional[int] = None,
+        intermediate_commit_count: Optional[int] = None,
+        intermediate_commit_size: Optional[int] = None,
+        satellite_sync_wait: Optional[Union[int, float]] = None,
+        read_collections: Optional[str] = None,
+        write_collections: Optional[str] = None,
+        stream: Optional[bool] = None,
+        skip_inaccessible_cols: Optional[bool] = None,
```
waxlamp, line 39: I think this is fine, but in general we really only need to add the signatures that we're specifically using in the codebase.

AlmightyYakob: Yeah good point, not sure why I stubbed out everything, but we've got it now :man_shrugging:

multinet/api.py
```diff
@@ -43,6 +43,17 @@ def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:  # noq
...
 
+@bp.route("/workspaces/<workspace>/tables", methods=["POST"])
+@use_kwargs({"table": fields.Str()})
+@swag_from("swagger/workspace_aql_tables.yaml")
+def create_aql_table(workspace: str, table: str) -> Any:
+    """Create a table from an AQL query."""
+    aql = request.data.decode()
+    table = db.create_aql_table(workspace, table, aql)
+
+    return (table, "200 OK")
```
JackWilb, line 54: To be a little more consistent with the other routes
```suggestion
    return table
```

multinet/db.py
```diff
@@ -300,6 +325,26 @@ def workspace_table_keys(
     return keys
 
 
+# TODO: Ensure current user has writer role
```
JackWilb, line 328: This should be on the API endpoint like the rest of them, not in the db file. It can likely be added with this PR if we wait until #412 lands or we can wait and add it after. If we wait, let's open an issue

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
+"""Tests for creating a table from an AQL query."""
+
+
+def test_malformed_aql(handled_workspace, server):
+    """Test that invalid/malformed AQL results in an error."""
+    table_name = "malformed_table"
+    malformed_aql = """FOR members RETURN member"""
+    malformed_aql_error = """unexpected RETURN declaration"""
```
JackWilb, line 8:
```suggestion
    malformed_aql = "FOR members RETURN member"
    malformed_aql_error = "unexpected RETURN declaration"
```

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
...
+    assert resp.status_code == 400
+    assert malformed_aql_error in resp.data.decode()
+
+
+def test_mutating_aql(populated_workspace, server):
+    """Test that an AQL query which updates documents fails."""
+    workspace, _, node_table, _ = populated_workspace
+
+    mutating_aql = f"""FOR thing in {node_table} UPDATE thing in {node_table}"""
+    mutating_aql_error = """AQL: read only"""
```
JackWilb, line 25:
```suggestion
    mutating_aql = f"FOR thing in {node_table} UPDATE thing in {node_table}"
    mutating_aql_error = "AQL: read only"
```

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert resp.status_code == 400
+    assert mutating_aql_error in resp.data.decode()
+
+
+def test_existing_table(populated_workspace, server):
+    """Test that attempt to create a table with an existing name fails."""
+    workspace, _, node_table, _ = populated_workspace
+
+    aql = f"""FOR thing in {node_table} RETURN thing"""
```
JackWilb, line 42:
```suggestion
    aql = f"FOR thing in {node_table} RETURN thing"
```

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert resp.status_code == 409
+    assert resp.data.decode() == node_table
+
+
+def test_create_node_table(populated_workspace, server):
+    """Test that creating a node table succeeds."""
+    workspace, _, node_table, edge_table = populated_workspace
+
+    aql = f"""FOR doc in {node_table} RETURN doc"""
```
JackWilb, line 58:
```suggestion
    aql = f"FOR doc in {node_table} RETURN doc"
```

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert resp.status_code == 200
+    assert resp.data.decode() == new_table_name
+
+
+def test_create_edge_table(populated_workspace, server):
+    """Test that creating an edge table succeeds."""
+    workspace, _, node_table, edge_table = populated_workspace
+
+    aql = f"""FOR doc in {edge_table} RETURN doc"""
```
JackWilb, line 75:
```suggestion
    aql = f"FOR doc in {edge_table} RETURN doc"
```

test/test_aql_table_creation.py
```diff
@@ -0,0 +1,104 @@
...
+
+    assert resp.status_code == 200
+    assert resp.data.decode() == new_table_name
+
+
+def test_unsupported_table(populated_workspace, server):
+    """Test that creating a non edge/node table results in an error."""
+    workspace, _, node_table, edge_table = populated_workspace
+
+    aql = f"""FOR doc in {edge_table} RETURN {{ name: 'something' }}"""
```
JackWilb, line 92:
```suggestion
    aql = f"FOR doc in {edge_table} RETURN {{ name: 'something' }}"
```

multinet/db.py
AlmightyYakob: I think it makes the most sense to address this in a separate PR, after #412 lands. I'll create an issue for it.

AlmightyYakob: #416

devops/arangodb.yml
```diff
@@ -111,3 +95,19 @@
         name: arangodb3
         state: restarted
       become: true
+
+    - name: Template file to server
```
waxlamp, line 99: Instead of such a generic description, this should say what is exactly happening (remember this is what we end up seeing when we run provisioning). Maybe:

```suggestion
    - name: Install arangosh script for user provisioning to server
```

devops/arangodb.yml
```diff
@@ -111,3 +95,19 @@
...
+        src: create-readonly-user.js
+        dest: /tmp/create-readonly-user.js
+        owner: ubuntu
+        mode: 444
+
+    - name: Ensure there is a read-only user
+      command: arangosh --server.password {{ arangodb_root_password }} --server.endpoint ssl://127.0.0.1:8529 --javascript.execute /tmp/create-readonly-user.js
+      become: true
+
+    - name: Delete templated file
```
waxlamp, line 110: Same comment as above. Maybe:

```suggestion
    - name: Delete user provisioning script
```

PR #421, opened 2020-07-07
Change order of checking edge vs node table

### LINKED ISSUE #397: Allow specification of required network properties
This is something I discovered when talking to @gotdairyya around a problem she was having importing data. She was trying to import an edge table that had the `_key` column, as well as the required `_from` and `_to` columns. Since we check if something is a node table first, it was seeing this and incorrectly inferring that it was a node table. For this specific issue, switching the order of these checks would probably be sufficient, but it demonstrates a larger point about our current data import mechanisms:

1. We currently don't have a way to specify the type of a table (node vs edge) when uploading a CSV file. This is supposed to be inferred by our backend, but we don't currently do a good job of that (as described above), and there could be residual edge cases there. I don't think it would be a bad idea to allow this.

2. We don't allow specifying the keys of the "source", "target", or "key" columns when uploading an edge table (or a graph). A decent amount of data shaping may be required for uploads to our backend to succeed.

I believe we have other issues that reference these same limitation, but the specific points I've laid out here seem like a good first step to address them, without dealing with too many edge cases. 

I'm curious to hear what @waxlamp and @JackWilb have to say regarding this. @gotdairyya if I represented the problem incorrectly, or there's things I've left out, please correct me.

AlmightyYakob:
Fixes #397 

This will conflict with #412, so once that's completed/merged, I can rebase this branch.

### DIFF
--- a/test/data/membership_with_keys.csv
+++ b/test/data/membership_with_keys.csv
@@ -1,0 +1,4 @@
+_key,_to,_from
+1783831,clubs/4,members/260
+1783830,clubs/2,members/260
+1783827,clubs/2,members/258
--- a/multinet/uploaders/csv.py
+++ b/multinet/uploaders/csv.py
@@ -125,10 +125,10 @@ def validate_csv(
     if not rows:
         raise ValidationFailed([MissingBody()])
 
-    if is_node_table(rows, key_field):
-        validate_node_table(rows, key_field, overwrite)
-    elif is_edge_table(rows):
+    if is_edge_table(rows):
         validate_edge_table(rows)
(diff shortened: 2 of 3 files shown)


test/data/membership_with_keys.csv
```diff
@@ -0,0 +1,320 @@
+_key,_to,_from
```
waxlamp, line 1: Can this file be made shorter? Or alternatively, is there a reason / lack of harm in leaving it this long?

AlmightyYakob: Good point, there's no real need for it to be this long. I'll truncate it.

PR #422, opened 2020-07-07
User Search

### LINKED ISSUE #420: Add user search
Prerequisite functionality for multinet-app/multinet-client#107

Could be something like `GET /api/user/search`, which would perform a prefix search on the `users` collection. The fields to search which initially seem the most useful are email and name. We might want a way to specify the field to search as a query param on the endpoint, returning a `4XX` error if the specified field doesn't exist, or isn't allowed to be searched on.

jjnesbitt:
Fixes multinet-app/multinet-server#420 

~~TODO:~~
~~- [ ] Add tests~~
multinet-app/multinet-api#123


multinet/user.py
```diff
@@ -137,3 +138,20 @@ def filtered_user(user: User) -> FilteredUser:
...
+
+    coll = user_collection()
+    aql = read_only_db("_system").aql
+
+    query = f"""
+        FOR doc in {coll.name}
+          FILTER CONTAINS(LOWER(doc.name), LOWER('{query}'))
+            OR CONTAINS(LOWER(doc.email), LOWER('{query}'))
+
+          RETURN doc
```
JackWilb, line 154: Returning just the doc without a limit might be an issue. If you just search for `gmail`, you'll get a ton of users, which might crush the server (once we have enough users)

JackWilb (changes requested):

JackWilb:
I did verify that the functionality does work from a local development environment, although it can't be tested on the deploy preview

multinet/auth/swagger/user/search.yaml
```diff
@@ -0,0 +1,17 @@
...
+consumes:
+    - text/plain
+parameters:
+    - name: query
+      description: Query to search for users with
+      in: query
+
+responses:
+  200:
+    description: List of matching users
```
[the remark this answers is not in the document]
waxlamp: The simpler description is fine. A "list of matching" users when there are no matches is, by definition, an empty list.

multinet/user.py
waxlamp: Agreed. Putting in a limit of 50 should be good for now.

multinet/user.py
```diff
@@ -137,3 +138,20 @@ def filtered_user(user: User) -> FilteredUser:
...
+    """
```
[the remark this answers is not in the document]
waxlamp: > It might be that this is an enhancement, too, and that we could open a low priority issue for this

Let's pursue this as an enhancement. @JackWilb, can you file an issue and link it back to this comment thread?

waxlamp (changes requested):
Jack's idea of limiting the user results is a good one.

Beyond that, you should look into how an index is created for the user collection on the name and email fields, and what needs to be done to maintain that index (e.g., when a new user is added or deleted). Could you drop a doc in our shared Multinet drive folder documenting what you find out about this? This bit can be done as a followup.

multinet/user.py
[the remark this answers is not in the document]
JackWilb: Opened #423

AlmightyYakob:
> Jack's idea of limiting the user results is a good one.
> 
> Beyond that, you should look into how an index is created for the user collection on the name and email fields, and what needs to be done to maintain that index (e.g., when a new user is added or deleted). Could you drop a doc in our shared Multinet drive folder documenting what you find out about this? This bit can be done as a followup.

I intentionally held off adding indexes to this, as the specific index that would be used here (a [fulltext index](https://www.arangodb.com/docs/3.6/appendix-glossary.html#fulltext-index)) would limit/change the search mode a bit. I'll file an issue that we can follow up on.

waxlamp (dismissed):

multinet/user.py
```diff
@@ -137,3 +138,21 @@ def filtered_user(user: User) -> FilteredUser:
...
+def search_user(query: str) -> Cursor:
+    """Search for users given a partial string."""
+
+    coll = user_collection()
+    aql = read_only_db("_system").aql
+
+    query = f"""
+        FOR doc in {coll.name}
+          FILTER CONTAINS(LOWER(doc.name), LOWER('{query}'))
+            OR CONTAINS(LOWER(doc.email), LOWER('{query}'))
```
JackWilb, line 152: Can we sanitize these parameters? For example, the following query string could be a problem, `'); <query to get user sessions>; --`

JackWilb: Looks like [this](https://github.com/ArangoDB-Community/pyArango#queries--aql) is how you do it

AlmightyYakob: Good point, we definitely should sanitize this input (and all input to AQL queries for that matter). It does seem like using bind vars is preferred, instead of injecting the values with f-strings. I can try that out.

waxlamp: Good catch, @JackWilb.

I don't think we need to sanitize *all* AQL input, since that's why we have the readonly user. But any time we're constructing an AQL query expression based on user input, we should definitely do it via bind vars or whatever the recommended secure method is.

waxlamp: The other thought I had was, we could also just replace this AQL search with a brute force linear search as well. But since this is already here and we have a path to a secure usage of it, we can press forward.

JackWilb: Agreed. Your point about the readonly user definitely helps us in cases where an attacker would try to delete/modify data, but not in the cases where an attacker would want to read sensitive information (such as session ids, other collections, etc.). Sanitizing all queries where there are user inputs does solve that problem.

AlmightyYakob: I've updated user search to use `bind_vars` (f116cd2), and created an issue to complete doing this throughout the codebase (#440).

JackWilb (approved):

multinet/user.py
```diff
@@ -137,3 +138,22 @@ def filtered_user(user: User) -> FilteredUser:
...
     return from_dict(User, dataclasses.asdict(user))
+
+
+def search_user(query: str) -> Cursor:
+    """Search for users given a partial string."""
+
+    coll = user_collection()
+    aql = read_only_db("_system").aql
+
+    bind_vars = {"@users": coll.name, "query": query}
```
waxlamp, line 149: What does the `@` do?

multinet/user.py
```diff
@@ -137,3 +138,22 @@ def filtered_user(user: User) -> FilteredUser:
...
+    query = """
+        FOR doc in @@users
```
waxlamp, line 151: Why `@@` here?

waxlamp (approved):

multinet/user.py
AlmightyYakob: It's just the required syntax for bind vars https://www.arangodb.com/docs/stable/aql/fundamentals-bind-parameters.html

multinet/user.py
AlmightyYakob: When referencing collections specifically, this syntax is needed https://www.arangodb.com/docs/stable/aql/fundamentals-bind-parameters.html

PR #430, opened 2020-07-15
Add endpoints for getting/setting workspace permissions

### LINKED ISSUE #419: Add endpoints for getting and setting workspace user roles
Prerequisite functionality for https://github.com/multinet-app/multinet-client/pull/107

This would be a pair of endpoints that would look something like:

* `GET /api/workspaces/{workspace}/permissions`
* `PUT /api/workspaces/{workspace}/permissions`

The data model for each of these would contain a list of users, and their permission on the workspace. Something like:

```
[
  {
    user: <sub>
    role: <'reader' | 'writer' | 'maintainer' | 'owner'>
  },
  ...
]
```

The workflow for modifying the permissions of a workspace would be as follows:

1. Retrieve list of users with their respective roles
2. Modify list
3. Call `PUT` to set the new permissions

This would obviously be gated behind the `owner` role on a workspace.

jjnesbitt:
Fixes multinet-app/multinet-server#419 
Depends on multinet-app/multinet-server#436 

This adds endpoints to retrieve and set the permissions on a workspace. It, however, does not yet allow for the changing of the owner of a workspace, as that is intended to be a separate action. It also adds a temporary transformation of the underlying permission data (f20bc1f), which is needed for the client to function properly. This temporary fix is intended to be rolled back once the [ORM](https://github.com/multinet-app/multinet-rfcs/pull/12) is implemented, as that's a natural point to change how we're storing this data.


.github/workflows/main.yml
```diff
@@ -32,11 +32,15 @@ jobs:
       - name: Check generated requirements against existing requirements
         run: diff requirements.txt generated-requirements.txt
 
+      - name: Capture pipenv version
+        run: echo "::set-output name=pipenv-version::$(pipenv --version | sed 's/,//g')"
```
AlmightyYakob, line 36: First off, thanks for doing this.

I think if we're gonna bother using `sed`, why not just remove all of the output that isn't the version anyway.

```suggestion
        run: echo "::set-output name=pipenv-version::$(pipenv --version | sed 's/pipenv, version //g')"
```

Also, why is the `::set-output` needed here?

AlmightyYakob: Also, I think once everything here is resolved, we can probably squash these two commits into one.

waxlamp: The reason I just removed the comma was that, if there's any new release of Pipenv at all and the version string changes, I want the cache to be invalidated. I would have left the comma there except that GH Actions doesn't allow them in cache hash strings. It is probably low-risk to do it your way, but I don't see any real advantages to it. As I said, my preference was to just leave the version string raw, but the comma gums that up. What do you think?

That `::set-output...::` thing is a GH Actions "workflow command". You can inject certain commands through an `echo` invocation to make certain things happen. `set-output` causes whatever comes after the "closing" `::` to appear in the `steps.<id>.outputs.<output-name>` global property, and it becomes available in subsequent steps. Here I used that to inject the version string into the cache hash value. Here's the documentation for [workflow commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions). Very cool stuff!

waxlamp: > Also, I think once everything here is resolved, we can probably squash these two commits into one.

Agreed. I meant to mention this in a comment but forgot. When you're happy with everything, feel free to do an interactive rebase to squash those commits as needed (or, if you're feeling very pedantic, you could cherry pick the combined commit to another PR, then rebase this on top of that--but I don't really think that's necessary).

AlmightyYakob: > The reason I just removed the comma was that, if there's any new release of Pipenv at all and the version string changes, I want the cache to be invalidated. I would have left the comma there except that GH Actions doesn't allow them in cache hash strings. It is probably low-risk to do it your way, but I don't see any real advantages to it. As I said, my preference was to just leave the version string raw, but the comma gums that up. What do you think?

The main reason I suggested it is because currently there's still a space in the cache string. It's not a huge deal, and I guess is allowed, but that was the main reason.

> That `::set-output...::` thing is a GH Actions "workflow command". You can inject certain commands through an `echo` invocation to make certain things happen. `set-output` causes whatever comes after the "closing" `::` to appear in the `steps.<id>.outputs.<output-name>` global property, and it becomes available in subsequent steps. Here I used that to inject the version string into the cache hash value. Here's the documentation for [workflow commands](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions). Very cool stuff!

:+1:

waxlamp: > > The reason I just removed the comma was that, if there's any new release of Pipenv at all and the version string changes, I want the cache to be invalidated. I would have left the comma there except that GH Actions doesn't allow them in cache hash strings. It is probably low-risk to do it your way, but I don't see any real advantages to it. As I said, my preference was to just leave the version string raw, but the comma gums that up. What do you think?
> 
> The main reason I suggested it is because currently there's still a space in the cache string. It's not a huge deal, and I guess is allowed, but that was the main reason.
> 

Since spaces are allowed, I am not worried about them remaining in the string. I'm more worried about a future version  update *changing* the format of the version output (and thus making our postprocessing logic broken somehow), so I made the minimal change to the string to get it to work (i.e., zapping the commas).

multinet/api.py
```diff
@@ -25,11 +27,62 @@
...
     RequiredParamsMissing,
 )
-from multinet.user import current_user
+from multinet.user import current_user, find_user_from_id
 
 bp = Blueprint("multinet", __name__)
 
 
+# Included here due to circular imports
+# TODO: Remove once implementing new ORM and permission storage
```
waxlamp, line 36: Is your plan to finish the ORM work first, then rebase on (or merge from) that work here?

AlmightyYakob: I figured that this would make it in first, so the comment was there to leave a breadcrumb for circling back. Do you think it's worth waiting for the ORM work to make it in first? Managing permissions from the client relies on this endpoint.

waxlamp: If the circular dependency is causing test failures, I'm not sure that we should simply merge it in. If there's an easy way to fix that issue, then we should do it. If that will cause a major headache then we may need to go out on a limb and merge in "broken" code but I'd like to avoid that if possible.

What do you think?

AlmightyYakob: The main reason I didn't solve the circular dependency issue in a more robust way is because it seemed a bit out of scope for this issue, and the functions that are causing them to happen (these conversion functions) are temporary and would be removed when the ORM change comes down. Also, if we encounter this same issue during the ORM refactor, that's a perfectly valid time to fix it. I just didn't want to spend a ton of time reorganizing the code base just for it not to matter once we ultimately reorganize it for the ORM.

waxlamp: I agree with what you're saying. But this will not work if the server can't actually run because of the circular import. If you can certify that this does work properly, then we're all good. I would just ask you to file an issue pointing to these code blobs (a TODO in the codebase isn't quite as trackable as I'd like).

AlmightyYakob: > I agree with what you're saying. But this will not work if the server can't actually run because of the circular import. If you can certify that this does work properly, then we're all good.

This does work properly (as in, no circular imports at runtime), although having someone else test this out may be a good idea.

> I would just ask you to file an issue pointing to these code blobs (a TODO in the codebase isn't quite as trackable as I'd like).

#442

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+def _permissions_id_to_user(permissons: WorkspacePermissions) -> Dict:
+    """Transform permission documents to directly container user info."""
```
waxlamp, line 38: The grammar here is a bit off. Should it be `contain`?

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+
+    new_permissions = deepcopy(permissons)
+    for role, users in new_permissions.items():
+        if role == "public":
+            continue
+
+        if role == "owner":
+            assert isinstance(users, str)
```
waxlamp, line 46: We shouldn't have any `assert` statements in the main code. What role is it serving here?

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+
+            user = find_user_from_id(users)
+            if user is not None:
+                # Not sure why mypy complains about `user` here
```
waxlamp, line 50: What is the exact type error you get?

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+                new_permissions["owner"] = asdict(user)  # type: ignore
```
waxlamp, line 51: What's this actually doing? It's placing the full user object in place of the sub string?

multinet/api.py
waxlamp: Ah I think I might understand. Is this something `mypy` makes use of to narrow from Union types?

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+        else:
+            assert isinstance(users, list)
+
+            new_users = []
+            for sub in users:
+                user = find_user_from_id(sub)
```
waxlamp, line 57: Does this make the loop O(N^2)? Is that ok?

multinet/api.py
waxlamp: I think a longer docstring would be useful here. Reading the source code is somewhat confusing due to the types shifting and the structure of our basic data changing. A paragraph following the summary comment you have here would be great because newcomers to the codebase (in this case, me!) would better understand what this utility function is for, and why it's written in this way, etc.

multinet/api.py
```diff
@@ -24,11 +27,62 @@
...
+                if user is not None:
+                    new_users.append(asdict(user))
+
+            # Ignoring requirement of literal string keys
+            new_permissions[role] = new_users  # type: ignore
+
+    return cast(Dict, new_permissions)
```
waxlamp, line 64: Is `WorkspacePermissions` a dataclass? How does this `cast` operator work to turn that into a `Dict`?

multinet/api.py
```diff
@@ -41,14 +95,32 @@ def get_workspaces() -> Any:
...
-    """Retrieve a single workspace."""
+@bp.route("/workspaces/<workspace>/permissions", methods=["GET"])
+@require_maintainer
+@swag_from("swagger/get_workspace_permissions.yaml")
+def get_workspace_permissions(workspace: str) -> Any:
+    """Retrieve the permissions of a workspace."""
     metadata = db.get_workspace_metadata(workspace)
 
-    return {"name": metadata["name"], "permissions": metadata["permissions"]}
+    return _permissions_id_to_user(metadata["permissions"])
```
waxlamp, line 105: This seems to be the only call to `_permissions_id_to_user`. Why not inline the function here?

multinet/api.py
```diff
@@ -41,14 +95,32 @@ def get_workspaces() -> Any:
...
+    if set(request.json.keys()) != {
+        "owner",
+        "maintainers",
+        "writers",
+        "readers",
+        "public",
+    }:
+        raise MalformedRequestBody(request.json)
+
+    perms = _permissions_user_to_id(request.json)
```
waxlamp, line 122: Same question as above about inlining a one-time-use function.

multinet/api.py
```diff
@@ -41,14 +95,32 @@ def get_workspaces() -> Any:
...
+def set_workspace_permissions(workspace: str) -> Any:
+    """Set the permissions on a workspace."""
+    if set(request.json.keys()) != {
+        "owner",
+        "maintainers",
+        "writers",
+        "readers",
+        "public",
+    }:
+        raise MalformedRequestBody(request.json)
```
waxlamp, line 120: I think I might disagree with this. A `PUT` call usually means that you specify only what you want changed, and the server figures out how to apply your "diff".

This way of operating also works, but there's a danger: suppose you `GET` the permissions, then you have faulty logic for updating that object so that you accidentally put in changes you didn't want. Then on `PUT`, you will end up with bad state.

This is probably a weak argument. I'm willing to hear why it should be required to pass back the whole permissions object. One argument that feels at least somewhat compelling is that the server logic could become more complex to handle partial updates.

In any case, there's no need to change things in this PR. I'd like to have the discussion and if we decide that partial `PUT`s are warranted, we'll just file an issue and worry about it later.

multinet/db.py
```diff
@@ -222,13 +229,38 @@ def get_workspace_metadata(name: str) -> Workspace:
 
     # Find the metadata record for the named workspace. If it's not there,
     # something went very wrong, so bail out.
-    metadata = workspace_mapping(name)
+    metadata = workspace_mapping.__wrapped__(name)
```
waxlamp, line 232: Remind me what `.__wrapped__()` does here? It just calls the undecorated version of the function?

Why do we need that here? If you include a comment explaining the need, that would be good.

(Hopefully this becomes moot after the ORM stuff lands but still, in the meantime we should cover our donkeys.)

multinet/db.py
```diff
@@ -222,13 +229,38 @@ def get_workspace_metadata(name: str) -> Workspace:
...
+def set_workspace_permissions(
+    name: str, permissions: WorkspacePermissions
+) -> WorkspacePermissions:
+    """Update the permissions for a given workspace."""
+    if not workspace_exists(name):
+        raise WorkspaceNotFound(name)
+
+    doc = copy.deepcopy(workspace_mapping(name))
+    if doc is None:
+        raise DatabaseCorrupted()
```
waxlamp, line 248: We should think about including more information in the `DatabaseCorrupted` exception. Even just an error message.

I'll file an issue for this.

multinet/db.py
```diff
@@ -222,13 +229,38 @@ def get_workspace_metadata(name: str) -> Workspace:
...
+
+    # TODO: Do user object validation once ORM is implemented
```
waxlamp, line 250: Can you make sure to file an issue for this TODO?

multinet/db.py
```diff
@@ -222,13 +229,38 @@ def get_workspace_metadata(name: str) -> Workspace:
...
+
+    # To ensure owner is only changed explicitly
```
waxlamp, line 252: What do you mean by this?

multinet/db.py
```diff
@@ -222,13 +229,38 @@ def get_workspace_metadata(name: str) -> Workspace:
...
+    new_permissions = copy.deepcopy(permissions)
+    new_permissions["owner"] = doc["permissions"]["owner"]
+
+    doc["permissions"] = new_permissions
+    return_doc = workspace_mapping_collection().get(
+        workspace_mapping_collection().update(doc, check_rev=False)
```
waxlamp, line 258: Why `check_rev=False`?

waxlamp: Also, what does the `update()` method return? I'm surprised it doesn't give you back the new doc!

multinet/swagger/get_workspace_permissions.yaml
```diff
@@ -1,4 +1,4 @@
-Retrieve a single workspace.
+Retrieve the permissions on a workspace.
```
waxlamp, line 1: Doesn't more in this file need to be updated?

multinet/swagger/set_workspace_permissions.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - $ref: "#/parameters/workspace"
+  - name: permissions
+    description: Permission Document
+    in: body
+
+responses:
+  200:
+    description: Metadata about the specified workspace
+    schema:
+      $ref: "#/definitions/workspace"
```
waxlamp, line 13: If `#/definitions/workspace` now details workspace permissions, its name should probably change to reflect that.

multinet/types.py
```diff
@@ -5,16 +5,15 @@
...
-    {
-        "owner": str,  # this is the `sub` property of the `User` dataclass
-        "maintainers": List[str],
-        "writers": List[str],
-        "readers": List[str],
-        "public": bool,
-    },
-)
+
+class WorkspacePermissions(TypedDict):
```
waxlamp, line 9: Why still inherit from `TypedDict`? I thought you were on a mission to exterminate them?

multinet/types.py
```diff
@@ -5,16 +5,15 @@
...
+    """Permissions on a Workspace."""
+
+    owner: str  # this is the `sub` property of the `User` dataclass
```
waxlamp, line 12: I wonder if we should just declare a type alias, `UserSub` that equals `str`. Just something to think about. Maybe it's more trouble than it's worth.

test/test_meta.py
```diff
@@ -7,7 +7,7 @@ def test_generated_workspace(managed_workspace, managed_user, server):
     """Test that a generated workspace exists when querying the API."""
 
     with conftest.login(managed_user, server):
-        resp = server.get(f"/api/workspaces/{managed_workspace}")
+        resp = server.get(f"/api/workspaces/{managed_workspace}/permissions")
```
waxlamp, line 10: Shouldn't we check that the permissions are what we expect?

Perhaps we need a bit more complex fixtures to make this happen easily. I have some ideas.

It's ok to not address this now, but please file an issue instead.

waxlamp: Oh I see, you ran that test below. n/m.

waxlamp (changes requested):
This looks great, thanks!

I left a whole bunch of questions and a couple of minor suggestions.

multinet/api.py
AlmightyYakob: Yeah, it's to deal with `mypy`. Since all other fields besides `public` are arrays, we have to tell `mypy` that in this case, it's just a string, since `mypy` is too stupid to figure that out on its own.

multinet/api.py
AlmightyYakob: Exactly

multinet/api.py
AlmightyYakob: It was giving me an error regarding user being a `str` or something. However I'm actually about to push a change which will fix the type errors, so this will be moot.

multinet/api.py
AlmightyYakob: I don't think it's a problem, since this it's really O(n*m), where n is the number of keys in the permissions dict (5), and m is the number of users in each role (could just use the role with the most users, to be safe). So it's really linear.

multinet/api.py
AlmightyYakob: It's a `TypedDict`, and I'm not sure how the `cast` operator works internally, but it's used here to remove the restrictions of it being a `WorkspacePermissions` dict, because it isn't anymore, due to the replacements above. I'm actually about to push a change where I move this operation, but it'll still be in this function.

multinet/api.py
AlmightyYakob: I just preferred to keep those temporary functions clearly separated and marked from the rest of the api. Since eventually, we'll remove those functions, but the api endpoint will remain.

multinet/api.py
AlmightyYakob: Same response as above

multinet/api.py
AlmightyYakob: > I think I might disagree with this. A `PUT` call usually means that you specify only what you want changed, and the server figures out how to apply your "diff".

I disagree, I think of `PUT` as an updating an existing resource to match what you specify (more or less). You seem to be describing the `PATCH` method.


> This way of operating also works, but there's a danger: suppose you `GET` the permissions, then you have faulty logic for updating that object so that you accidentally put in changes you didn't want. Then on `PUT`, you will end up with bad state.

I think this could happen in any way of utilizing the endpoint. At the end of the day, it's our responsibility to make sure the client uses this endpoint properly. Plus, you still can't modify owners from these endpoints, so in the worst case the owner can still fix things after an error.

> This is probably a weak argument. I'm willing to hear why it should be required to pass back the whole permissions object. One argument that feels at least somewhat compelling is that the server logic could become more complex to handle partial updates.

Yeah this is one of the main reasons, and I don't really see a benefit to specifying only the change rather than the whole new object. It makes the logic on the server more complex, and it also makes the logic on the client much more complex.

multinet/db.py
waxlamp: See #443.

multinet/api.py
waxlamp: You are right, especially in pointing out that I'm thinking of `PATCH` rather than `PUT`. I agree with everything you said here.

multinet/types.py
AlmightyYakob: I am still on a mission, I just changed the declaration of `WorkspacePermissions` to use the class based syntax, as it's more clear.

multinet/types.py
AlmightyYakob: Perhaps, but I think the with the new ORM changes it would become moot.

multinet/db.py
AlmightyYakob: I mean that the owner can't be changed through setting the workspace permissions like everything else, it needs to be explicitly changed through it's own endpoint or something.

multinet/db.py
AlmightyYakob: Yeah, `update` only returns the metadata (so `_rev`, `_id`, etc.). The reason I added `check_rev=False`, is because I was getting a `DocumentRevisionError`, which was due to the call to `workspace_mapping` above returning an old version of the document, with an old `_rev` (since it's cached). 

Looking at this now, I actually don't like what I did, as it just ignores the problem. I'm going to push a fix that actually addresses the problem here.

multinet/db.py
waxlamp: Oh I see. I'd suggest different wording for the comment:

```suggestion
    # Disallow changing workspace ownership through this function.
```

test/test_meta.py
AlmightyYakob: Yeah, this just updates the endpoint with the new URL.

multinet/db.py
waxlamp: > Looking at this now, I actually don't like what I did, as it just ignores the problem. I'm going to push a fix that actually addresses the problem here.

That's why I spoke up, something felt off about setting that flag.

multinet/db.py
AlmightyYakob: Yeah it calls the un-decorated function. The reason I added that is because when calling `get_workspace_metadata`, we want the most recent version of the metadata, and this function itself isn't cached, so we couldn't just call the un-decorated version of `get_workspace_metadata`, as it doesn't exist. With the changes that I said I'd be making in my other comment, this might actually not be necessary anymore.

I'd still be okay with using the un-cached function, as long as it doesn't impact performance. What do you think?

multinet/db.py
AlmightyYakob: 6a786b4

multinet/api.py
AlmightyYakob: c60d404

multinet/db.py
AlmightyYakob: https://github.com/multinet-app/multinet-server/issues/444

multinet/db.py
waxlamp: Correctness matters more than performance in this case, so do whatever makes the most sense. I'd just include a comment about why it's this way, just to spare confusion or errors in the future.

multinet/swagger/set_workspace_permissions.yaml
AlmightyYakob: 98a0c81

multinet/swagger/get_workspace_permissions.yaml
AlmightyYakob: 98a0c81

multinet/db.py
AlmightyYakob: 2b31899 I undid the use of the wrapped function, so I didn't think a comment was necessary anymore

waxlamp (approved):

PR #435, opened 2020-07-23
Use backticks to protect hyphens in table names

### LINKED ISSUE #434: Hyphens in table names cause AQL failures
Table names may have the following characters a-z, A-Z, 0-9, -, _ as per the arango documentation [here](https://www.arangodb.com/docs/stable/data-modeling-naming-conventions-collection-and-view-names.html). However, when the table has a "-" in the name, arango can sometimes treat it as a subtraction of 2 elements (see [here](https://www.arangodb.com/docs/stable/aql/fundamentals-syntax.html#names) for details). 

That last documentation link suggests that tables names be enclosed in backticks to prevent the AQL interpreter considering the dashes to be arithmetic operators. We'll need to go through our codebase and replace the collection names with the original enclosed in backticks.

For reference, the table sparked this investigation is `preCovid-19_flights_edges` and the error that would be thrown when trying to run `RETURN LENGTH(preCovid-19_flights_edges)` was 
```
Query: AQL: syntax error, unexpected identifier, expecting ) near '_flights_edges)' at position 1:26 (while parsing)
```

The code that generated this error is [here](https://github.com/multinet-app/multinet-server/blob/master/multinet/db.py#L324-L329) in the db.py file and is likely where we need to fix this problem.

waxlamp:
I think I got all of them.

TODO:
- [ ] use bind vars (see https://github.com/multinet-app/multinet-server/issues/434#issuecomment-666523967)

Fixes #434
Depends on #436

### DIFF
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -313,7 +313,7 @@ def workspace_table_rows(
     """Stream the rows of a table in CSV form."""
 
     query = f"""
-    FOR d in {table}
+    FOR d in `{table}`
         LIMIT {offset}, {limit}
         RETURN d
     """
@@ -324,7 +324,7 @@ def workspace_table_rows(
 def workspace_table_row_count(workspace: str, table: str) -> int:
     """Return the number of rows in a table."""
     count_query = f"""
-    RETURN LENGTH({table})
+    RETURN LENGTH(`{table}`)
     """
     return next(aql_query(workspace, count_query))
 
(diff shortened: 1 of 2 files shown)


waxlamp:
Not sure of the best way to test this. Maybe by creating a table with a hyphen, then checking to see that getting its row count works. @AlmightyYakob can you give me some advice on this?

AlmightyYakob:
> Not sure of the best way to test this. Maybe by creating a table with a hyphen, then checking to see that getting its row count works. @AlmightyYakob can you give me some advice on this?

That's probably a good way to go about it. If it were to fail it would raise an error from here:
https://github.com/multinet-app/multinet-server/blob/ba7f385e7e5226be28b782fabf612adae579ca49/multinet/db.py#L456-L465

So ensuring that error doesn't get raised might work.

jjnesbitt (commented):
LGTM, I think after tests are added it should be good

PR #439, opened 2020-07-29
Add `methods` argument to all `@bp.route` decorators

mvandenburgh:


waxlamp:
@mvandenburgh, this PR looks just fine.

When you get around to switching this from draft status, could you also edit the title to remove the bit about iterators and lists?

PR #448, opened 2020-08-10
Multipart Upload Endpoints

mvandenburgh:
Adds three of the `/uploads` post endpoints from https://github.com/multinet-app/multinet-rfcs/pull/13.

### DIFF
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -43,7 +43,7 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
 
     # Set max file upload size to 2 MB
     # TODO: decide on a non-arbitrary limit
-    app.config["MAX_CONTENT_LENGTH"] = 2000000
+    app.config["MAX_CONTENT_LENGTH"] = 32 * 1024 * 1024
 
     app.secret_key = flask_secret_key()
 
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -41,8 +41,7 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     CORS(app, origins=allowed_origins, supports_credentials=True)
     Swagger(app, template_file="swagger/template.yaml")
 
-    # Set max file upload size to 2 MB
-    # TODO: decide on a non-arbitrary limit
+    # Set max file upload size to 32 MB
     app.config["MAX_CONTENT_LENGTH"] = 32 * 1024 * 1024
 
     app.secret_key = flask_secret_key()
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -41,6 +41,10 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     CORS(app, origins=allowed_origins, supports_credentials=True)
     Swagger(app, template_file="swagger/template.yaml")
 
+    # Set max file upload size to 2 MB
+    # TODO: decide on a non-arbitrary limit
+    app.config["MAX_CONTENT_LENGTH"] = 2000000
+
     app.secret_key = flask_secret_key()
 
     # Set up logging.
@@ -52,6 +56,8 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     app.register_blueprint(uploaders.newick.bp, url_prefix="/api/newick")
     app.register_blueprint(uploaders.nested_json.bp, url_prefix="/api/nested_json")
     app.register_blueprint(uploaders.d3_json.bp, url_prefix="/api/d3_json")
+
+    app.register_blueprint(uploaders.multipart_upload.bp, url_prefix="/api/uploads")
 
     app.register_blueprint(downloaders.csv.bp, url_prefix="/api")
     app.register_blueprint(downloaders.d3_json.bp, url_prefix="/api")
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -355,3 +355,10 @@ def chunk_upload(upload_id: str) -> Any:
     collection = db.db("_system").collection(upload_id)
     collection.insert({sequence: blob})
     return sequence
+
+
+@bp.route("/uploads/<upload_id>", methods=["DELETE"])
+def delete_upload_collection(upload_id: str) -> Any:
+    """Delete the database collection associated with the given upload_id."""
+    db.db("_system").delete_collection(upload_id)
+    return ''
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -6,6 +6,7 @@ from flask import Blueprint, request
 from webargs import fields
 from webargs.flaskparser import use_kwargs
 
+from uuid import uuid4
 from typing import Any, Optional, List, Dict, cast
 from multinet.types import EdgeDirection, TableType
 from multinet.auth.util import (
@@ -326,3 +327,26 @@ def delete_table(workspace: str, table: str) -> Any:
     """Delete a table."""
     db.delete_table(workspace, table)
     return table
+
+
+@bp.route("/uploads", methods=["POST"])
+def create_upload():
+    """Create a collection for multipart upload."""
+    upload_id = uuid4().hex
+    sysdb = db.db("_system")
+    sysdb.create_collection(upload_id)
+    return upload_id
+
+
+@bp.route("/uploads/<upload_id>/chunk", methods=["POST"])
+def chunk_upload(upload_id):
+    """Upload a chunk to the specified collection."""
+    sequence = request.args.get("sequence")
+    chunk = dict(request.files)["chunk"].read()
+
+    # convert bytes to array of ints since arrango doesn't support binary blobs
+    blob = list(chunk)
+
+    collection = db.db("_system").collection(upload_id)
+    collection.insert({sequence: blob})
+    return sequence
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -361,4 +361,4 @@ def chunk_upload(upload_id: str) -> Any:
 def delete_upload_collection(upload_id: str) -> Any:
     """Delete the database collection associated with the given upload_id."""
     db.db("_system").delete_collection(upload_id)
-    return ''
+    return ""
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -6,8 +6,6 @@ from flask import Blueprint, request
 from webargs import fields
 from webargs.flaskparser import use_kwargs
 
-from base64 import b64encode
-from uuid import uuid4
 from typing import Any, Optional, List, Dict, cast
 from multinet.types import EdgeDirection, TableType
 from multinet.auth.util import (
@@ -328,37 +326,3 @@ def delete_table(workspace: str, table: str) -> Any:
     """Delete a table."""
     db.delete_table(workspace, table)
     return table
-
-
-@bp.route("/uploads", methods=["POST"])
-def create_upload() -> str:
-    """Create a collection for multipart upload."""
-
-    sysdb = db.db("_system")
-
-    # note: arangodb collections can't start with a number
-    upload_id = f"upload_{uuid4().hex}"
-
-    sysdb.create_collection(upload_id)
-    return upload_id
-
-
-@bp.route("/uploads/<upload_id>/chunk", methods=["POST"])
-@use_kwargs({"sequence": fields.Str()})
-def chunk_upload(upload_id: str, sequence: str) -> Any:
-    """Upload a chunk to the specified collection."""
-    chunk = dict(request.files)["chunk"].read()
-
-    # convert bytes to base64 string since arango doesn't support binary blobs
-    blob = b64encode(chunk).decode("ascii")
-
-    collection = db.db("_system").collection(upload_id)
-    collection.insert({sequence: blob})
-    return sequence
-
-
-@bp.route("/uploads/<upload_id>", methods=["DELETE"])
-def delete_upload_collection(upload_id: str) -> Any:
-    """Delete the database collection associated with the given upload_id."""
-    db.db("_system").delete_collection(upload_id)
-    return ""
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -344,9 +344,9 @@ def create_upload() -> str:
 
 
 @bp.route("/uploads/<upload_id>/chunk", methods=["POST"])
-def chunk_upload(upload_id: str) -> Any:
+@use_kwargs({"sequence": fields.Str()})
+def chunk_upload(upload_id: str, sequence: str) -> Any:
     """Upload a chunk to the specified collection."""
-    sequence = request.args.get("sequence")
     chunk = dict(request.files)["chunk"].read()
 
     # convert bytes to base64 string since arango doesn't support binary blobs
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -6,6 +6,7 @@ from flask import Blueprint, request
 from webargs import fields
 from webargs.flaskparser import use_kwargs
 
+from base64 import b64encode
 from uuid import uuid4
 from typing import Any, Optional, List, Dict, cast
 from multinet.types import EdgeDirection, TableType
@@ -344,8 +345,8 @@ def chunk_upload(upload_id):
     sequence = request.args.get("sequence")
     chunk = dict(request.files)["chunk"].read()
 
-    # convert bytes to array of ints since arrango doesn't support binary blobs
-    blob = list(chunk)
+    # convert bytes to base64 string since arango doesn't support binary blobs
+    blob = b64encode(chunk).decode('ascii')
 
     collection = db.db("_system").collection(upload_id)
     collection.insert({sequence: blob})
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -331,22 +331,26 @@ def delete_table(workspace: str, table: str) -> Any:
 
 
 @bp.route("/uploads", methods=["POST"])
-def create_upload():
+def create_upload() -> str:
     """Create a collection for multipart upload."""
-    upload_id = uuid4().hex
+
     sysdb = db.db("_system")
+
+    # note: arangodb collections can't start with a number
+    upload_id = f"upload_{uuid4().hex}"
+
     sysdb.create_collection(upload_id)
     return upload_id
 
 
 @bp.route("/uploads/<upload_id>/chunk", methods=["POST"])
-def chunk_upload(upload_id):
+def chunk_upload(upload_id: str) -> Any:
     """Upload a chunk to the specified collection."""
     sequence = request.args.get("sequence")
     chunk = dict(request.files)["chunk"].read()
 
     # convert bytes to base64 string since arango doesn't support binary blobs
-    blob = b64encode(chunk).decode('ascii')
+    blob = b64encode(chunk).decode("ascii")
 
     collection = db.db("_system").collection(upload_id)
     collection.insert({sequence: blob})
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -624,7 +624,7 @@ def uploads_collection() -> StandardCollection:
 def create_upload_document() -> Any:
     """Insert empty multipart upload temp document."""
     # returns arango document key as upload_id for client
-    return uploads_collection().insert({"0": {}}).get("_key")
+    return uploads_collection().insert({}).get("_key")
 
 
 def insert_file_chunk(key: str, sequence: str, chunk: str) -> str:
--- a/multinet/db.py
+++ b/multinet/db.py
@@ -28,7 +28,7 @@ from multinet.types import (
     WorkspacePermissions,
 )
 from multinet.auth.types import User
-from multinet.errors import InternalServerError
+from multinet.errors import InternalServerError, NonExistantUploadDocument
 from multinet.validation.csv import validate_csv
 from multinet import util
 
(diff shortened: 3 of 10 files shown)


multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+
+
+@bp.route("/uploads", methods=["POST"])
+def create_upload() -> str:
+    """Create a collection for multipart upload."""
+
+    sysdb = db.db("_system")
+
+    # note: arangodb collections can't start with a number
+    upload_id = f"upload_{uuid4().hex}"
```
waxlamp, line 340: For workspace database names, we've been using `w-` followed by a UUID generated value. We could probably go with `u-` for these.

multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+
+    sysdb.create_collection(upload_id)
+    return upload_id
+
+
+@bp.route("/uploads/<upload_id>/chunk", methods=["POST"])
+@use_kwargs({"sequence": fields.Str()})
+def chunk_upload(upload_id: str, sequence: str) -> Any:
+    """Upload a chunk to the specified collection."""
+    chunk = dict(request.files)["chunk"].read()
```
waxlamp, line 350: What if there are no files associated with the request?

waxlamp: Also, do you know if there's a way to read out the file size first, and abort with an error if it's above a certain limit? Otherwise this could be used to DoS the server.

multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+
+    # convert bytes to base64 string since arango doesn't support binary blobs
+    blob = b64encode(chunk).decode("ascii")
+
+    collection = db.db("_system").collection(upload_id)
+    collection.insert({sequence: blob})
```
waxlamp, line 356: We generally want to keep database access limited to the DB layer.

Can you refactor this into a function in that file that performs this action?

multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+@bp.route("/uploads", methods=["POST"])
+def create_upload() -> str:
+    """Create a collection for multipart upload."""
+
+    sysdb = db.db("_system")
+
+    # note: arangodb collections can't start with a number
+    upload_id = f"upload_{uuid4().hex}"
+
+    sysdb.create_collection(upload_id)
```
waxlamp, line 342: See the below comment about putting these operations in the DB layer.

multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+    collection = db.db("_system").collection(upload_id)
+    collection.insert({sequence: blob})
+    return sequence
+
+
+@bp.route("/uploads/<upload_id>", methods=["DELETE"])
+def delete_upload_collection(upload_id: str) -> Any:
+    """Delete the database collection associated with the given upload_id."""
+    db.db("_system").delete_collection(upload_id)
+    return ""
```
waxlamp, line 364: Is it required to return something? What do we do in other instances of DELETE calls?

multinet/api.py
```diff
@@ -326,3 +328,37 @@ def delete_table(workspace: str, table: str) -> Any:
...
+
+    collection = db.db("_system").collection(upload_id)
+    collection.insert({sequence: blob})
+    return sequence
+
+
+@bp.route("/uploads/<upload_id>", methods=["DELETE"])
+def delete_upload_collection(upload_id: str) -> Any:
+    """Delete the database collection associated with the given upload_id."""
+    db.db("_system").delete_collection(upload_id)
```
waxlamp, line 363: It seems silly in this case especially, but even this one-liner ought to be indirected through db.py.

multinet/api.py
waxlamp, line 342: I think rather than put these collections at the top level of `sysdb`, there should be a collection there called `uploads`, and these collections should become documents within that collection.

multinet/api.py
mvandenburgh: > Also, do you know if there's a way to read out the file size first, and abort with an error if it's above a certain limit? Otherwise this could be used to DoS the server.

Yes, looks like Flask supports this.
https://flask.palletsprojects.com/en/1.1.x/config/#MAX_CONTENT_LENGTH

From what I understand, Flask checks the actual file length and not just the `CONTENT_LENGTH` header of the request, so this should cover us in regards to DoS attacks

mvandenburgh:
The `RequiredParamsMissing` exception was returning a `list`, which is invalid for Flask so I changed it to a `dict`.

multinet/uploaders/multipart_upload.py
```diff
@@ -0,0 +1,44 @@
...
+
+# Import types
+from typing import Any
+
+from multinet.errors import RequiredParamsMissing
+
+bp = Blueprint("uploads", __name__)
+bp.before_request(util.require_db)
+
+
```
AlmightyYakob, line 18: We need to have swagger docs for these endpoints, so that they can be tried out by developers by visiting the swagger page.

AlmightyYakob: And regarding this, I'd feel free to make a new section called "Uploads". I believe you'd add a new entry to this:

https://github.com/multinet-app/multinet-server/blob/2d07777b2acf3e9dcde4c6e29f620a78f6ac285d/multinet/swagger/template.yaml#L220-L230

multinet/db.py
```diff
@@ -608,3 +608,44 @@ def count_text(filt: str) -> str:
...
+    if not sysdb.has_collection("uploads"):
+        sysdb.create_collection("uploads")
+
+    return sysdb.collection("uploads")
+
+
+def create_upload_document() -> Any:
+    """Insert empty multipart upload temp document."""
+    # returns arango document key as upload_id for client
+    return uploads_collection().insert({"0": {}}).get("_key")
```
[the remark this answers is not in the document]
mvandenburgh: It can be - for some reason I thought it caused errors, but it works fine.

multinet/__init__.py
```diff
@@ -41,6 +41,10 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     CORS(app, origins=allowed_origins, supports_credentials=True)
     Swagger(app, template_file="swagger/template.yaml")
 
+    # Set max file upload size to 2 MB
+    # TODO: decide on a non-arbitrary limit
+    app.config["MAX_CONTENT_LENGTH"] = 2000000
```
waxlamp, line 46: 32MB is a probably a good limit.

```suggestion
    app.config["MAX_CONTENT_LENGTH"] = 32 * 1024 * 1024
```

multinet/__init__.py
```diff
@@ -53,6 +57,8 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     app.register_blueprint(uploaders.nested_json.bp, url_prefix="/api/nested_json")
     app.register_blueprint(uploaders.d3_json.bp, url_prefix="/api/d3_json")
 
+    app.register_blueprint(uploaders.multipart_upload.bp, url_prefix="/api/uploads")
```
waxlamp, line 60: I don't think this should go in the `uploaders` package but I'm not sure. @AlmightyYakob, thoughts?

multinet/db.py
```diff
@@ -28,7 +28,7 @@
     WorkspacePermissions,
 )
 from multinet.auth.types import User
-from multinet.errors import InternalServerError
+from multinet.errors import InternalServerError, NonExistantUploadDocument
```
waxlamp, line 31: This should be called `UploadNotFound`.

multinet/db.py
```diff
@@ -608,3 +608,44 @@ def count_text(filt: str) -> str:
...
+
+def insert_file_chunk(key: str, sequence: str, chunk: str) -> str:
+    """Insert b64-encoded string `chunk` into temporary document."""
+    collection = uploads_collection()
+
+    document = collection.get(key)
+    if document is None:
+        raise NonExistantUploadDocument()
+
+    document[sequence] = chunk
```
waxlamp, line 638: Do we need to check whether there's already a chunk under this `sequence` ID?

multinet/db.py
```diff
@@ -608,3 +608,44 @@ def count_text(filt: str) -> str:
...
+    collection.update(document)
+
+    return key
+
+
+def delete_document(collection: StandardCollection, key: str) -> Dict:
+    """Delete a document in the given collection with the specified key."""
+    document = collection.get({"_key": key})
+    if document is None:
+        raise NonExistantUploadDocument()
```
waxlamp, line 648: The error refers to upload documents, but the function is general purpose.

It seems like this should be called `delete_upload_document()` and be specialized to just those, right?

multinet/errors.py
```diff
@@ -157,7 +157,7 @@ class RequiredParamsMissing(ServerError):
 
     def __init__(self, missing: List[str]):
         """Initialize the exception."""
-        self.missing = missing
+        self.missing = {"missing": missing}
```
waxlamp, line 160: I am curious, where did you run into this causing a problem? (Just surprised we never ran into it before.)

multinet/errors.py
```diff
@@ -242,3 +242,15 @@ def __init__(self, message: str = ""):
...
+class NonExistantUploadDocument(ServerError):
+    """Exception for errors when uploading chunks to nonexistant uploads collection."""
+
+    def __init__(self, message: str = ""):
+        """Initialize error message."""
+        self.message = message
+
+    def flask_response(self) -> FlaskTuple:
+        """Generate a 404 error."""
+        return (self.message, "404 Upload Does Not Exist")
```
waxlamp, line 256:
```suggestion
        return (self.message, "404 Upload Not Found")
```

multinet/errors.py
```diff
@@ -242,3 +242,15 @@ def __init__(self, message: str = ""):
     def flask_response(self) -> FlaskTuple:
         """Generate a 400 error."""
         return (self.message, "400 Error during AQL Execution")
+
+
+class NonExistantUploadDocument(ServerError):
```
waxlamp, line 247: I believe there's a `NotFound` base class you can use here. Not including a code suggestion because the rest of the class would need to change as well (it would become simpler, IIUC).

multinet/uploaders/multipart_upload.py
```diff
@@ -0,0 +1,44 @@
...
+    stringified_blob = b64encode(chunk.read()).decode("ascii")
+
+    db.insert_file_chunk(upload_id, sequence, stringified_blob)
+    return sequence
+
+
+@bp.route("/<upload_id>", methods=["DELETE"])
+def delete_upload_collection(upload_id: str) -> Any:
+    """Delete the database collection associated with the given upload_id."""
+    return db.delete_document(db.uploads_collection(), upload_id)
```
waxlamp, line 44: This is connected to the other comment on the matter but I believe this line should read:

```suggestion
    return db.delete_upload_document(upload_id)
```

multinet/errors.py
mvandenburgh: `RequiredParamsMissing` is only used once in `master` [here](https://github.com/multinet-app/multinet-server/blob/master/multinet/api.py#L274), I'm guessing that's why it wasn't noticed. It happens as a result of the exception returning a `list` to Flask, since Flask doesn't allow lists to be returned as endpoint responses unless they are converted to JSON with `jsonify`.

This is the error that currently occurs in `master` if that endpoint I linked above is called without an `edge_table` parameter -
`TypeError: The view function did not return a valid response. The return type must be a string, dict, tuple, Response instance, or WSGI callable, but it was a list.`

multinet/db.py
mvandenburgh: Yes, since having two documents with the same `sequence` ID would mean a corrupted file.

multinet/api.py
mvandenburgh: Done. Also resolves https://github.com/multinet-app/multinet-server/pull/448#discussion_r469379410

multinet/__init__.py
```diff
@@ -41,6 +41,10 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     CORS(app, origins=allowed_origins, supports_credentials=True)
     Swagger(app, template_file="swagger/template.yaml")
 
+    # Set max file upload size to 2 MB
+    # TODO: decide on a non-arbitrary limit
```
waxlamp, line 45: Please fix the comment and remove the TODO line.

multinet/uploaders/swagger/chunk_upload.yaml
```diff
@@ -0,0 +1,41 @@
...
+---
+parameters:
+  - $ref: "#/parameters/upload_id"
+  - name: sequence
+    in: query
+    description: The sequence number of the multipart upload
+    required: true
+    schema:
+      type: number
+      example: 0
```
waxlamp, line 11: There's a missing parameter--the chunk data itself. See `aql.yaml` for an example of how to specify a body parameter in Swagger.

multinet/__init__.py
AlmightyYakob: Yeah I agree, it should probably have its own `upload` blueprint, but I'd be okay dealing with it in a later PR.

mvandenburgh:
> I'm just now realizing that this PR isn't implementing things the way that's specified in the RFC. The RFC specifies that each upload maps to a collection, which each chunk mapping to a document in that collection. The way this is currently implemented maps each upload to a document in a collection, storing all of the chunks into that document. This is definitely not something we want to do, as we'd probably run into limits around document size that way.

I was following @waxlamp's comment here https://github.com/multinet-app/multinet-server/pull/448#discussion_r469382193.

> We'll also need a way to keep track of which collections correspond to which uploads, and we'd also need a designated database to store these collections. In [this](https://github.com/multinet-app/multinet-rfcs/pull/13#discussion_r465337151) conversation, we talked about having an `_uploads` database, which would store upload data as mentioned above. We also mentioned that we'd need a collection to store the metadata to keep track of uploads, which would (at least initially) basically be a mapping of an upload ID to the name of that upload collection in `_uploads`. This collection would probably be an `uploads` collection in the `_system` database.

Is it possible to just make the upload ID the name of the upload collection in `_uploads`? Or are there issues with that approach?

AlmightyYakob:
> > I'm just now realizing that this PR isn't implementing things the way that's specified in the RFC. The RFC specifies that each upload maps to a collection, which each chunk mapping to a document in that collection. The way this is currently implemented maps each upload to a document in a collection, storing all of the chunks into that document. This is definitely not something we want to do, as we'd probably run into limits around document size that way.
> 
> I was following @waxlamp's comment here [#448 (comment)](https://github.com/multinet-app/multinet-server/pull/448#discussion_r469382193).

@waxlamp is there a reason to diverge from the RFC?

> > We'll also need a way to keep track of which collections correspond to which uploads, and we'd also need a designated database to store these collections. In [this](https://github.com/multinet-app/multinet-rfcs/pull/13#discussion_r465337151) conversation, we talked about having an `_uploads` database, which would store upload data as mentioned above. We also mentioned that we'd need a collection to store the metadata to keep track of uploads, which would (at least initially) basically be a mapping of an upload ID to the name of that upload collection in `_uploads`. This collection would probably be an `uploads` collection in the `_system` database.
> 
> Is it possible to just make the upload ID the name of the upload collection in `_uploads`? Or are there issues with that approach?

That's one possible approach, however if we want to store metadata about each upload (filetype, current sequence number, etc.), we'd probably want the `uploads` collection in the `_system` db to contain that info.

waxlamp:
> > > I'm just now realizing that this PR isn't implementing things the way that's specified in the RFC. The RFC specifies that each upload maps to a collection, which each chunk mapping to a document in that collection. The way this is currently implemented maps each upload to a document in a collection, storing all of the chunks into that document. This is definitely not something we want to do, as we'd probably run into limits around document size that way.
> > 
> > 
> > I was following @waxlamp's comment here [#448 (comment)](https://github.com/multinet-app/multinet-server/pull/448#discussion_r469382193).
> 
> @waxlamp is there a reason to diverge from the RFC?
> 
> > > We'll also need a way to keep track of which collections correspond to which uploads, and we'd also need a designated database to store these collections. In [this](https://github.com/multinet-app/multinet-rfcs/pull/13#discussion_r465337151) conversation, we talked about having an `_uploads` database, which would store upload data as mentioned above. We also mentioned that we'd need a collection to store the metadata to keep track of uploads, which would (at least initially) basically be a mapping of an upload ID to the name of that upload collection in `_uploads`. This collection would probably be an `uploads` collection in the `_system` database.
> > 
> > 
> > Is it possible to just make the upload ID the name of the upload collection in `_uploads`? Or are there issues with that approach?
> 
> That's one possible approach, however if we want to store metadata about each upload (filetype, current sequence number, etc.), we'd probably want the `uploads` collection in the `_system` db to contain that info.

@AlmightyYakob, @mvandenburgh, and I met to discuss this stuff in person.

Main conclusions: I made a mistake in directing Mike to complete this PR as stated in the quoted comment. We will follow the RFC instead of my ad hoc comment: uploads will be represented by *collections* in a special database, and chunks will be *documents* in each collection. The indirection table in sysdb that will be used for tracking upload metadata will not be implemented right now, since we don't need it to get the basic API for uploads implemented.

multinet/db.py
```diff
@@ -608,3 +610,48 @@ def count_text(filt: str) -> str:
...
+def insert_file_chunk(upload_id: str, sequence: str, chunk: str) -> str:
+    """Insert b64-encoded string `chunk` into temporary collection."""
+    uploads_db = uploads_database()
+    if not uploads_db.has_collection(upload_id):
+        raise UploadNotFound(upload_id)
+
+    collection = uploads_db.collection(upload_id)
+
+    if collection.get(sequence) is not None:
+        raise AlreadyExists("Document", sequence)
```
waxlamp, line 641: The resource that already exists is a chunk, not just a generic document. I suggest:

```suggestion
        raise AlreadyExists("Upload Chunk", `${upload_id}/${sequence}`)
```

waxlamp:
@mvandenburgh, actually, the Swagger docs don't seem to be working: https://multinet-pr-448.herokuapp.com/apidocs/

multinet/db.py
waxlamp: (I am still recommending this change, despite dismissing the review it is part of.)

mvandenburgh:
@waxlamp swagger docs should be fixed now. I also applied the change you suggested for the `AlreadyExists` error.

multinet/db.py
JackWilb: The string interpolation here is wrong. This is js style, not python

multinet/swagger/template.yaml
```diff
@@ -173,6 +173,15 @@ parameters:
...
 
+  upload_id:
+    name: upload_id
+    in: path
+    description: ArangoDB key for multipart upload file document
+    required: true
+    schema:
+      type: integer
+      example: 50841
+
```
AlmightyYakob, line 184: I don't feel strongly about what I put in the description, just wanted to update the language to refer to collections instead of documents

```suggestion
  upload_id:
    name: upload_id
    in: path
    description: The collection ID associated with this upload
    required: true
    schema:
      type: string
      example: u-1234abcd

```

multinet/uploaders/swagger/delete_upload_collection.yaml
```diff
@@ -0,0 +1,19 @@
+Delete a multipart-uploaded chunk from the database
```
AlmightyYakob, line 1:
```suggestion
Delete a multipart-upload collection from the database
```

multinet/uploaders/swagger/delete_upload_collection.yaml
```diff
@@ -0,0 +1,19 @@
...
+---
+parameters:
+  - $ref: "#/parameters/upload_id"
+responses:
+  200:
+    description: The `upload_id` of the deleted collection
+    schema:
+      type: object
+      example: {'_key': 50841, '0': {}, '1': {}}
```
AlmightyYakob, line 10:
```suggestion
      type: string
      example: u-1234abcd
```

multinet/uploaders/swagger/delete_upload_collection.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+  404:
+    description: Document with key of `upload_id` not found
+    schema:
+      type: number
+      example: 50841
```
AlmightyYakob, line 16:
```suggestion
      type: string
      example: u-1234abcd
```

multinet/uploaders/swagger/create_upload.yaml
```diff
@@ -0,0 +1,11 @@
+Create a new document in `uploads` collection for multipart upload storage
+---
+responses:
+  200:
+    description: Unique ID (ArangoDB key) for document that was created
+    schema:
+      type: number
+      example: 56039
```
AlmightyYakob, line 8:
```suggestion
Create a new collection in the `uploads` database for multipart upload storage
---
responses:
  200:
    description: Unique ID for the collection that was created
    schema:
      type: string
      example: u-1234abcd
```

multinet/uploaders/swagger/delete_upload_collection.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - $ref: "#/parameters/upload_id"
+responses:
+  200:
+    description: The `upload_id` of the deleted collection
+    schema:
+      type: object
+      example: {'_key': 50841, '0': {}, '1': {}}
+
+  404:
+    description: Document with key of `upload_id` not found
```
AlmightyYakob, line 13:
```suggestion
    description: Collection with ID of `upload_id` not found
```

multinet/uploaders/swagger/chunk_upload.yaml
```diff
@@ -0,0 +1,46 @@
...
+      type: array
+      items:
+        type: string
+      example: {"missing": ['chunk'] }
+
+  404:
+    description: Collection with specified `upload_id` does not exist
+    schema:
+      type: number
+      example: 50841
```
AlmightyYakob, line 37:
```suggestion
      type: string
      example: u-1234abcd
```

AlmightyYakob:
Something I just realized, due to the `register_legacy_workspaces` function, the `uploads` collection will have an entry created in the workspace mapping once the server is restarted.

https://github.com/multinet-app/multinet-server/blob/0c8799a0d454d5a3ddcc08b171f74f8eb9170928/multinet/db.py#L93-L103

The change I've put below should be all that's needed to fix this:

```
system_databases = {"_system", "uploads"}
databases = {name for name in sysdb.databases() if name not in system_databases}
```

We probably should remove this function, as this is more akin to a database migration, which we have yet to set up infrastructure for. However, that is a separate issue.

Issue #453, opened 2020-08-20
Consider replacing existing cache system with memcached


AlmightyYakob:
Our current caching system is pretty barbaric. In production we run several workers with `gunicorn`, and because our use of `lru_cache` only applies to each worker, this can lead to sporadic and undesired behavior in production. [Here's](https://krzysztofzuraw.com/blog/2017/gunicorn-lru-cache-pitfall.html) an article describing this exact situation.

I realize caching is a code smell, but I'd wager one of the biggest reasons for that is our implementation. Some of the notable issues with our current caching setup:

* No synchronization between workers
* Potential to poison the cache by directly modifying the return of a cached function, thereby updated the cache itself

Both of these can be fixed by using an actual caching system. There's well defined ways to deal with caching across multiple workers, one of which being memcached. Heroku even has a [guide](https://devcenter.heroku.com/articles/flask-memcache#add-caching-to-flask) for adding memcached to an application. Another option is Redis, but I feel like that's adding more complexity than is necessary for our needs.

Ideally we wouldn't use caching, but since our application is setup in a distributed manner, we may see too many performance hits without a caching system at all. If we use a designated caching system like memcached, then there's no longer a synchronization issue between workers, and would shift the potential for error to our implementation, which hopefully is good.

Let me know what you guys think.

waxlamp:
This is a good idea. I don't know much about memcached personally but I know we have used it plenty across projects in D&A.

If we decided to use memcached, would it *completely* replace our ad hoc usage of `@lru_cache`? It would be something we'd use in production but not worry about locally for development, right?

If the answer to those two things is "yes", then I'd say to go ahead and write an RFC describing how we would use it in production.

Thanks for bringing this up. I agree with you (as long as the devops work is not too costly to set this up) that this is a far better solution than the "barbaric" thing we're currently doing.

Issue #406, opened 2020-06-25
User `import_bulk()` instead of `insert_many()`


waxlamp:
`import_bulk()` is faster than `insert_many()` because it doesn't track error conditions as finely, simply raising an exception on the first insertion failure, rather than doing all the inserts and individually tracking which insertions failed. For us, we would prefer to fail early in this case, since we wish to do these imports as all-or-nothing in general.

See [original discussion](https://github.com/multinet-app/multinet-server/pull/403#issuecomment-649748816).

PR #455, opened 2020-08-21
Add ORM

### LINKED ISSUE #442: Implement ORM
The [relevant RFC](https://github.com/multinet-app/multinet-rfcs/pull/12) has been accepted, so this can be implemented.

When this is addressed, the temporary functions in `api.py` must be removed, and the relevant code updated https://github.com/multinet-app/multinet-server/pull/430#discussion_r464475107. The referenced code is converting the underlying permissions data into a different data model (which includes full user objects, instead of just the `sub` value). Ideally we should be storing this data as full user objects, or we should have a serializer built-in, that does this conversion for us, if we're worried about data duplication.

jjnesbitt:
Fixes multinet-app/multinet-server#442.

This adds an ORM for database operations, and better logical divisions between workspaces, tables, graphs, and users. Notably, this PR **does not** address the temporary permissions transforms being done when interfacing with the permissions endpoints. I held off on this because this is a big enough change as it is, and if I also included that change, it would require a data migration. I've created multinet-app/multinet-server#456 to track this.

All tests are passing (at least locally for me). Please take your time as there's a lot to unpack here.


AlmightyYakob:
Also just a note, one of my biggest gripes with the current implementation is the sheer amount of methods on the `Workspace` class. Ideally these would be more split up between the classes, but I'm not sure if there's a sane way to do it. If you guys come up with any suggestions regarding this, please lmk.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+    @staticmethod
+    def list_public() -> Generator[str, None, None]:
+        """Return a list of all public workspace names."""
+        coll = workspace_mapping_collection()
+        return (doc["name"] for doc in coll.find({"permissions.public": True}))
+
+    @staticmethod
+    def from_dict(d: Dict) -> Workspace:
+        """Construct a workspace from a dict."""
+        workspace = Workspace(name=d["name"])
```
JackWilb, line 119: We could check if `d["name"]` exists first and throw InternalServerError if not

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+
+        internal = d.get("internal")
+        if internal:
+            workspace.internal = internal
+
+        permissions = d.get("permissions")
+        if permissions:
+            workspace.permissions = WorkspacePermissions(**permissions)
```
JackWilb, line 127:
```suggestion
            workspace.set_permissions(WorkspacePermissions(**permissions))
```

multinet/uploaders/d3_json.py
```diff
@@ -107,28 +108,20 @@ def upload(workspace: str, graph: str) -> Any:
...
-        nodes_coll = space.create_collection(node_table_name, edge=False)
+        node_table = loaded_workspace.create_table(node_table_name, edge=False)
 
-    if space.has_collection(edge_table_name):
-        links_coll = space.collection(edge_table_name)
+    if loaded_workspace.has_table(edge_table_name):
+        edge_table = loaded_workspace.table(edge_table_name)
     else:
-        links_coll = space.create_collection(edge_table_name, edge=True)
+        edge_table = loaded_workspace.create_table(edge_table_name, edge=True)
```
JackWilb, line 119: This seems like it could be buggy. Should we reject the upload if there is already a table with a name that we need?

multinet/uploaders/nested_json.py
```diff
@@ -98,38 +99,31 @@ def upload(workspace: str, graph: str) -> Any:
...
-        int_nodetable = space.create_collection(int_nodetable_name)
+        int_nodetable = loaded_workspace.create_table(int_nodetable_name, edge=False)
 
-    if space.has_collection(leaf_nodetable_name):
-        leaf_nodetable = space.collection(leaf_nodetable_name)
+    if loaded_workspace.has_table(leaf_nodetable_name):
+        leaf_nodetable = loaded_workspace.table(leaf_nodetable_name)
     else:
-        leaf_nodetable = space.create_collection(leaf_nodetable_name)
+        leaf_nodetable = loaded_workspace.create_table(leaf_nodetable_name, edge=False)
```
JackWilb, line 115: Again, this seems like it will cause us headaches with tables not containing what we expect. We should consider rejecting the upload if the tables exist.

multinet/uploaders/d3_json.py
```diff
@@ -107,28 +108,20 @@ def upload(workspace: str, graph: str) -> Any:
         del link["target"]
 
     # Create or retrieve the workspace
```
JackWilb, line 110: This comment is wrong

multinet/uploaders/newick.py
```diff
@@ -86,17 +87,17 @@ def upload(workspace: str, graph: str) -> Any:
...
-        edgetable = space.create_collection(edgetable_name, edge=True)
+        edgetable = loaded_workspace.create_table(edgetable_name, edge=True)
 
-    if space.has_collection(nodetable_name):
-        nodetable = space.collection(nodetable_name)
+    if loaded_workspace.has_table(nodetable_name):
+        nodetable = loaded_workspace.table(nodetable_name)
     else:
-        nodetable = space.create_collection(nodetable_name)
+        nodetable = loaded_workspace.create_table(nodetable_name, edge=False)
```
JackWilb, line 100: Consider rejecting upload instead

multinet/uploaders/newick.py
```diff
@@ -105,29 +106,25 @@ def read_tree(parent: Optional[str], node: newick.Node) -> None:
...
             edgetable.insert(
-                {
-                    "_from": "%s/%s" % (nodetable_name, parent),
-                    "_to": "%s/%s" % (nodetable_name, key),
-                    "length": node.length,
-                }
+                [
+                    {
+                        "_from": "%s/%s" % (nodetable_name, parent),
+                        "_to": "%s/%s" % (nodetable_name, key),
```
JackWilb, line 119:
```suggestion
                        "_from": f"{nodetable_name}/{parent}",
                        "_to": f"{nodetable_name}/{key}",
```

JackWilb (changes requested):

JackWilb:
Looks great so far, Jake! That commit where you remove 500 lines of code felt really nice

multinet/db.py
```diff
@@ -58,42 +58,41 @@
     port=int(os.environ.get("ARANGO_PORT", "8529")),
     protocol=os.environ.get("ARANGO_PROTOCOL", "http"),
 )
-restricted_keys = {"_rev", "_id"}
 
 
-def db(name: str) -> StandardDatabase:
+def db(name: str, readonly: bool = True, immediate: bool = False) -> StandardDatabase:
```
waxlamp, line 38: What does `immediate` do?

multinet/db.py
```diff
@@ -58,42 +58,41 @@
...
-        name,
-        username="root",
-        password=os.environ.get("ARANGO_PASSWORD", "letmein"),
-        verify=True,
+
+    username = "readonly" if readonly else "root"
+    password = (
+        os.environ.get("ARANGO_READONLY_PASSWORD", "letmein")
+        if readonly
+        else os.environ.get("ARANGO_PASSWORD", "letmein")
```
waxlamp, line 45: You can maybe factor out the if-else expression to apply to just the environment variable.

multinet/db.py
```diff
@@ -58,42 +58,41 @@
...
-        verify=True,
-    )
+
+@lru_cache()
+def system_db(readonly: bool = True) -> StandardDatabase:
+    """Return the singleton `_system` db handle."""
+    return db("_system", readonly)
 
 
+# TODO: Fix/remove this
```
waxlamp, line 57: What needs to be checked/removed? If appropriate please file an issue and link here, or remove the comment.

multinet/db.py
```diff
@@ -616,10 +615,10 @@ def count_text(filt: str) -> str:
 @lru_cache(maxsize=1)
 def uploads_database() -> StandardDatabase:
     """Return the database used for storing multipart upload collections."""
-    sysdb = db("_system")
+    sysdb = system_db(readonly=False)
     if not sysdb.has_database("uploads"):
         sysdb.create_database("uploads")
-    return db("uploads")
+    return db("uploads", readonly=False)
```
waxlamp, line 160: The function should take a `readonly` parameter, like `workspace_mapping_collection()` does.

multinet/db.py
```diff
@@ -656,3 +655,4 @@ def delete_upload_collection(upload_id: str) -> str:
     uploads_db.delete_collection(upload_id)
 
     return upload_id
+# TODO: Refactor the below functions into an `Upload` class
```
waxlamp: What does this comment refer to?

multinet/user.py
```diff
@@ -1,163 +1,204 @@
 """User data and functions."""
+from __future__ import annotations  # noqa: T484
```
waxlamp, line 2: What's this all about?

multinet/user.py
```diff
@@ -1,163 +1,204 @@
...
+        # Keeps track of multinet metadata
+        self.multinet: Optional[MultinetInfo] = None
+
+    @staticmethod
+    def exists(sub: str) -> bool:
+        """Search the user collection for a user that has the matching `sub` value."""
+        return User.get(sub) is not None
+
+    @staticmethod
+    def get(sub: str) -> Optional[Dict]:
```
waxlamp, line 81: Use a more specific type than `Dict`.

multinet/user.py
```diff
@@ -1,163 +1,204 @@
...
+from typing import Optional, Dict, Generator, Any
 
 MULTINET_COOKIE = "multinet-token"
 
 
-def user_collection() -> StandardCollection:
-    """Return the collection that contains user documents."""
-    sysdb = db("_system")
+@dataclass
+class MultinetInfo:
```
waxlamp, line 19: I feel like this needs a much better class name.

Why can't this information be included in the `UserInfo` class directly?

multinet/user.py
```diff
@@ -1,163 +1,204 @@
...
+
+        try:
+            return User.from_dict(
+                next(coll.find({"multinet.session": session_id}, limit=1))
+            )
+        except StopIteration:
+            return None
+
+    @staticmethod
+    def from_dict(d: Dict) -> User:
```
waxlamp, line 121: More specific type than `Dict`.

multinet/auth/util.py
```diff
@@ -27,23 +25,23 @@ def wrapper(*args: Any, **kwargs: Any) -> Any:
...
-def is_reader(user: Optional[UserInfo], workspace: Workspace) -> bool:
+def is_reader(user: Optional[User], workspace: Workspace) -> bool:
     """Indicate whether `user` has read permissions for `workspace`."""
-    perms = workspace["permissions"]
+    perms = workspace.permissions
 
     # A non-logged-in user, by definition, is a reader of public workspaces.
     if user is None:
-        return perms["public"]
+        return bool(perms.public)
```
waxlamp, line 34: Isn't `perms.public` already a `bool`?

multinet/downloaders/csv.py
```diff
@@ -31,13 +26,15 @@ def download(workspace: str, table: str) -> Any:
...
+    loaded_workspace = Workspace(workspace)
+    if not loaded_workspace.has_table(table):
         raise NotFound("table", table)
 
-    limit = workspace_table_row_count(workspace, table)
-    table_rows = workspace_table_rows(workspace, table, 0, limit)
-    fields = workspace_table_keys(workspace, table, filter_keys=True)
+    loaded_table = loaded_workspace.table(table)
+    limit = loaded_table.row_count()
+    table_rows = loaded_table.rows(0, limit)["rows"]
```
waxlamp, line 35: Can't you just call `.rows()` without the `limit` parameter and get all the rows? Having to call `.row_count()` first seems like bad design.

multinet/uploaders/d3_json.py
waxlamp: If that's a bug, it's independent of this PR. Could you file an issue please?

multinet/db.py
```diff
@@ -132,358 +107,33 @@ def workspace_mapping(name: str) -> Optional[Dict]:
...
-    """Return a specific graph named `name` in workspace `workspace`."""
-    get_graph_collection(workspace, graph)
+    return sysdb.collection("users")
 
-    # Get the lists of node and edge tables.
-    node_tables = graph_node_tables(workspace, graph)
-    edge_table = graph_edge_table(workspace, graph)
 
-    return {"nodeTables": node_tables, "edgeTable": edge_table}
+def search_user(query: str) -> Cursor:
```
waxlamp, line 120: Why can't this be a static method of the `User` class?

multinet/db.py
```diff
@@ -132,358 +107,33 @@ def workspace_mapping(name: str) -> Optional[Dict]:
...
-    """
-    nodes = aql_query(workspace, node_query)
-
-    # Get the total node count.
-    count_query = f"""
-    FOR c in [{", ".join(node_tables)}]
-      FOR d in c
-        COLLECT WITH COUNT INTO count
-        RETURN count
+          LIMIT 50
```
waxlamp, line 132: I know it has nothing to do with this PR, but let's file an issue to make this search function pageable.

multinet/db.py
```diff
@@ -500,118 +150,7 @@ def _run_aql_query(
...
-    query = query_text(filter_clause)
-    count = count_text(filter_clause)
-
-    return {
-        "edges": list(aql_query(workspace, query)),
-        "count": next(aql_query(workspace, count)),
-    }
-
-
+# TODO: Refactor the below functions into an `Upload` class
```
waxlamp, line 153: File an issue for this please, and link it in the comment.

multinet/util.py
```diff
@@ -2,21 +2,81 @@
...
-from multinet.types import EdgeTableProperties
+from multinet import workspace
+
 from multinet.errors import DatabaseNotLive, DecodeFailed
 
 TEST_DATA_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), "../test/data"))
+restricted_document_keys = {"_rev", "_id"}
+
+
+# TODO: Remove once permission storage is updated
```
waxlamp, line 22: File/link issue here please.

multinet/util.py
```diff
@@ -2,21 +2,81 @@
...
+                user = multinet.user.User.from_id(sub)
+                if user is not None:
+                    new_users.append(user.asdict())
+
+            new_permissions[role] = new_users
+
+    return new_permissions
+
+
+# TODO: Remove once permission storage is updated
```
waxlamp, line 57: Link issue here.

multinet/api.py
```diff
@@ -1,48 +1,43 @@
...
     user = current_user()
 
-    # Filter all workspaces based on whether it should be shown to the user who
-    # is logged in.
-    stream = util.stream(w["name"] for w in db.get_workspaces() if is_reader(user, w))
-    return stream
+    if user is not None:
+        return util.stream(user.available_workspaces())
+
+    return util.stream(Workspace.list_public())
```
waxlamp, line 40: The docstring says "available" but the method call is `list_public()`. Should it be `list_available()`, or are we conflating the meanings of purely public workspaces and all available workspaces?

mypy_stubs/arango/cursor.pyi
```diff
@@ -1,3 +1,21 @@
-from typing import Any, Generator
+from typing import Any
 
-Cursor = Generator[Any, None, None]
+# Below is autogenerated with stubgen
```
waxlamp, line 3: Neat, what is `stubgen`?

mypy_stubs/arango/graph.pyi
```diff
@@ -1,8 +1,11 @@
...
+from typing import Dict, List, Optional, Any
 from arango.collection import EdgeCollection, VertexCollection  # type: ignore
 
 class Graph:
     def edge_collection(self, name: str) -> EdgeCollection: ...
     def vertex_collection(self, name: str) -> VertexCollection: ...
     def vertex_collections(self) -> List[str]: ...
     def edge_definitions(self) -> List[Dict]: ...
+    def vertex(
+        self, vertex: Any, rev: Optional[Any] = ..., check_rev: bool = ...
```
waxlamp, line 10: We can't do better than `Any` here?

waxlamp (changes requested):
This is great work. I left a *lot* of comments. What I suggest is that non-critical requested changes be filed as issues (with a link to the comment), while more critical ones be fixed before merge. We can discuss in the comment threads what is critical vs. not.

The other change I would like to see: I think that ORM classes are related enough to what's left in db.py that they should go into their own subpackage together: `db.py` should move to `db/__init__.py`, and the ORM classes should be moved to `db/orm/`. Let me know what you think.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def insert(self, rows: List[Dict]) -> List[Dict]:
+        """
+        Insert rows into this table.
+
+        Returns the metadata from the documents inserted.
+        """
+        return self.handle.insert_many(rows)
+
+    # This may be a reason to split Table up into EdgeTable and NodeTable
```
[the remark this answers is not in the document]
AlmightyYakob: This was mainly a question of if it makes sense to split the `Table` class into a `NodeTable` and `EdgeTable` class, which inherit from `Table`. All of the common functionality would be in `Table`, and then methods specific to edge/node tables (like the function below this line, which is specific to edge tables) would be placed in their respective classes. Right now I think it's just something to keep in mind, not something that needs to be addressed now. Either way, I could file an issue for it if you think it's appropriate.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def row_count(self) -> int:
+        """Return the number of rows in a table."""
+        return self.handle.count()
+
+    def keys(self) -> Iterable[str]:
+        """Return all the keys in a table."""
+        return self.handle.keys()
+
+    def row_fields(self, filter_keys: bool = False) -> List[str]:
```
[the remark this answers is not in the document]
AlmightyYakob: `keys()` returns the keys of all the documents in the table (arangodb collection). That is, all of the `_key` fields on each document. This function returns the set of keys that are present in _each_ document. Technically speaking, it returns the keys on the first document it finds in the collection, but since we ensure a consistent schema, I think this is fine.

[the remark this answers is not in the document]
AlmightyYakob: Actually, now that you've brought it up, I don't think we should even have an option for allowing the system arango keys (`_key`, `_id`, `_rev`), since they're redundant, and we know they'll always be there. I don't see what use case we'd need them for, or why we'd need to retrieve them from this function instead of just specifying them. If you're okay with it I can remove the option entirely.

Separately, if anyone has a better name than `row_fields` please lmk.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+        """Return a the fields present on each row in this table."""
+        keys = []
+        cur = self.handle.find({}, limit=1)
+
+        # TODO: Determine what should happen if cur is empty
```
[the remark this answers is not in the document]
AlmightyYakob: Technically it's possible to create a table without uploading any data to it, although none of our uploaders do that.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+        self.handle.rename(new_name)
+        self.name = new_name
+
+    def insert(self, rows: List[Dict]) -> List[Dict]:
+        """
+        Insert rows into this table.
+
+        Returns the metadata from the documents inserted.
+        """
+        return self.handle.insert_many(rows)
```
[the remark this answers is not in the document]
AlmightyYakob: Good catch, should I fix that in this PR?

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+
+        table_keys: A map of all referenced tables to the specific keys referenced.
+        from_tables: A set containing the tables referenced in the _from column.
+        to_tables: A set containing the tables referenced in the _to column.
+
+        Raises an InternalServerError if this table is not an edge table.
+        """
+        props = self.handle.properties()
+        if not props["edge"]:
+            raise InternalServerError(f"Table {self.name} is not an edge table.")
```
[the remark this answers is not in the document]
AlmightyYakob: @JackWilb I see your point regarding it being a 400 error, but there's actually no way to reach this error through the client, except for an error with our classification/handling of the data internally. However, you could probably invoke this intentionally by calling the endpoints directly. Put it in a bit of a weird spot. Specifically, you could invoke this error if you were to call [this](https://api.multinet.app/apidocs/#/graph/post_api_workspaces__workspace__graphs__graph_) endpoint with a node_table in the `edge_table` parameter. In that case, I don't think it should be a `RuntimeError`, since it is the user's fault at that point. This becomes more of an issue when other collaborators are hitting our API directly, instead of just through the client.

For the reasons I listed, I agree with Jack and think it should be a 400 level error. Curious what you think @waxlamp.

multinet/graph.py
```diff
@@ -0,0 +1,119 @@
...
+        for coll_name in coll_names:
+            coll = self.handle.vertex_collection(coll_name)
+            result_set_size += coll.all().count()
+
+            if limit is None or query_size < limit:
+                remaining_limit = min(limit, limit - query_size) if limit else None
+                cur = coll.all(skip=offset, limit=remaining_limit)
+
+                query_size += cur.count()
+                total_nodes.extend(list(cur))
```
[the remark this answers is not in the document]
AlmightyYakob: I tried using functions that already exist in `python-arango` as I think it makes sense if possible. However if it causes a significant reduction in performance compared to an AQL query, then obviously we should use that instead.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+        self.aql = aql
+
+    def rows(self, offset: Optional[int] = None, limit: Optional[int] = None) -> Dict:
+        """Return the desired rows in a table."""
+        rows = self.handle.find({}, skip=offset, limit=limit)
+        count = self.handle.all().count()
+
+        return {"count": count, "rows": list(rows)}
+
+    def row(self, doc: Union[Dict, str]) -> Union[Dict, None]:
```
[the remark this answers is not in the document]
AlmightyYakob: If it's a `str`, it can be the document ID or key. https://python-driver-for-arangodb.readthedocs.io/en/master/specs.html#arango.collection.StandardCollection.get

multinet/db.py
```diff
@@ -118,7 +118,7 @@ def workspace_mapping_collection() -> StandardCollection:
 
 # Caches the document that maps an external workspace name to it's internal one
 @lru_cache()
-def workspace_mapping(name: str) -> Optional[WorkspaceDocument]:
+def workspace_mapping(name: str) -> Optional[Dict]:
```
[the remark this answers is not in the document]
AlmightyYakob: I tried to avoid dealing with `TypedDict`s in the first pass, since that was where much of the pain points came from originally. If we can deal with them in a sane way I'm willing to add them, but I think we need to come up with a methodology for using them.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+from multinet.table import Table
+
+from typing import Any, List, Dict, Generator, Optional
+
+
+@dataclass
+class WorkspacePermissions:
+    """The permissions on a workspace."""
+
+    # TODO: Change str to User once updating permissions storage
```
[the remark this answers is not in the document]
AlmightyYakob: Here's the issue https://github.com/multinet-app/multinet-server/issues/456. I can link it in the comment.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+    owner: str
+    maintainers: List[str] = dataclasses.field(default_factory=lambda: [])
```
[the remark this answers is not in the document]
AlmightyYakob: It's what's required when setting defaults on dataclass fields that aren't primitives.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+        if Workspace.exists(name):
+            raise AlreadyExists("Workspace", name)
+
+        internal = util.generate_arango_workspace_name()
+
+        try:
+            system_db(readonly=False).create_database(internal)
+        except DatabaseCreateError:
+            # Could only happen if there's a name collision
+            raise InternalServerError("Error creating workspace")
```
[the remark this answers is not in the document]
AlmightyYakob: We could do that, however we probably wouldn't want to introduce unbounded while loops in our code base, so we would probably only do this a finite number of times before defaulting to raising this error. You could argue that based on [this](https://github.com/multinet-app/multinet-server/pull/358#discussion_r410721018) comment, the difference between the chance of this happening, and that chance multiplied by a finite number, is small enough to not worry about. That is, as soon as we start to worry about this chance, it's not clear where you'd draw the line.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+        permissions = WorkspacePermissions(owner=owner.sub).__dict__
+        workspace_dict = {
+            "name": name,
+            "internal": internal,
+            "permissions": permissions,
+        }
+
+        coll = workspace_mapping_collection(readonly=False)
+        coll.insert(workspace_dict, sync=True)
+        workspace_mapping.cache_clear()
```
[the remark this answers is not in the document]
AlmightyYakob: We could, but I didn't remove it here because I think if we're going to be possibly impacting the performance of the application, it makes sense to do that in it's own PR, so that the effects are easier to observe.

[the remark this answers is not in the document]
AlmightyYakob: Related to #453

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+        """Fetch and return the metadata for this workspace."""
+        doc = workspace_mapping(self.name)
+        if not doc:
+            raise WorkspaceNotFound(self.name)
+
+        # Copy so modifications to return don't poison cache
+        return copy.deepcopy(doc)
+
+    # Graphs
+    def graphs(self) -> List[Dict]:
```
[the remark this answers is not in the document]
AlmightyYakob: Same response as my other comment regarding TypedDicts

multinet/db.py
AlmightyYakob: Initializes the connection to the database immediately by sending a test request. This is generally not desired, as it means more upfront requests when we may not need them.

multinet/db.py
AlmightyYakob: It refers to the upload functions from Mike V's PR. It must have got split apart from his changes, but it's visible in the final change.

multinet/user.py
AlmightyYakob: If you have a class method that returns an instance of that class, and you want to type annotate it properly, this is required. In this case, I believe they're all static methods. Here's an example https://github.com/multinet-app/multinet-server/blob/550d608f6d0dc7452d35672b4edb3d94ea3969e0/multinet/user.py#L92-L100

multinet/user.py
AlmightyYakob: `UserInfo` represents the data that can be used to create a new User, while `MultinetInfo` represents the additional data that is added by our system, and that we use. This may grow, which is why having a separate dataclass to represent it is beneficial.

multinet/auth/util.py
AlmightyYakob: Good point, I'll change it.

multinet/downloaders/csv.py
AlmightyYakob: Yeah, I'll make this change. This is vestigial from the previous implementation, not sure why I didn't notice this.

multinet/api.py
AlmightyYakob: If it gets to this line, it means the user isn't logged in, and thus the only available workspaces are the ones that are public.

mypy_stubs/arango/cursor.pyi
AlmightyYakob: It's a tool included with mypy that makes it easy to generate boilerplate stubs https://mypy.readthedocs.io/en/stable/stubgen.html#automatic-stub-generation-stubgen

mypy_stubs/arango/graph.pyi
AlmightyYakob: We can, my bad

multinet/table.py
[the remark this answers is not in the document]
waxlamp: Good argument for 4xx.

I guess it could be a [422](https://httpstatuses.com/422) but tbh just a 400 with an appropriate message (such as the one you've got here) would be fine.

multinet/uploaders/d3_json.py
waxlamp: Opened #458.

multinet/uploaders/nested_json.py
waxlamp: Opened #458.

multinet/uploaders/newick.py
waxlamp: Opened #458.

multinet/graph.py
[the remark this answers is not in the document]
waxlamp: :+1:

multinet/graph.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def node_tables(self) -> Iterable[str]:
+        """Return all node tables in this graph."""
+        return self.handle.vertex_collections()
+
+    def edge_table(self) -> str:
+        """Return the edge table of this graph."""
+        edge_defs = self.handle.edge_definitions()
+
+        # Currently assume one edge definition per graph
```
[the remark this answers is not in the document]
waxlamp: Something like:

```suggestion
        # Multinet graphs have just one edge definition.
```

multinet/table.py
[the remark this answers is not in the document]
waxlamp: Great, thanks for the explanation.

BTW I noticed we can use `Optional` here as well:

```suggestion
    def row(self, doc: Union[Dict, str]) -> Optional[Dict]:
```

multinet/table.py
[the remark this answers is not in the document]
waxlamp: Ok. Then (like I said above) I think this approach is correct.

```suggestion
```

multinet/table.py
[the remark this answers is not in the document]
waxlamp: Removing that option sounds great.

I think we should call these "column headers" (or just "headers"), in keeping with the tabular data theme.

multinet/table.py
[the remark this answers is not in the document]
waxlamp: No, let's fix it in a future PR aimed at #406. I just wanted to leave the breadcrumb here. I'll mark this conversation "resolved".

multinet/table.py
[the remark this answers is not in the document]
waxlamp: Thanks for the explanation.

I opened #459 to track this. I would just delete the comment:

```suggestion
```

multinet/db.py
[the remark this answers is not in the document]
waxlamp: Let's worry about this in a followup. I can take it on myself: #460.

multinet/workspace.py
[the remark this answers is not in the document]
waxlamp: Right. Let's just do nothing.

multinet/workspace.py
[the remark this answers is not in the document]
waxlamp: Great. I forgot we were talking about using `memcached`!

multinet/workspace.py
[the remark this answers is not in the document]
waxlamp: #460

multinet/db.py
waxlamp: Why did you include it? Do we require immediate access somewhere?

multinet/db.py
waxlamp: I did notice this after I left the comment. Oops.

waxlamp: Can you file an issue and link it in the comment?

multinet/user.py
waxlamp: #460

multinet/user.py
waxlamp: Makes sense. I still think we should come up with a better name than "MultinetInfo". But I see that you're simply updating what was there before, so I've filed #461 and will resolve this conversation.

multinet/user.py
waxlamp: #460

multinet/db.py
waxlamp: #462

multinet/api.py
waxlamp: Ah, my reading comprehension skills failed.

A few comments in this block of code might be in order, to help guide readers as to what's going on.

multinet/table.py
```diff
@@ -0,0 +1,109 @@
...
+        return self.handle.count()
+
+    def keys(self) -> Iterable[str]:
+        """Return all the keys in a table."""
+        return self.handle.keys()
+
+    def row_fields(self, filter_keys: bool = False) -> List[str]:
+        """Return a the fields present on each row in this table."""
+        keys = []
+        cur = self.handle.find({}, limit=1)
```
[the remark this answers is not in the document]
waxlamp: @JackWilb if this satisfies you, please resolve the conversation thread.

multinet/workspace.py
AlmightyYakob: Would be covered under https://github.com/multinet-app/multinet-server/issues/460

multinet/workspace.py
AlmightyYakob: This code was actually redundant/incorrect, as this is all handled in the constructor of `Workspace`, and has been removed in b93c5ea

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+    """Workspace."""
+
+    # Keys that aren't saved to the database
+    exclude_keys = {"handle", "readonly_handle", "readonly"}
+
+    def __init__(self, name: str):
+        """Create a Workspace Object."""
+        self.name = name
+
+        # TODO: Don't access database right away
```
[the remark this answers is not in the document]
AlmightyYakob: #463

multinet/db.py
AlmightyYakob: While true, I personally think it would make the code more messy.

multinet/db.py
AlmightyYakob: The main reason I didn't include it in that class is that it doesn't really operate on a specific user, or interact with the `User` class in anyway, like most of the other static methods do. I understand that it would be good to have for organization though, so I can move it.

multinet/db.py
AlmightyYakob: Not sure, I can remove it.

multinet/workspace.py
```diff
@@ -0,0 +1,356 @@
...
+
+        coll = workspace_mapping_collection()
+        coll.update(doc)
+
+    def get_permissions(self) -> WorkspacePermissions:
+        """Fetch and return the permissions on this workspace."""
+        doc = self.get_metadata()
+
+        self.permissions = self.permissions or doc["permissions"]
+        self.internal = self.internal or doc["internal"]
```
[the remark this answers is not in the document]
AlmightyYakob: This was actually due to a misuse of the Workspace class. I've removed the `or` expressions in 1995a03

multinet/table.py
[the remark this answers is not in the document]
AlmightyYakob: 47ec4bd

multinet/db.py
waxlamp: Ok.

waxlamp (changes requested):
All the changes look good. Did you see my note in my first review about moving the ORM class files into an `orm` package within `db`? That's the last thing I'd request before we merge.

Really great work on this!

AlmightyYakob:
> All the changes look good. Did you see my note in my first review about moving the ORM class files into an `orm` package within `db`? That's the last thing I'd request before we merge.
> 
> Really great work on this!

Yes I recall that now. What do you think about keeping the `db.py` file where it is, and placing the class files into a `models` folder? That way, it's fairly explicit, and leaves just primitive database operations in the `db.py` file.

waxlamp:
> > All the changes look good. Did you see my note in my first review about moving the ORM class files into an `orm` package within `db`? That's the last thing I'd request before we merge.
> > Really great work on this!
> 
> Yes I recall that now. What do you think about keeping the `db.py` file where it is, and placing the class files into a `models` folder? That way, it's fairly explicit, and leaves just primitive database operations in the `db.py` file.

The name `model` is probably better than `orm`. But I still think it should be `db.model.Graph` etc. If we move `db.py` to `db/__init__.py`, and put all the class files in `db/model/`, then the import statements for the primitive operations formerly in `db.py` wouldn't have to change, right?

I think I'd prefer that solution, to make it all the way explicit that `db.model.Graph` is a database model object.

JackWilb (approved):
Looks good to me! Tested creating/deleting workspaces, graphs, and tables, viewing nodes, edges, pagination, graphs, visualizing data locally. Appears to all be working.

waxlamp (approved):
Looks great. Thanks for running this marathon, @AlmightyYakob!

PR #457, opened 2020-08-24
Add CORS function for wildcard origin matching

### LINKED ISSUE #384: More fine-grained control over allowed origins for cookies
We're using Flask-CORS currently to allow a whitelist of domains to send cookies for authentication (and therefore for the majority of API access). We're looking for a way to have the equivalent of the allowed origins being set to `*` while still allowing credential transfer (which is not allowed by Flask-CORS).

There's a possibility we can do something like this using the [after_request](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Flask.after_request) hook. This [tutorial](https://dev.to/g33konaut/understanding-cors-aaf) shows in principle how this general idea can work.

Another possibility is to implement the header-based authentication system that, e.g., Girder uses.

We should do the following as part of resolving this issue:
1. Investigate whether there are valid security reasons for Flask-CORS's policy of not allowing the setup we are after.
2. Investigate whether Girder's header-based approach is feasible for us (it may not be if the complexity of implementing it for Multinet is severely higher than our current approach, or if there are security flaws inherent in the approach).
3. Investigate whether the after_request or tutorial approaches would work for Multinet.
4. Decide whether to close this issue as a wontfix, or to implement a chosen solution. It is worth considering as part of this step that as a research project, what we already have so far is likely to be more than sufficient for our purposes.

AlmightyYakob:
Fixes #384

This adds the ability for allowed origins to be specified with wildcards. This would allow, for example, a specific subdomain of a domain to be allowed through CORS, without needing to allow the entire site, or any unwanted similar subdomains of that site. One of the main use cases here is to allow for Netlify deploy previews to access the testing application, without the need for the testing app to specify allowed origins of `*`, which is what's currently done.

~~Notably, this removes the use of `flask-cors`, replacing it entirely with our custom CORS handling. I originally was going to let this new handling and `flask-cors` co-exist, but given that `flask-cors` runs after this new function, and that my function performs a superset of functionality that `flask-cors` does, keeping it was redundant.~~

I've re-implemented this in flask-cors, since it supports it, and is a clean change.

### DIFF
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -15,7 +15,7 @@ from multinet import api
 from multinet.db import register_legacy_workspaces
 from multinet import uploaders, downloaders
 from multinet.errors import ServerError
-from multinet.util import load_secret_key
+from multinet.util import load_secret_key, regex_allowed_origins
 
 sentry_dsn = os.getenv("SENTRY_DSN", default="")
 sentry_sdk.init(dsn=sentry_dsn, integrations=[FlaskIntegration()])
@@ -37,8 +37,7 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     if config is not None:
         app.config.update(config)
 
-    allowed_origins = get_allowed_origins()
-    CORS(app, origins=allowed_origins, supports_credentials=True)
+    CORS(app, origins=regex_allowed_origins(get_allowed_origins()))
     Swagger(app, template_file="swagger/template.yaml")
 
     # Set max file upload size to 32 MB
--- a/multinet/__init__.py
+++ b/multinet/__init__.py
@@ -7,7 +7,7 @@ from flask_cors import CORS
 from flasgger import Swagger
 from sentry_sdk.integrations.flask import FlaskIntegration
 
-from typing import Optional, MutableMapping, Any, Tuple, Union, List
+from typing import Optional, MutableMapping, Any, Tuple, Union
 
 from multinet import auth
 from multinet.auth import google
@@ -15,19 +15,10 @@ from multinet import api
 from multinet.db import register_legacy_workspaces
 from multinet import uploaders, downloaders
 from multinet.errors import ServerError
-from multinet.util import load_secret_key, regex_allowed_origins
+from multinet.util import load_secret_key, regex_allowed_origins, get_allowed_origins
 
 sentry_dsn = os.getenv("SENTRY_DSN", default="")
 sentry_sdk.init(dsn=sentry_dsn, integrations=[FlaskIntegration()])
-
-
-def get_allowed_origins() -> List[str]:
-    """Read in comma-separated list of allowed origins from environment."""
-    allowed_origins = os.getenv("ALLOWED_ORIGINS", default=None)
-    if allowed_origins is None:
-        return []
-
-    return [s.strip() for s in allowed_origins.split(",")]
 
 
 def create_app(config: Optional[MutableMapping] = None) -> Flask:
--- a/multinet/util.py
+++ b/multinet/util.py
@@ -8,7 +8,7 @@ from dataclasses import asdict
 from functools import lru_cache
 from uuid import uuid1, uuid4
 from flask import Response, current_app
-from typing import Any, Generator, Dict, List, Iterable
+from typing import Any, Generator, Dict, Set, List, Iterable
 
 from multinet import db
 from multinet.db.models import workspace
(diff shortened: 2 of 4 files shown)


multinet/__init__.py
```diff
@@ -21,13 +21,48 @@
...
     allowed_origins = os.getenv("ALLOWED_ORIGINS", default=None)
     if allowed_origins is None:
-        return []
+        return set()
 
-    return [s.strip() for s in allowed_origins.split(",")]
+    return {s.strip() for s in allowed_origins.split(",")}
+
+
+def register_cors_func(app: Flask, with_credentials: bool = False) -> None:
```
waxlamp, line 33: I liked the form of this better that included the `allowed_origins` argument. Why not keep that?

multinet/__init__.py
```diff
@@ -37,8 +72,9 @@ def create_app(config: Optional[MutableMapping] = None) -> Flask:
     if config is not None:
         app.config.update(config)
 
-    allowed_origins = get_allowed_origins()
-    CORS(app, origins=allowed_origins, supports_credentials=True)
+    app.config["ALLOWED_ORIGINS"] = get_allowed_origins()
+
+    register_cors_func(app, True)
```
waxlamp, line 77: In keeping with retaining the old calling sequence:

```suggestion
    register_cors_func(app, with_credentials = True)
```

multinet/__init__.py
```diff
@@ -21,13 +21,48 @@
...
+        allowed_origins: Set[str] = app.config["ALLOWED_ORIGINS"]
+        normal_allowed_origins = allowed_origins - {"*"}
+
+        origin = request.environ.get("HTTP_ORIGIN")
+        if origin is None or cors_header in res.headers or not allowed_origins:
+            return res
+
+        # Two tests are done, because if both "*" and the actual origin are present,
+        # we want the actual origin to "win", in case of credential use, since
+        # credential use on origin "*" is not allowed.
```
JackWilb, line 54: Is this part of the spec or a security check we've added? Would this cause issues with accessing data on the vis app deploy previews? I'm asking since you need to pass credentials to get the data out, and we'd likely use the wildcard for the previews

JackWilb (commented):
LGTM with Roni's suggestions. I just had a question for you regarding the implementation of the deploy previews

multinet/__init__.py
waxlamp: It's part of how CORS works, but it applies specifically to the case where you supply `*` as the allowed origin. This shouldn't interfere with what we want to do, which is allow credentials for allowed origins featuring `*` *as a wildcard*; the approach is to use `fnmatch` to say "yes" or "no" to individual domains so it actually has nothing to do with the restriction on `*` in CORS itself.

@AlmightyYakob, please correct me if I got anything wrong.

AlmightyYakob: Yes, it's to match the CORS spec, which is that if the allowed origin is exactly `*`, then credential support is not allowed. Since we can specify multiple allowed origins, this check is to make sure that if we supply a pattern which matches the origin more strictly than `*`, that it is preferred.

multinet/__init__.py
AlmightyYakob: It's because now the allowed origins are stored in `app.config["ALLOWED_ORIGINS"]`, and are grabbed from there. This is necessary if we want testing to work properly, as otherwise the allowed origins are only set on app creation, and can't be modified. I also think that storing this option in the flask config is cleaner anyway,

waxlamp: But you can still store these in the app config, and just pass them to `CORS_wildcard()` as a keyword argument, right? Instead of referencing `app.config["ALLOWED_ORIGINS"]` in line 45 below, it could just refer to the argument, and then you would pass in `allowed_origins = app.config["ALLOWED_ORIGINS"]`. Why would this interfere with testing?

multinet/__init__.py
```diff
@@ -21,13 +21,48 @@
 sentry_sdk.init(dsn=sentry_dsn, integrations=[FlaskIntegration()])
 
 
-def get_allowed_origins() -> List[str]:
+def get_allowed_origins() -> Set[str]:
     """Read in comma-separated list of allowed origins from environment."""
     allowed_origins = os.getenv("ALLOWED_ORIGINS", default=None)
     if allowed_origins is None:
-        return []
+        return set()
```
waxlamp, line 28: I think flake8 may complain about this line.

```suggestion
        return {}
```

AlmightyYakob: You actually need to use `set()` to initialize an empty set, as `{}` represents an empty dictionary.

waxlamp: Good point.

AlmightyYakob:
@waxlamp @JackWilb I've rebased this and implemented it using flask-cors. It's a lot cleaner now, and should be good to go.

waxlamp (approved):
The test suite is actually longer than the change set! Looks good; thanks for this @AlmightyYakob!

JackWilb (approved):
Looks good to me. Would this would allow previews to access data using `ALLOWED_ORIGINS=https://deploy-preview-*--multinet-adjmatrix.netlify.app`, for example?

AlmightyYakob:
> Looks good to me. Would this would allow previews to access data using `ALLOWED_ORIGINS=https://deploy-preview-*--multinet-adjmatrix.netlify.app`, for example?

Yes, that's the underlying use case motivating this change.

PR #469, opened 2020-09-04
Fix cookie use

### LINKED ISSUE #468: Fix use of cookies
Currently, we don't take into account `SameSite` when setting cookies. If hosting the client/api from the same base address (as we do in production), this isn't an issue, because the browser sees the two addresses as the same site. However, if the api and client are not served from the same base address, then passing cookies along with requests (as we do for any authenticated call), will fail, as the browser will block the inclusion of this cookie in the request. This means that while login would "succeed", you wouldn't actually be able to take advantage of the fact that you're logged in, because any attempt to use these credentials would be blocked by the browser. Relevant [article](https://web.dev/samesite-cookies-explained/) explaining the use of `SameSite`.

We currently use Flask's [session](https://pythonbasics.org/flask-sessions/) object for storing all cookies. While this provides us cryptographically signed cookies, it limits us in our ability to control the options of individual cookie entries. We'll probably have to switch to manually using [`set_cookie`](https://flask.palletsprojects.com/en/1.1.x/api/#flask.Response.set_cookie) on flask responses when it comes to setting the user's session. However, the flask `session` can still be used to store misc data, like what url to return to after login is completed.

Below is a diagram that may be helpful in illustrating the interaction between cookie settings and CORS, when interacting with an api server on a different domain than the website.

CORS_SameSite

AlmightyYakob:
Fixes #468.

This changes our handling of cookies from using Flask's `session` object, to directly handling cookies. This is to allow the setting of the `SameSite` and `Secure` cookie options mentioned in the linked issue. Notably, Flask's `session` is still used in one place: the Google login flow. This is only to temporarily store the `return_url` parameter, so that once login is finished, the api can redirect the browser to where it came from. There is no security risk here, and it's actually a pretty nice way to store this temporary data.

This PR also renames some variables, to make it more clear what's going on.

Starting this as a draft PR so I can test out it's functionality on a live site (using https).

### DIFF
--- a/multinet/auth/google.py
+++ b/multinet/auth/google.py
@@ -6,7 +6,16 @@ import os
 
 from dacite import from_dict
 from flasgger import swag_from
-from flask import Flask, redirect, request, session, make_response, url_for
+from flask import (
+    Flask,
+    redirect,
+    request,
+    Response,
+    session,
+    make_response,
+    url_for,
+    current_app,
+)
 from werkzeug.wrappers import Response as ResponseWrapper
 from flask.blueprints import Blueprint
 from authlib.integrations.flask_client import OAuth
@@ -14,8 +23,9 @@ from authlib.integrations.flask_client import OAuth
 from webargs.flaskparser import use_kwargs
 from webargs import fields
 
-from multinet.db.models.user import User, MULTINET_COOKIE
+from multinet.db.models.user import User
 from multinet.auth.types import GoogleUserInfo
+from multinet.auth.util import LOGIN_TOKEN_COOKIE
 
 from typing import Dict, Optional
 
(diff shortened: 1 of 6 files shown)


multinet/auth/google.py
```diff
@@ -136,7 +146,15 @@ def authorized(state: str, code: str) -> ResponseWrapper:
     user.save()
 
     return_url = session.pop("return_url", default_return_url())
-    resp = make_response(redirect(ensure_external_url(return_url)))
-    session[MULTINET_COOKIE] = cookie
-
+    resp: Response = make_response(redirect(ensure_external_url(return_url)))
+
+    # If running in development, don't set secure and samesite cookie attributes
+    development = current_app.config.get("ENV") == "development"
```
waxlamp, line 152: Instead of computing this bool here, I wonder if we could do it at app startup time, or maybe move it to a util function?

It's not critical to change this now, but I'd at least like to get an issue filed in that case.

waxlamp (commented):
I don't fully understand everything here, but the code looks more or less ok to me.

Leaving this review as a comment since it's still in draft mode.

AlmightyYakob:
Superseded by #477

PR #476, opened 2020-10-07
Add a var for the arango root password in .env

JackWilb:
Adds ARANGO_PASSWORD to .env.default so that it may be cloned to .env and easily updated. This simplifies generic deployments that will use `pipenv run gunicorn` by allowing it to be read into the environment. 

ARANGO_PASSWORD is the name of the env var used for the root user.


waxlamp (commented):
Looks ok to me, but I'll let @AlmightyYakob give the formal approval, since he likely understands the security implications of this better than I do.

AlmightyYakob (approved):
:+1: No security implications, this should've been in there from the beginning.

PR #477, opened 2020-10-13
Return to using cookies for login

AlmightyYakob:
Partly reverts #472. Notably, JWTs are still used, as I believe they're a good method for issuing logins.

### DIFF
--- a/multinet/auth/google.py
+++ b/multinet/auth/google.py
@@ -6,7 +6,7 @@ import os
 
 from dacite import from_dict
 from flasgger import swag_from
-from flask import Flask, redirect, request, session, make_response, url_for
+from flask import Flask, redirect, request, Response, session, make_response, url_for
 from werkzeug.wrappers import Response as ResponseWrapper
 from flask.blueprints import Blueprint
 from authlib.integrations.flask_client import OAuth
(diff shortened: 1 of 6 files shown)


multinet/auth/google.py
```diff
@@ -139,9 +148,18 @@ def authorized(state: str, code: str) -> ResponseWrapper:
...
     encoded_login_token = encode_auth_token(create_login_token(user.get_session()))
-    resp.set_cookie(MULTINET_LOGIN_TOKEN, encoded_login_token)
+
+    # If running in development, don't set secure and samesite cookie attributes
+    development = current_app.config.get("ENV") == "development"
+    resp.set_cookie(
+        MULTINET_LOGIN_TOKEN,
+        value=encoded_login_token,
+        secure=True if not development else False,
+        samesite="None" if not development else None,
```
JackWilb, line 161: From MDN, "None" means:
> Cookies will be sent in all contexts, i.e sending cross-origin is allowed.

Is this the behavior we want? I guess so to interface with applications that are not on multinet.app

AlmightyYakob: Yes, a good example is preview deploys. We also do have CORS set up, so a request from another origin which is not in our list will be blocked by the browser.

JackWilb (approved):

multinet/auth/google.py
```diff
@@ -134,10 +147,19 @@ def authorized(state: str, code: str) -> ResponseWrapper:
...
-    encoded_login_token = encode_auth_token(create_login_token(user.get_session()))
     return_url = session.pop("return_url", default_return_url())
-    return_url_with_fragment = f"{return_url}#loginToken={encoded_login_token}"
-    resp = make_response(redirect(ensure_external_url(return_url_with_fragment)))
+    resp: Response = make_response(redirect(ensure_external_url(return_url)))
+
+    encoded_login_token = encode_auth_token(create_login_token(user.get_session()))
+
+    # If running in development, don't set secure and samesite cookie attributes
+    development = current_app.config.get("ENV") == "development"
```
waxlamp, line 156: Could be good to make this some kind of util function. I will file an issue.

waxlamp: See #478.

waxlamp (approved):

PR #480, opened 2020-11-05
Replace dacite, dataclasses with pydantic

AlmightyYakob:
During my implementation of the table metadata [RFC](https://github.com/multinet-app/multinet-rfcs/pull/18), I ran into some limitations of dacite (notably, it wasn't able to serialize into a field which had a string literal type). I figured it was about time we switched from that to a more robust and featured tool, such as pydantic.


multinet/validation/csv.py
```diff
@@ -1,29 +1,25 @@
...
 from multinet.validation import ValidationFailure, DuplicateKey, UnsupportedTable
 
 
-@dataclass
 class InvalidRow(ValidationFailure):
     """Invalid syntax in a CSV file."""
 
     row: int
-    fields: List[str]
+    columns: List[str]
```
AlmightyYakob, line 14: I renamed this because apparently `fields` is a reserved keyword in pydantic, and I figured `columns` was just as representative of the underlying data, if not more.

requirements.txt
```diff
@@ -1,39 +1,46 @@
+#
+# These requirements were autogenerated by pipenv
+# To regenerate from the project's Pipfile, run:
+#
+#    pipenv lock --requirements
+#
+
```
waxlamp, line 7: I think this header is generated by the new version of Pipenv right? Would you mind setting up a new PR that upgrades the pipenv version?

waxlamp (approved):
Looks great!

requirements.txt
AlmightyYakob: Yes it's from the new pipenv version, but I don't think there's a way we can specify pipenv in our dependencies.

waxlamp: Oh you're right. As long as everything works with the new version, then we should be fine.

multinet/validation/__init__.py
```diff
@@ -1,14 +1,14 @@
...
+from typing import List, Dict, Any
 
 
 class ValidationFailure(BaseModel):
     """Base class for any validation errors."""
 
-    def validation_dict(self) -> Dict:
-        """Return a dict representation of the Validation Failure."""
-        return {**self.dict(), "type": self.schema()["title"]}
+    def dict(self, **kwargs: Any) -> Dict:  # noqa: A003
```
waxlamp, line 9: Is this ok to do? What bad things could possibly happen because of overloading a built-in like this?

waxlamp (approved):
Approving so you can merge at your discretion, but I'd like an answer to my question below (mainly for my own education).

multinet/validation/__init__.py
AlmightyYakob: Well this isn't an overload of a built-in, since this is a class method, and the built-in `dict` is just a function. I think flake8 just sees a function named `dict` and complains about it.

waxlamp: Oh, I see. Seems like a bug in `flake8`.

Issue #405, opened 2020-06-25
Pull all migration actions together into one place


waxlamp:
This might be a collection of scripts (or one huge script) in a specific place in the repo, with a pipenv script attached to it.

Currently there is at least one Python script (for patching workspace permission information), and a server startup that syncs the workspace_mapping collection with the databases. If there are other migration things, they should be included in this.

We should also investigate whether the `flask migrate` command could be used here in some way.

TODO:
- [ ] pull `register_legacy_workspaces()` into a migration script (see #342)
- [ ] Add migration for table metadata (see https://github.com/multinet-app/multinet-server/pull/484#discussion_r519953325)

PR #484, opened 2020-11-06
Introduce Table Metadata

### LINKED ISSUE #482: Implement table metadata
This will follow the guidelines in [RFC 0009](https://github.com/multinet-app/multinet-rfcs/tree/master/accepted/0009-table-metadata).

AlmightyYakob:
Closes #482 

This is a basic first approach to table metadata. In this approach, table metadata is passed as a query string argument to the CSV upload endpoint. This is not ideal, and is only in place until [this](https://github.com/multinet-app/multinet-rfcs/issues/20) issue is addressed. 

One of the side effects is that this PR slightly changes how tables are created. Now, the entire workspace instance is passed into the `Table` constructor, which the constructor selects things from. This was mainly needed for referring to the workspace metadata collection. I actually think this approach is cleaner overall.

### DIFF
--- a/multinet/api.py
+++ b/multinet/api.py
@@ -5,7 +5,7 @@ from webargs import fields
 from webargs.flaskparser import use_kwargs
 
 from typing import Any, Optional
-from multinet.types import EdgeDirection, TableType
+from multinet.types import EdgeDirection, UnionTableType
 from multinet.auth.util import (
     require_login,
     require_reader,
@@ -73,7 +73,9 @@ def set_workspace_permissions(workspace: str) -> Any:
 @require_reader
 @use_kwargs({"type": fields.Str()})
 @swag_from("swagger/workspace_tables.yaml")
-def get_workspace_tables(workspace: str, type: TableType = "all") -> Any:  # noqa: A002
+def get_workspace_tables(
+    workspace: str, type: UnionTableType = "all"  # noqa: A002
+) -> Any:
     """Retrieve the tables of a single workspace."""
     tables = Workspace(workspace).tables(type)
     return util.stream(tables)
@@ -98,6 +100,23 @@ def create_aql_table(workspace: str, table: str) -> Any:
 def get_table_rows(workspace: str, table: str, offset: int = 0, limit: int = 30) -> Any:
     """Retrieve the rows and headers of a table."""
     return Workspace(workspace).table(table).rows(offset, limit)
+
+
+@bp.route("/workspaces/<workspace>/tables/<table>/metadata", methods=["GET"])
+@require_reader
+@swag_from("swagger/get_metadata.yaml")
+def get_table_metadata(workspace: str, table: str) -> Any:
+    """Retrieve the metadata of a table, if it exists."""
+    metadata = Workspace(workspace).table(table).get_metadata()
+    return "" if metadata is None else metadata.dict()
+
+
+@bp.route("/workspaces/<workspace>/tables/<table>/metadata", methods=["PUT"])
+@require_reader
+@swag_from("swagger/set_metadata.yaml")
+def set_table_metadata(workspace: str, table: str) -> Any:
+    """Retrieve the rows and headers of a table."""
+    return Workspace(workspace).table(table).set_metadata(request.json).dict()
 
 
 @bp.route("/workspaces/<workspace>/graphs", methods=["GET"])
--- a/multinet/db/models/table.py
+++ b/multinet/db/models/table.py
@@ -1,11 +1,19 @@
 """Operations that deal with tables."""
 from __future__ import annotations  # noqa: T484
+
 from arango.collection import StandardCollection
 from arango.aql import AQL
+from pydantic import ValidationError as PydanticValidationError
 
 from multinet import util
-from multinet.types import EdgeTableProperties
-from multinet.errors import ServerError, FlaskTuple
+from multinet.db.models import workspace
+from multinet.types import (
+    EdgeTableProperties,
+    ArangoEntityDocument,
+    EntityMetadata,
+    TableMetadata,
+)
+from multinet.errors import ServerError, FlaskTuple, InvalidMetadata
 
 from typing import List, Set, Dict, Iterable, Union, Optional
 
@@ -25,22 +33,23 @@ class NotAnEdgeTable(ServerError):
 class Table:
     """Tables store tabular data, and are the root of all data storage in Multinet."""
 
-    def __init__(self, name: str, workspace: str, handle: StandardCollection, aql: AQL):
+    def __init__(self, name: str, workspace: workspace.Workspace):
         """
         Initialize all Table parameters, but make no requests.
 
-        The `workspace` parameter is the name of the workspace this table belongs to.
-
-        The `handle` parameter is the handle to the arangodb collection for which this
-        class instance is associated.
-
-        The `aql` parameter is the AQL handle of the creating Workspace, so that this
-        class may make AQL requests when necessary.
+        The `name` parameter is the name of this table.
+        The `workspace` parameter is the workspace this table belongs to.
         """
         self.name = name
-        self.workspace = workspace
-        self.handle = handle
-        self.aql = aql
+
+        # Used for inserting/modifying table metadata
+        self.metadata_collection = workspace.entity_metadata_collection()
+
+        # Used for querying table items
+        self.handle: StandardCollection = workspace.handle.collection(name)
+
+        # Used for running AQL queries when necessary
+        self.aql: AQL = workspace.handle.aql
 
     def rows(self, offset: Optional[int] = None, limit: Optional[int] = None) -> Dict:
         """Return the desired rows in a table."""
@@ -71,6 +80,32 @@ class Table:
             keys = list(util.filter_unwanted_keys(doc).keys())
 
         return keys
+
+    def get_metadata(self) -> Optional[ArangoEntityDocument]:
+        """Retrieve metadata for this table, if it exists."""
+        try:
+            doc = next(self.metadata_collection.find({"item_id": self.name}, limit=1))
+        except StopIteration:
+            return None
+
+        return ArangoEntityDocument(**doc)
+
+    def set_metadata(self, raw_data: Dict) -> ArangoEntityDocument:
+        """Set metadata for this table."""
+        try:
+            data = TableMetadata(**raw_data)
+        except PydanticValidationError:
+            raise InvalidMetadata(raw_data)
+
+        existing = self.get_metadata()
+        if existing is not None:
+            existing.table = data
+            doc_meta = self.metadata_collection.update(existing.dict())
+        else:
+            entity = EntityMetadata(item_id=self.name, table=data, graph=None)
+            doc_meta = self.metadata_collection.insert(entity.dict())
+
+        return ArangoEntityDocument(**self.metadata_collection.get(doc_meta))
 
     def rename(self, new_name: str) -> None:
         """Rename a table."""
--- a/multinet/swagger/get_metadata.yaml
+++ b/multinet/swagger/get_metadata.yaml
@@ -1,0 +1,20 @@
+Retrieve the metadata of a table.
+---
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+
+responses:
+  200:
+    description: The permissions for the given workspace
+    schema:
+      $ref: "#/definitions/workspace_permissions"
+
+  404:
+    description: Specified workspace or table could not be found
+    schema:
+      type: string
+      example: workspace_or_table_that_doesnt_exist
+
+tags:
+  - table
--- a/multinet/swagger/set_metadata.yaml
+++ b/multinet/swagger/set_metadata.yaml
@@ -1,0 +1,35 @@
+Set the metadata of a table.
+---
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+  - name: metadata
+    in: body
+    description: The metadata to set (overwrites existing data)
+    required: true
+    schema:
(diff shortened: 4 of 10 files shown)


multinet/types.py
```diff
@@ -1,9 +1,63 @@
...
-from typing import Dict, Set
+from pydantic import BaseModel, Field
+from typing import List, Dict, Set, Optional, Any
 from typing_extensions import Literal, TypedDict
 
 EdgeDirection = Literal["all", "incoming", "outgoing"]
-TableType = Literal["all", "node", "edge"]
+
+TableType = Literal["node", "edge"]
+UnionTableType = Literal["all", TableType]
```
waxlamp, line 9: I checked the codebase and `TableType` isn't used anywhere. So I think it makes sense to just add `"all"` to `TableType` and not bother with this new `UnionTableType` (which sounds like it sets out the types for a thing called a "union table").

multinet/api.py
```diff
@@ -100,6 +102,23 @@ def get_table_rows(workspace: str, table: str, offset: int = 0, limit: int = 30)
     return Workspace(workspace).table(table).rows(offset, limit)
 
 
+@bp.route("/workspaces/<workspace>/tables/<table>/metadata", methods=["GET"])
+@require_reader
+@swag_from("swagger/get_metadata.yaml")
+def get_table_metadata(workspace: str, table: str) -> Any:
+    """Retrieve the metadata of a table, if it exists."""
+    metadata = Workspace(workspace).table(table).get_metadata()
```
JackWilb, line 110: Will this return gracefully if the table/workspace is not found? How about if it is found but the metadata doesn't exist?

multinet/api.py
```diff
@@ -100,6 +102,23 @@ def get_table_rows(workspace: str, table: str, offset: int = 0, limit: int = 30)
...
+    return "" if metadata is None else metadata.dict()
```
waxlamp, line 111: Since this returns a `dict` when there is metadata, shouldn't it return `{}` instead of `""` when there isn't any?

multinet/api.py
```diff
@@ -100,6 +102,23 @@ def get_table_rows(workspace: str, table: str, offset: int = 0, limit: int = 30)
...
+
+
+@bp.route("/workspaces/<workspace>/tables/<table>/metadata", methods=["PUT"])
+@require_reader
+@swag_from("swagger/set_metadata.yaml")
+def set_table_metadata(workspace: str, table: str) -> Any:
+    """Retrieve the rows and headers of a table."""
+    return Workspace(workspace).table(table).set_metadata(request.json).dict()
```
JackWilb, line 119: Again, what if the workspace/table isn't found

multinet/db/models/table.py
```diff
@@ -25,22 +33,23 @@ def flask_response(self) -> FlaskTuple:
 class Table:
     """Tables store tabular data, and are the root of all data storage in Multinet."""
 
-    def __init__(self, name: str, workspace: str, handle: StandardCollection, aql: AQL):
+    def __init__(self, name: str, workspace: workspace.Workspace):
```
waxlamp, line 36: You're right, this is a good change.

multinet/db/models/table.py
```diff
@@ -72,6 +81,32 @@ def headers(self) -> List[str]:
...
+        except PydanticValidationError:
+            raise InvalidMetadata(raw_data)
+
+        existing = self.get_metadata()
+        if existing is not None:
+            existing.table = data
+            doc_meta = self.metadata_collection.update(existing.dict())
+        else:
+            entity = EntityMetadata(item_id=self.name, table=data, graph=None)
+            doc_meta = self.metadata_collection.insert(entity.dict())
```
JackWilb, line 106: Does this enable the PUT? What if we POST new data? We might want to overwrite the whole metadata object, not just update it.

multinet/swagger/get_metadata.yaml
```diff
@@ -0,0 +1,20 @@
...
+---
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+
+responses:
+  200:
+    description: The permissions for the given workspace
+    schema:
+      $ref: "#/definitions/workspace_permissions"
```
JackWilb, line 11: This seems wrong, unless you're just reusing the return definition

multinet/db/models/table.py
waxlamp, line 106: This is not at all a big deal, but it feels like you could condense this code quite a bit just by letting the `EntityMetadata` initializer not have to take the `table` and `graph` arguments and set them both to `None` initially, then explicitly set the one that you do want. Then, line 100 could be `entity = self.get_metadata() or EntityMetadata(item_id=self.name)`, followed immediately by `entity.table = data` and then `doc_meta = self.metadata_collection.upsert(entity.dict())`. The total code block would look like this:

```suggestion
        entity = self.get_metadata() or EntityMetadata(item_id=self.name)
        entity.table = data
        doc_meta = self.metadata_collection.insert(entity.dict(), overwrite=True)
```

We'd have to make sure that the `overwrite = True` kwarg makes `.insert()` behave like an upsert. And this may not be worth doing at all, but I just like how the shorter code reads.

This is also making me think the whole `table`/`graph` metadata section split is an antipattern but we can fix that if it turns out to be a problem.

multinet/swagger/set_metadata.yaml
```diff
@@ -0,0 +1,35 @@
+Set the metadata of a table.
+---
+parameters:
+  - $ref: "#/parameters/workspace"
+  - $ref: "#/parameters/table"
+  - name: metadata
+    in: body
+    description: The metadata to set (overwrites existing data)
```
JackWilb, line 8: Seems to update, not just overwrite everything

multinet/swagger/set_metadata.yaml
```diff
@@ -0,0 +1,35 @@
...
+          - key: length
+            type: number
+
+
+
+responses:
+  200:
+    description: The permissions for the given workspace
+    schema:
+      $ref: "#/definitions/workspace_permissions"
```
JackWilb, line 26: Looks like another place that needs to be updated

multinet/swagger/set_metadata.yaml
```diff
@@ -0,0 +1,35 @@
...
+      type: object
+      example:
+        columns:
+          - key: test
+            type: label
+
+          - key: length
+            type: number
+
+
```
waxlamp, line 20:
```suggestion
```

multinet/uploaders/csv.py
```diff
@@ -95,6 +101,10 @@ def upload(
 
     # Create table and insert the data
     loaded_table = loaded_workspace.create_table(table, edges)
+
+    if metadata:
+        loaded_table.set_metadata(json.loads(metadata))
```
JackWilb, line 106: Does this need a try/except to catch invalid JSON?

multinet/swagger/set_metadata.yaml
waxlamp: Not sure why this is showing removal of only one blank line. My suggestion is to reduce the three blank lines between these sections to just one.

multinet/uploaders/csv.py
waxlamp, line 106: This will raise an exception if you pass non-json as the metadata argument. Should we catch it here and turn it into a 4xx error?

multinet/uploaders/swagger/csv.yaml
```diff
@@ -34,6 +34,21 @@ parameters:
...
+    schema:
+      type: object
+      example:
+        columns:
+          -
+            key: city
+            type: label
+          -
+            key: size
+            type: number
```
waxlamp, line 50: I know this broken item style is allowed, but elsewhere in our YAML files (even in other ones in this PR) we don't use it. For consistency I'd prefer to remove these linebreaks.

waxlamp (changes requested):
This looks awesome. Thanks for being so fast. I have a few requested changes. If you'd prefer me to submit those changes for a counterreview by you (in the interest of time, mainly) please let me know.

JackWilb (changes requested):
Here are a couple of questions and a couple of copy-paste problems I found

multinet/types.py
AlmightyYakob: Whoops, this is leftover from my initial approach. I actually removed [this](https://github.com/multinet-app/multinet-rfcs/blob/91534820bcdcd620eefc15dff8a54f0fdeb0900c/accepted/0009-table-metadata/README.md#L31) line from the implementation, as it seemed redundant and just got in the way. I'll update the RFC and revert this change.

multinet/uploaders/csv.py
AlmightyYakob: Yep

AlmightyYakob: Yep

multinet/api.py
AlmightyYakob: I'd say you're right, but this made me think of something; How should we handle existing tables? In the current implementation, any new table will at least have empty metadata, but this does nothing for existing tables. I think it would make sense to ensure that this metadata exists, by handling the case where it doesn't right in `get_metadata`. So the current case where it returns `None`, would instead initialize the metadata as empty, and return that.

multinet/uploaders/swagger/csv.yaml
AlmightyYakob: It's used elsewhere in this file, that's why I wrote it this way:

https://github.com/multinet-app/multinet-server/blob/6ef8fdca1aee350fa18395c398a5f454e5c1ff4e/multinet/uploaders/swagger/csv.yaml#L9-L36

multinet/api.py
AlmightyYakob: 1. If the table/workspace aren't found, the relevant 4XX errors will be raised and returned.
2. The `get_metadata` call returns `None` if the metadata doesn't exist. However, see my other comment with Roni regarding this.

multinet/api.py
AlmightyYakob: Same answer as above

multinet/db/models/table.py
AlmightyYakob: So just to clarify, `PUT` is intended to be used when creating/updating an entity, where as `POST` is intended to be used when creating new entities.

This code enables `PUT` at the table level, as in any existing table data is overwritten by the data passed into this function. You're correct that this function doesn't allow for updating the entire metadata object at once, which is by design. This class method is designed to only affect the table section of the metadata, and we don't really have a use case otherwise at the moment. I think in the future we'll update the API to better deal with metadata, but as a first approach, we're just focusing on table metadata.

Does that answer your question?

multinet/uploaders/swagger/csv.yaml
waxlamp: Ok. Then don't worry about it.

multinet/swagger/set_metadata.yaml
AlmightyYakob: Related to the other comment, but it overwrites the table metadata.

multinet/db/models/table.py
waxlamp: What do you mean by "enable the PUT"?

multinet/api.py
AlmightyYakob: This issue would cover that right? https://github.com/multinet-app/multinet-server/issues/405

waxlamp: Yes but it would be good to add this particular migration to that issue.

multinet/db/models/table.py
AlmightyYakob: It [looks](https://python-driver-for-arangodb.readthedocs.io/en/master/specs.html#arango.collection.StandardCollection.insert) like `overwrite` combined with `overwrite_mode` would give us the `upsert` behavior we desire. I think your code does read better, so thanks for that. I would commit this, but since I'll be changing the behavior for missing metadata as discussed above, this code will change slightly. I'll make this change after updating the behavior for `get_metadata`.

AlmightyYakob: I'm not sure what @JackWilb meant but what I mean is replacing the existing data with the specified new data (so `PUT` as in set this entity to this value, regardless of what might already be there).

JackWilb: @waxlamp : Sorry "enable PUT requests".

@AlmightyYakob: You've answered most of my question (the stuff related to PUT). However, I see that you use `.update()` here, which only updates the keys that are in `entity.dict`. Is it possible to remove a key using `update()`? I was wondering if we'd have a use case where we'd want to clear out keys

AlmightyYakob: I'm not sure if you can remove top level keys with `update`, but again this is only operating on the metadata for a table, scoped to the `table` key. Since this is done by creating a `TableMetadata` instance, any missing keys will take a value of whatever is default for that key (so likely `None` or some other default). So as far as Arango is concerned, every key will be updated, because the class instance has something for each of those keys, whether it's specified or default.

E.g. if you were to pass `{}` into the set metadata call, the result would be:

```
{
  ...,
  "table": {
    "columns": []
  }
}
```

Since the default for `columns` is `[]`.

AlmightyYakob: Actually, I was looking at newer docs than what we're running. It seems if `overwrite=True`, the existing doc will be replaced with the doc we specify, which I think is the correct behavior.

AlmightyYakob: 6118164

JackWilb: Perfect. I'll mark as resolved

multinet/api.py
waxlamp: @JackWilb, going to resolve this; unresolve if you still have a problem here.

waxlamp (approved):

JackWilb (approved):

PR #486, opened 2020-11-18
Use column typing when uploading CSVs

### LINKED ISSUE #471: CSV uploader doesn't handle data type properly
When uploading a CSV table, we don't preserve the data types properly, instead assuming it's a string. When reading in a field, we should try to use the appropriate [data type](https://www.arangodb.com/docs/3.7/aql/fundamentals-data-types.html), which would be one of the following:

* null
* boolean
* number
* string

However, sometimes it's not possible to infer the correct data type, as it could fit into multiple categories (e.g. a number that should instead be treated as a string). This is something that would be addressed by https://github.com/multinet-app/multinet-client/issues/7, and the corresponding backend change. I couldn't find a corresponding server issue for adding this, but I could have missed it. @waxlamp could you double check this? If that issue doesn't exist we can either create it, or update this issue.

### LINKED ISSUE #481: Implement column data types
This will implement API, validation, and data conversion as needed to support the column semantic types specification.

Use [RFC 0009](https://github.com/multinet-app/multinet-rfcs/tree/master/accepted/0009-table-metadata) as an implementation guide.

Depends on solving #482.

AlmightyYakob:
Closes #481
Closes #471


multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def process_row(
+    row_index: int, init_row: Dict[str, str], metadata: TableMetadata
+) -> Tuple[Dict[str, Any], List[ValidationFailure]]:
+    """Process a single row, returning the processed row, and any errors."""
+    validation_errors: List[ValidationFailure] = []
+
+    # Copy row
+    row: Dict[str, Any] = dict(init_row)
```
waxlamp, line 84: Maybe instead of `Any`, this could be a `Union` of the types that are eventually allowed.

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    for col in metadata.columns:
+        entry = row.get(col.key)
+
+        # If any of the following conditions are met, skip processing the entry
+        if entry is None:
+            validation_errors.append(MetadataColumnKeyNotFound(key=col.key))
```
waxlamp, line 91: The name of this exception seems funny to me. The error is that an expected column value was missing, right? That seems like a generic schema failure of some kind, rather than something as specific as "metadata column key not found". The phrase "column key" is also somewhat confusing.

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+            continue
+
+        process_entry = entry_processing_dict.get(col.type)
+        if process_entry is None:
+            continue
```
waxlamp, line 96: This gets engaged for `label` columns, right? A comment to explain this would be helpful.

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+        # Attempt normal entry processing
+        try:
+            # If the entry is an empty string, replace with None (null)
+            processed_entry = process_entry(entry) if entry else None
```
waxlamp, line 101: Do we want to use `entry.strip()` here? Or would that lead to trouble?

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+            )
+
+    return (row, validation_errors)
+
+
+def process_rows_with_metadata(
+    initial_rows: List[Dict[str, str]], metadata: TableMetadata
+) -> Tuple[List[Dict[str, Any]], List[ValidationFailure]]:
+    """Perform any processing of table rows with the supplied metadata."""
+    if not len(metadata.columns) or not len(initial_rows):
```
waxlamp, line 122:
```suggestion
    if not metadata.columns or not initial_rows:
```

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+                    row=row_index,
+                    column=col.key,
+                )
+            )
+
+    return (row, validation_errors)
+
+
+def process_rows_with_metadata(
+    initial_rows: List[Dict[str, str]], metadata: TableMetadata
```
waxlamp, line 119: Instead of the whole `metadata`, should this function just be passed the column types metadatum?

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+                    message=f"Cannot convert entry '{entry}' to type: {col.type}",
+                    row=row_index,
+                    column=col.key,
+                )
+            )
+
+    return (row, validation_errors)
+
+
+def process_rows_with_metadata(
```
waxlamp, line 118: I think this could just be named `process_rows`. The `_with_metadata` suffix is encoded in the signature and doesn't seem very useful to me.

multinet/uploaders/csv.py
```diff
@@ -84,8 +92,25 @@ def upload(
     except csv.Error:
         raise CSVReadError()
 
+    # TODO: This temporarily needs to be done here, so that validation of the metadata
+    # can be done before the table is actually created. Once the API is updated, this
+    # will change.
```
waxlamp, line 97: If you file an issue we can track this both here and in GitHub.

multinet/validation/__init__.py
```diff
@@ -33,3 +33,17 @@ class DuplicateKey(ValidationFailure):
...
     key: str
+
+
+class MetadataColumnKeyNotFound(ValidationFailure):
+    """A key in the column metadata was not found in the source data."""
+
+    key: str
+
+
+class IncompatibleMetadata(ValidationFailure):
```
waxlamp, line 44: It's not that the metadata is incompatible, it's that the data is not convertible. I think a better name here will help future devs. Maybe something like `TypeConversionFailure`.

multinet/validation/csv.py
```diff
@@ -58,49 +57,45 @@ def validate_edge_table(rows: Sequence[MutableMapping]) -> None:
 
         if fields:
             # i+2 -> +1 for index offset, +1 due to header row
-            data_errors.append(InvalidRow(columns=fields, row=i + 2))
+            validation_errors.append(InvalidRow(columns=fields, row=i + 2))
 
-    if len(data_errors) > 0:
-        raise ValidationFailed(data_errors)
+    return validation_errors
```
waxlamp, line 62: This is great, I like this better than directly raising the exception.

waxlamp (commented):
This looks awesome. Leaving my review as a comment rather than a request changes because while I did request some changes, I don't know if any of them are strictly required to make this whole. If you could address all my comments (making changes where you agree with them), then I can come back through with another review.

multinet/db/models/workspace.py
```diff
@@ -329,7 +329,10 @@ def create_aql_table(self, table: str, aql_query: str) -> Table:
         # In the future, the result of this validation can be
         # used to determine dependencies in virtual tables
         rows = list(self.run_query(aql_query))
-        validate_csv(rows, "_key", False)
+
+        errors = validate_csv(rows, "_key", False)
+        if len(errors):
```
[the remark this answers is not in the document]
AlmightyYakob: You're correct, empty lists are falsy in Python

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+def process_date_entry(entry: str) -> str:
+    """Try to read a date as an ISO 8601 string or unix timestamp."""
+
+    if entry.isdigit():
+        # Raises ValueError if the number is out of range
+        return datetime.fromtimestamp(int(entry)).isoformat()
+
+    # Raises ValueError if it cannot parse the string into a datetime object
+    return dateutilparser.parse(entry).isoformat()
```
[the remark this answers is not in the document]
AlmightyYakob: It doesn't, it only handles string datetimes.

multinet/metadata/utils.py
AlmightyYakob: I agree the name could be better, but yes, this error is raised when the underlying data is missing a column which was specified in the metadata.

multinet/metadata/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return cast_col_entry(entry)
+
+
+def process_date_entry(entry: str) -> str:
+    """Try to read a date as an ISO 8601 string or unix timestamp."""
+
+    if entry.isdigit():
+        # Raises ValueError if the number is out of range
+        return datetime.fromtimestamp(int(entry)).isoformat()
```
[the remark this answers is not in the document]
AlmightyYakob: Good point. It also seems that Arango chooses to store timestamps with millisecond precision, whereas I think conventionally timestamps are specified in second precision, so we'll have to do a conversion. Unfortunately, I don't think there's a surefire way to know if a specified timestamp uses second or millisecond precision, but I think going with the convention is fine for now.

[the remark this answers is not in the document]
AlmightyYakob: My bad, the second part I mentioned is actually irrelevant, as we're not storing dates as timestamps in arango, but as datestrings. So I think we should still assume that incoming timestamps use second precision. Just wanted to clear up confusion.

multinet/metadata/utils.py
AlmightyYakob: I'm honestly not sure, but I can't think of too many cases where you'd need leading/trailing spaces, so it might be a safe bet

multinet/metadata/utils.py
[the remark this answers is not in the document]
AlmightyYakob: 8969159

multinet/metadata/utils.py
AlmightyYakob: 44e5fcb

multinet/validation/__init__.py
AlmightyYakob: 1a9c194

multinet/metadata/utils.py
AlmightyYakob: d24fcbc

multinet/metadata/utils.py
AlmightyYakob: d24fcbc

multinet/metadata/utils.py
AlmightyYakob: a9bf0f7

AlmightyYakob:
@waxlamp I reorganized this code somewhat, sorry for any confusion there. Mainly I didn't like that the module was named `metadata`, when in reality it didn't do much with metadata besides taking it as one of the inputs for processing rows. I think `processing` makes more sense, as that's really what we're doing in that module.

multinet/metadata/utils.py
AlmightyYakob: @waxlamp Any ideas? Maybe `InvalidColumn` or something? `InvalidMetadata` would work too, but that's already a server error for that, so it might be confusing to have a validation error under that name as well.

multinet/metadata/utils.py
AlmightyYakob: I think for now we should leave it as is, and if we encounter an error as a result of not stripping this input, we can address it then. My reasoning for this is we shouldn't unconditionally modify input data unless we have a compelling reason to.

multinet/uploaders/csv.py
AlmightyYakob: https://github.com/multinet-app/multinet-server/issues/493

multinet/metadata/utils.py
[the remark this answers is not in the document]
waxlamp: If we store them as ISO date strings, then the precision can go up to microseconds, AFAICT:

```python
import datetime
datetime.datetime.now().isoformat()
```

gives me a result of `2020-11-25T10:00:03.144373`. Similarly, `datetime.datetime.fromtimestamp(0.001).isoformat()` gives me `1969-12-31T19:00:00.001000` and `datetime.datetime.fromtimestamp(0).isoformat()` gives `1969-12-31T19:00:00`. All of these strings parse nicely into JavaScript `Date` objects, so I think in general converting to float and pushing these through string conversion into Arango will work out just fine.

Is that the current situation?

Pipfile
```diff
@@ -20,6 +20,7 @@ gunicorn = "==20.0.4"
 authlib = "==0.14.1"
 pyjwt = ">=1.7.1"
 pydantic = ">=1.7.2"
+python-dateutil = "==2.8.1"
```
waxlamp, line 23: Since this library is out of 0.x, I think you should use a version range here (like `pydantic` does above).

waxlamp (changes requested):
One minor update to a Pipfile entry, and one confusion to clear up about how date processing works. Besides those small items, this looks really great.

multinet/metadata/utils.py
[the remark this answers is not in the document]
AlmightyYakob: Yes that is the current situation, i was just stating that I think the units of the incoming timestamps should be seconds, since that's the most common way to specify timestamps. They can still be floats, so there's no loss of precision.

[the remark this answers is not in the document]
waxlamp: Oh I think I understand. When you said "millisecond precision" you meant that, e.g., JavaScript timestamps are expressed in milliseconds (your use of "precision" had confused me). Standard timestamps do indeed come in *units* of seconds, but their *precision* is in microseconds due to the use of floats.

Ok I think we're on the same page and this was just a vocabulary issue. Thanks!

multinet/metadata/utils.py
AlmightyYakob: We decided offline on `MissingColumn`

waxlamp: Let's go with `MissingColumn`.

waxlamp (approved):

waxlamp (approved):

Issue #491, opened 2020-11-20
Remove "entity metadata" concept


waxlamp:
In a discussion with @AlmightyYakob, we decided that we don't have a compelling need to combine table and network metadata types together in one "entity type" model, because it will likely complicate the logic for teasing apart the metadata types in later processing.

The plan for now is to remove the `EntityMetadata` and `GraphMetadata` models, and reconstitute the network metadata concept if and when it is needed.

Relevant code: https://github.com/multinet-app/multinet-server/blob/2cf8989b61bf3d9cf152e7df8cb462e41351f1d9/multinet/types.py#L12-L34

AlmightyYakob:
O

AlmightyYakob:
Once integration with `multinetjs` is added, this change will have to be applied there as well.

Issue #398, opened 2020-06-22
Write an "fsck" script to validate data integrity


waxlamp:
Since we use some indirection to manage the state of our database, it would make sense to write a script that can validate that the data state is ok. This comes up specifically because of how we non-atomically build new workspaces by creating an Arango database to house data for the new workspace, and adding an entry to the workspace mapping collection to track it. If there's a failure in between these steps, we'd be left with a bad DB state.

See original discussion [here](https://github.com/multinet-app/multinet-server/pull/396#discussion_r442343182).

PR #492, opened 2020-11-23
Add migrations

jjnesbitt:
This is a POC that I've spent a little time on, and wanted to get feedback on. Since we may have to do a migration as a part of multinet-app/multinet-server#491, it seemed like a good time to tackle this.

The general approach I've come up with is to have a parent `Migration` class, which all migrations must inherit from. This contains a `run` function, which any migration must override, and is the entry point for any migration code. I've laid out some examples of how the inheritance works in the `migrations/test.py` file.

These migrations are loaded by loading all classes that inherit from `Migration` defined under the `migrations` module, and then if needed, running their `run` functions to initiate a migration. The code within the `run` method (and any supporting class methods) would be normal python code, that can make use of importing the models, db functions, etc.

In order to avoid running redundant migrations, it's my goal to define a `migrations` collection in the `_system` database, which would store a document for each migration that has been previously run. That way, instead of re-running a migration when it might not be necessary, if a migration document exists in the `migrations` collection, it is skipped.

Not all of this has been implemented yet, as I'm just doing a POC and want to make sure this is the right path before doing more work on it. However, so far I quite like how this is structured. All that's required to add a new migration is to add a migration class under the `migration` module. I'd think as part of our deploy/CI, we would run `pipenv run migrate` before actually serving the application, to ensure that everything is up to date. This would almost always be a no-op, besides checking the `migrations` collection.

Please let me know what you guys think. If we mostly agree that this is a good approach, I can continue my work here.


multinet/migrations/test.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+# This performs a migration successfully
+class TestMigrationA(Migration):
+    """Test Migration."""
+
+    @staticmethod
+    def run() -> None:
+        """Run Migration."""
+        print("MIGRATION!")
```
waxlamp, line 12: :joy:

waxlamp (commented):
This is looking good. Some thoughts:
1. Instead of recording in the DB that a migration has run, we should strive to make each migration idempotent instead. Then we can simply run every migration and be confident that the DB is in fact up to date. This could be difficult or impossible to do, but my intuition tells me that it *should* be possible.
2. This infrastructure may also help us to create an "fsck" script that checks to see whether the schema is healthy, whether there are things to "garbage collect", etc.
3. Did you look into existing migration frameworks for ArangoDB? I found a couple in my 30 seconds of googling but they all were independent GitHub projects with not a lot of stars/forks.

AlmightyYakob:
> 1. Instead of recording in the DB that a migration has run, we should strive to make each migration idempotent instead. Then we can simply run every migration and be confident that the DB is in fact up to date. This could be difficult or impossible to do, but my intuition tells me that it _should_ be possible.

I think we should strive to make our migrations idempotent regardless, but I don't see a reason to tie our hands like that. Moreover, even if the migration itself _is_ idempotent, it will likely still have to iterate over everything it may have to change (it's "migration space" so to speak), to ensure that it doesn't have to change anything. To me that's an unnecessary waste of time, especially if we're trying to run these migrations on every application startup. Django stores migrations history in the DB as well, and is actually where I got this idea from. [Citation](https://stackoverflow.com/questions/27430688/how-does-django-know-which-migrations-have-been-run).

> 2. This infrastructure may also help us to create an "fsck" script that checks to see whether the schema is healthy, whether there are things to "garbage collect", etc.

Good idea, although I think it'd be good to separate a generalized script like that from migrations.

> 3. Did you look into existing migration frameworks for ArangoDB? I found a couple in my 30 seconds of googling but they all were independent GitHub projects with not a lot of stars/forks.

I looked into some briefly, and seemed to find something similar. Namely, the projects that I found around this rely on specific DSLs, encoded in YAML. I didn't like the idea of tying our hands to something like that, and thought it better to just have general migration classes that can run any operation we need. That way, we can use our internal functions/classes/etc.

waxlamp:
> > 1. Instead of recording in the DB that a migration has run, we should strive to make each migration idempotent instead. Then we can simply run every migration and be confident that the DB is in fact up to date. This could be difficult or impossible to do, but my intuition tells me that it _should_ be possible.
> 
> I think we should strive to make our migrations idempotent regardless, but I don't see a reason to tie our hands like that. Moreover, even if the migration itself _is_ idempotent, it will likely still have to iterate over everything it may have to change (it's "migration space" so to speak), to ensure that it doesn't have to change anything. To me that's an unnecessary waste of time, especially if we're trying to run these migrations on every application startup. Django stores migrations history in the DB as well, and is actually where I got this idea from. [Citation](https://stackoverflow.com/questions/27430688/how-does-django-know-which-migrations-have-been-run).
> 

I actually don't think we should run migrations on *every* startup--rather, we should be able to control it by supplying a flag. (By default, I'd say we *shouldn't* run migrations on every startup, since database migration is something I think of as a deliberate and rare action. I'm willing to be shown different best practices in other DBMS systems, though.) That reduces some of the annoyance in always running idempotent all-table processes, but if other systems keep track of what has been run in order not to run it again, that's fine with me.

We're also sort of in danger of trying to roll our own migration system here, when Django for instance comes with one that actual DB experts have crafted. So let's make sure to keep it simple so that we can always have a handle on how it works, and it doesn't try to do too much.

> > 1. This infrastructure may also help us to create an "fsck" script that checks to see whether the schema is healthy, whether there are things to "garbage collect", etc.
> 
> Good idea, although I think it'd be good to separate a generalized script like that from migrations.

Agreed. I was sort of thinking out loud, since there's at least a conceptual connection between performing DB migrations and making sure the DB is in a good state. Just FYI, #398 is already tracking an "fsck" idea.

> > 1. Did you look into existing migration frameworks for ArangoDB? I found a couple in my 30 seconds of googling but they all were independent GitHub projects with not a lot of stars/forks.
> 
> I looked into some briefly, and seemed to find something similar. Namely, the projects that I found around this rely on specific DSLs, encoded in YAML. I didn't like the idea of tying our hands to something like that, and thought it better to just have general migration classes that can run any operation we need. That way, we can use our internal functions/classes/etc.

Ok cool. I'll just repeat my warning from above: let's not try to get too clever with these migrations, though I do agree that Python is a better "DSL" for our needs than some 3rd party YAML-based thing.

AlmightyYakob:
> I actually don't think we should run migrations on _every_ startup--rather, we should be able to control it by supplying a flag. (By default, I'd say we _shouldn't_ run migrations on every startup, since database migration is something I think of as a deliberate and rare action. I'm willing to be shown different best practices in other DBMS systems, though.) That reduces some of the annoyance in always running idempotent all-table processes, but if other systems keep track of what has been run in order not to run it again, that's fine with me.
> 
> We're also sort of in danger of trying to roll our own migration system here, when Django for instance comes with one that actual DB experts have crafted. So let's make sure to keep it simple so that we can always have a handle on how it works, and it doesn't try to do too much.
> 

I may have misspoke there. I don't want to run migrations every time the app is created (i.e. the `create_app` function). Rather, I was aiming to specify a step before the app create/serve step, which runs migrations if needed. That way, in production, we can guarantee our database is up to date. I'm not sure how else we'd trigger app migrations (at least in a sane way).

> > I looked into some briefly, and seemed to find something similar. Namely, the projects that I found around this rely on specific DSLs, encoded in YAML. I didn't like the idea of tying our hands to something like that, and thought it better to just have general migration classes that can run any operation we need. That way, we can use our internal functions/classes/etc.
> 
> Ok cool. I'll just repeat my warning from above: let's not try to get too clever with these migrations, though I do agree that Python is a better "DSL" for our needs than some 3rd party YAML-based thing.

I agree. IMO the only clever thing about what I've put forward is the automatic loading of migrations, which is really more of a quality of life thing. Besides that, it's pretty straight forward: specify the migration code within the defined class.


Another question is do we want to support backwards migrations or not, as many migration tools support reversing migrations. I doubt we'd really see much use of it, but just thought I'd bring it up.

waxlamp:
> > I actually don't think we should run migrations on _every_ startup--rather, we should be able to control it by supplying a flag. (By default, I'd say we _shouldn't_ run migrations on every startup, since database migration is something I think of as a deliberate and rare action. I'm willing to be shown different best practices in other DBMS systems, though.) That reduces some of the annoyance in always running idempotent all-table processes, but if other systems keep track of what has been run in order not to run it again, that's fine with me.
> > We're also sort of in danger of trying to roll our own migration system here, when Django for instance comes with one that actual DB experts have crafted. So let's make sure to keep it simple so that we can always have a handle on how it works, and it doesn't try to do too much.
> 
> I may have misspoke there. I don't want to run migrations every time the app is created (i.e. the `create_app` function). Rather, I was aiming to specify a step before the app create/serve step, which runs migrations if needed. That way, in production, we can guarantee our database is up to date. I'm not sure how else we'd trigger app migrations (at least in a sane way).
> 

Sounds good.

> > > I looked into some briefly, and seemed to find something similar. Namely, the projects that I found around this rely on specific DSLs, encoded in YAML. I didn't like the idea of tying our hands to something like that, and thought it better to just have general migration classes that can run any operation we need. That way, we can use our internal functions/classes/etc.
> > 
> > 
> > Ok cool. I'll just repeat my warning from above: let's not try to get too clever with these migrations, though I do agree that Python is a better "DSL" for our needs than some 3rd party YAML-based thing.
> 
> I agree. IMO the only clever thing about what I've put forward is the automatic loading of migrations, which is really more of a quality of life thing. Besides that, it's pretty straight forward: specify the migration code within the defined class.
> 

:+1: 

> Another question is do we want to support backwards migrations or not, as many migration tools support reversing migrations. I doubt we'd really see much use of it, but just thought I'd bring it up.

At this point, I'd say no. But that would change if we ended up with some demand for it.

Pipfile
```diff
@@ -48,7 +48,7 @@ python_version = "3.7"
 allow_prereleases = true
 
 [scripts]
-serve = "flask run -p ${FLASK_SERVE_PORT}"
+serve = "flask run --eager-loading -p ${FLASK_SERVE_PORT}"
```
AlmightyYakob, line 51: This change is made to ensure that running the dev server locally fails immediately if there are unapplied migrations. Normally, this runs in a "lazy loading" mode, which won't throw this error until a request has been made.

JackWilb (commented):
How would this work with our production set up? Where would we see that we have unapplied migrations or is this just for local dev?

Is this just the infrastructure? Do you have any examples of a migration file to apply?

AlmightyYakob:
> How would this work with our production set up? Where would we see that we have unapplied migrations or is this just for local dev?

We would run the `migrate` command as a step before running the `serve` script, on deploy. That way, it ensures all migrations that exist are run before serving the app.

> Is this just the infrastructure? Do you have any examples of a migration file to apply?

The first migration within this infrastructure will likely be one to address https://github.com/multinet-app/multinet-server/issues/491.

JackWilb:
Sounds good to me! I'll let Roni give you the approval since he's more familiar with the server code

multinet/migrations/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        for migration in all_migrations
+        if migration.__name__ not in applied_migrations
+    ]
+
+
+def run_migrations() -> None:
+    """Run all defined migrations."""
+
+    migrations = get_unapplied_migrations()
+    if not len(migrations):
```
waxlamp, line 46: Optional suggestion:
```suggestion
    if not migrations:
```

multinet/migrations/__init__.py
```diff
@@ -0,0 +1,54 @@
...
+        print("All migrations up to date!")
+        return
+
+    for migration in migrations:
+        migration.run()
```
waxlamp, line 51: If the migration fails, what happens? You wouldn't want to then store the migration as having been completed, etc. I would like to make the failure case explicit here; let me know what you think.

multinet/migrations/util.py
```diff
@@ -0,0 +1,60 @@
...
+from multinet.db import system_db
+from multinet.migrations.base import BaseMigration
+from multinet.migrations.types import Migration, MigrationDocument
+from types import ModuleType
+from typing import List, Generator
+
+
+# TODO: Add creation and return of `migrations` collection in _system db. This will
+# hold a document for each migration that exists. This can be used to check the
+# migrations that have been run, what needs to be run, etc.
```
waxlamp, line 14: I'd like to never leave a TODO in the code like this if it's not referencing a GitHub issue. Can you either complete this work now, or file the issue?

multinet/migrations/util.py
```diff
@@ -0,0 +1,60 @@
...
+    """Return all migrations from the supplied list of modules."""
+
+    migrations: List[Migration] = []
+    for module in modules:
+        migrations.extend(
+            [
+                obj
+                for _, obj in inspect.getmembers(module, inspect.isclass)
+                if is_migration(obj)
+            ]
```
waxlamp, line 36: It probably doesn't matter for this case, but `extend()` takes an iterable, which means you can omit the square brackets to avoid coercing the list comprehension to an actual list object before it gets passed to `extend()`. This is more just giving a style tip than anything else, but technically the performance would be improved as well.

multinet/migrations/util.py
```diff
@@ -0,0 +1,60 @@
...
+    return migrations
+
+
+def migrations_collection() -> StandardCollection:
+    """Return the collection that stores previously applied migrations."""
+
+    if not system_db().has_collection("migrations"):
+        return system_db(False).create_collection("migrations")
+
+    return system_db(False).collection("migrations")
```
waxlamp, line 48: Making the boolean argument usage more explicit will aid in readability:
```suggestion
        return system_db(readonly=False).create_collection("migrations")

    return system_db(readonly=False).collection("migrations")
```

waxlamp (commented):
This looks good. I listed a few suggestions--see if you want to accept them and then pull on me for a final review.

multinet/migrations/__init__.py
AlmightyYakob: I agree, this is something I just hadn't addressed. I think my reasoning in this simple implementation is that if something were to go wrong, it would throw an error and halt execution. That could obviously be handled more elegantly here, so I can handle that case. I also think it'd important to return a non-zero exit code in the actual `migrate` script, if any do fail, to prevent further actions.

multinet/migrations/util.py
AlmightyYakob: Just something I forgot to remove, this is addressed now. I'll remove it.

multinet/migrations/__init__.py
AlmightyYakob: 9c3b77c

AlmightyYakob:
Closed in favor of https://github.com/multinet-app/multinet-api

waxlamp:
Lesson learned here--don't roll your own migrations 😸

Issue #433, opened 2020-07-17
Remove caching once it's no longer needed 


AlmightyYakob:
Once the need for our current caching system has become irrelevant, either through implementation of Foxx Microservices, code refactor, etc., we'll want to remove both the caching on these functions, as well as any actions which may become moot from the caching being removed (#432 is an example, if implemented).

PR #494, opened 2020-11-25
Remove caching

jjnesbitt:
This is largely a test to see how the removal of caching impacts the app performance. I'd say if it appears to have a large impact, then looking into replacing this with memcached (#453) is worth it. If this appears to have little to no impact, I think the best move is to remove it altogether (#433).


waxlamp:
> This is largely a test to see how the removal of caching impacts the app performance. I'd say if it appears to have a large impact, then looking into replacing this with memcached (#453) is worth it. If this appears to have little to no impact, I think the best move is to remove it altogether (#433).

Either way, it looks like the plan is to remove caching from inside the app. Good stuff.

AlmightyYakob:
I've tested this on the deploy preview, and it seems to boast similar or very slightly slower speeds than the current production app. Not an issue at all imo, and I think we should remove caching until we have a compelling reason otherwise. If/when that time comes, I think memcached is a good start.

AlmightyYakob:
You may notice there is still the `lru_cache` decorator being used in the following two places https://github.com/multinet-app/multinet-server/blob/eddb9c8389eca9fb966af5a5886bbfa374e7d92f/multinet/db/__init__.py#L51-L55
https://github.com/multinet-app/multinet-server/blob/eddb9c8389eca9fb966af5a5886bbfa374e7d92f/multinet/db/__init__.py#L132-L140

The difference between these two code sections and the rest of what was removed is that these responses are never modified, and should always return the same thing. We've not had a problem with these cache decorators before, and have only had issues when caching things that _do_ change, since each worker keeps its own cache. I can understand the desire to purge all caching from the app, but I don't really see the harm in keeping these two uses of it around.

waxlamp:
> You may notice there is still the `lru_cache` decorator being used in the following two places
> 
> https://github.com/multinet-app/multinet-server/blob/eddb9c8389eca9fb966af5a5886bbfa374e7d92f/multinet/db/__init__.py#L51-L55
> 
> 
> https://github.com/multinet-app/multinet-server/blob/eddb9c8389eca9fb966af5a5886bbfa374e7d92f/multinet/db/__init__.py#L132-L140
> 
> The difference between these two code sections and the rest of what was removed is that these responses are never modified, and should always return the same thing. We've not had a problem with these cache decorators before, and have only had issues when caching things that _do_ change, since each worker keeps its own cache. I can understand the desire to purge all caching from the app, but I don't really see the harm in keeping these two uses of it around.

The only harm I can think of is the small cognitive load for keeping track of why caching is used here, if it doesn't get us anything in return. If it doesn't improve performance significantly (which we should definitely test out) then I would say get rid of them. It is not a big deal if they stay, but it would be slightly better if they went.

waxlamp (approved):
This looks fine to me, though with the upcoming plan to use S3 to manage uploads, I think you could also remove the remaining caching instances to leave this whole caching thing off of our minds entirely.

Approving, but I'd urge you to either strike the remaining caching, or at least carry out some kind of performance test to see if it even matters.

PR #498, opened 2020-12-04
Fix failing CI

waxlamp:
The fix was to remove `uuid` from Pipfile.

@AlmightyYakob, could you verify that things work ok this way?


AlmightyYakob (approved):
Seems to work no problem :+1:

PR #501, opened 2021-01-21
Add script that sets the arango server.pem file on db.multinet.app

JackWilb:
This is the script that will run on the letsencrypt deploy hook. It combines the fullchain and privkey into one file in the home directory, sets permissions, and then restarts the arango service so it's using the updated cert. 

I modified this slightly for the db-testing server by using /etc/letsencrypt/live/db-testing.multinet.app/... as the folder to grab the certificate files from


waxlamp (commented):
It seems like this should be hooked into our ansible provisioning (Ansible has ways to send files over to the provisioned server, and methods for restarting services).

AlmightyYakob:
> It seems like this should be hooked into our ansible provisioning (Ansible has ways to send files over to the provisioned server, and methods for restarting services).

I agree, since our ansible provisioning (at [multinet-deploy](https://github.com/multinet-app/multinet-deploy)) is where we setup letsencrypt anyway, so we could install this script as a task after the letsencrypt tasks run. It would also allow us to encode these actions into ansible, making them more concrete/safe imo.

JackWilb:
@AlmightyYakob, @waxlamp I've updated this branch to use ansible for the provisioning of this script. I tested it successfully on db-testing.multinet.app and had no issue. 

I moved the ansible files to their own folder, updated the README with the correct instructions, and removed all the vagrant references (since vagrant will no longer work with this setup). 

Let me know your thoughts.

devops/ansible/ssl.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: Add cronjob for cert renewal
+  cron:
+    name: certbot renew
+    minute: "0"
+    hour: "0,12"
+    job: "python -c 'import random; import time; time.sleep(random.random() * 3600)' && certbot renew -q"
+    cron_file: /etc/crontab
+    user: root
+  become: true
```
AlmightyYakob, line 39: I thought certbot setup automatic renewal when installed?

JackWilb: It does from the snap, but the install instructions for that are more complicated than just grabbing it from apt. The apt version doesn't automatically do the cronjob

devops/ansible/ssl.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Set server name inside of the post cert renewal hook script
+  replace:
+    path: /etc/letsencrypt/renewal-hooks/deploy/arango-ssl.sh
+    regexp: '(SERVER_NAME)'
+    replace: '{{ server_name }}'
+  become: true
+
+- name: Generate a certificate
+  shell:
+    cmd: certbot certonly -d {{ server_name }} --standalone -m jwilburn@sci.utah.edu --agree-tos -n
```
waxlamp, line 23: It seems like your email address should be an input value (either via prompt or environment variable or something).

devops/ansible/ssl.yml
```diff
@@ -0,0 +1,39 @@
...
+  become: true
+
+- name: Run arango-ssl.sh (doesn't run above since it's not a renewal)
```
waxlamp, line 26: I don't understand what the parenthesis means.

devops/ansible/ssl.yml
```diff
@@ -0,0 +1,39 @@
...
+  shell:
+    cmd: /etc/letsencrypt/renewal-hooks/deploy/arango-ssl.sh
+  become: true
+
+- name: Add cronjob for cert renewal
+  cron:
+    name: certbot renew
+    minute: "0"
+    hour: "0,12"
+    job: "python -c 'import random; import time; time.sleep(random.random() * 3600)' && certbot renew -q"
```
waxlamp, line 36: What's the randomized delay for?

devops/ansible/arango-ssl.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+
+cat /etc/letsencrypt/live/SERVER_NAME/fullchain.pem /etc/letsencrypt/live/SERVER_NAME/privkey.pem > /home/ubuntu/server.pem
+chmod 644 /home/ubuntu/server.pem
+
+systemctl restart arangodb3
```
waxlamp, line 6: Is this script something that has to run on renewal? If not, these steps should be converted to Ansible playbook steps. If yes, maybe a comment block in the script explaining why these are not done via Ansible would be good.

waxlamp (commented):
Thanks for putting the work into this, Jack. Some questions below.

devops/ansible/ssl.yml
JackWilb: I add the script for making the server.pem file above, but it doesn't run when the cert is created, only when the cert is renewed. I was just saying that we have to manually trigger it, even though it seems like it should run on the hook

devops/ansible/ssl.yml
JackWilb: This is from the certbot site. It's so the letsencrypt server doesn't get crushed at 00:00 and 12:00. It makes a random delay before running the renew command. This code is directly from the old version of the docs (it was removed when they chose to prefer snaps)

devops/ansible/arango-ssl.sh
JackWilb: Yeah, this runs on a successful renewal hook. Thus when the cert renews, this is triggered

devops/ansible/ssl.yml
JackWilb: Fixed in 0dbcfbd

devops/ansible/arango-ssl.sh
JackWilb: Fixed in 196b0df

devops/ansible/arangodb.yml
```diff
@@ -49,33 +56,7 @@
...
-      become: true
-
-    - name: Add an exception to the permissions for the arangodb user
-      shell:
-        chdir: /home/ubuntu/
-        cmd: setfacl -m u:arangodb:r server.pem
-      become: true
-
-    - name: Add the SSL filepath to the config
+    - name: Add the SSL filepath to the config (cert file provided by ssl.yml)
```
AlmightyYakob, line 59: I'm still slightly confused as to what this block is accomplishing. Where is `/home/ubuntu/server.pem` populated?

JackWilb: It happens in the SSL file by manually triggering the shell script (`arango-ssl.sh`).

devops/ansible/arango-ssl.sh
waxlamp: This is fine for now, but it occurs to me that you can absolutely just put these steps in an independent playbook and invoke that on renewal.

devops/ansible/arangodb.yml
waxlamp: @AlmightyYakob is this ok?

AlmightyYakob (approved):

PR #503, opened 2021-01-22
Add eurovis 2019 data sample to the example data sets

JackWilb:
This uses data from twitter during  the Eurovis 2019 conference. The data was first pulled for [this repo](https://github.com/visdesignlab/mvnv-study/) and I'm repurposing it for this example. 

For the data processing, I used the les mis processing script as an example and expanded it a little. I reduced the nodes from 333 to 100 and then reduced the links to 1/10th of the links that existed between the remaining 100 nodes. I'm sure the script is not an optimal solution for this size reduction, but it works pretty well. 

You can see this data uploaded in a public workspace at: https://multinet.app/#/workspaces/eurovis-2019/


data/eurovis-2019/script/.gitignore
```diff
@@ -0,0 +1,3 @@
+connections.csv
+Eurovis2019Network.json
+people.csv
```
waxlamp, line 3: Unless these files are very big (which in this case, I can't imagine they are), my practice has been to simply run the processing script and check these result files into GitHub alongside the raw data. The processing script would be there largely as an indicator of provenance for the resulting data files.

So to be explicit: I would remove this .gitignore file, and add the three files it references.

waxlamp (changes requested):
Looks great! I'm just recommending you check in the resulting data files as has been done for the other datasets.

data/eurovis-2019/script/.gitignore
JackWilb: These files are checked in at `data/eurovis-2019/connections.csv` and `data/eurovis-2019/people.csv`. I copied your format of moving them up a level in the directory structure and out of the scripts folder.

waxlamp: Oh gotcha. That's so that if you run the scripts here, you don't see untracked files showing up in your git repo. I'm not sure my earlier practice was great, but since you've remained consistent with it, I'm fine with this approach.

waxlamp (approved):

